PR #68, opened 2016-12-01
Fix Environment APIs to return TaskDefinition field correctly

kiranmeduri:
#### Summary
Fix Environment APIs to return TaskDefinition field correctly

#### Implementation details
Fix Environment APIs to return TaskDefinition field correctly

#### Testing

[x] Binary built locally and unit-tests pass (`make`)  
[x] Build in docker succeeds (`make release`)

Fix Environment APIs to return TaskDefinition field correctly

#### Licensing


daemon-scheduler/pkg/api/v1/api_test.go
```diff
@@ -64,13 +70,119 @@ func (suite *APITestSuite) TestPing() {
...
+}
+
+func (suite *APITestSuite) assertSame(environment *types.Environment, environmentModel *models.Environment) {
+	assert.Equal(suite.T(), environment.Name, aws.StringValue(environmentModel.Name))
+	assert.Equal(suite.T(), environment.Cluster, environmentModel.InstanceGroup.Cluster)
+	assert.Equal(suite.T(), environment.DesiredTaskDefinition, environmentModel.TaskDefinition)
+}
+
+func (suite *APITestSuite) getRouter() *mux.Router {
+	return NewRouter(suite.api)
```
poojamaiya: I get that this actually tests the router and also prevents us from re-writing the router. But the reason I had re-written it was so we explicitly define invalid paths in the router to make sure that the handler functions handle them in case the call reaches it (Ex. https://github.com/blox/blox/blob/master/cluster-state-service/handler/api/v1/taskapis_test.go#L399). Might be an overkill. Your call!

daemon-scheduler/pkg/api/v1/api_test.go
```diff
@@ -64,13 +70,119 @@ func (suite *APITestSuite) TestPing() {
...
+	request, err := http.NewRequest("DELETE", "/v1/environments/"+name, nil)
+	assert.Nil(suite.T(), err, "Unexpected error")
+
+	responseRecorder := httptest.NewRecorder()
+	suite.router.ServeHTTP(responseRecorder, request)
+
+	assert.Equal(suite.T(), http.StatusNotFound, responseRecorder.Code)
+}
+
+func (suite *APITestSuite) assertSame(environment *types.Environment, environmentModel *models.Environment) {
```
poojamaiya: nit: assertSame -> assertSameEnvironment?

daemon-scheduler/pkg/api/v1/api_test.go
```diff
@@ -64,13 +70,119 @@ func (suite *APITestSuite) TestPing() {
...
 }
 
-func (suite *APITestSuite) getRouter() *mux.Router {
-	r := mux.NewRouter().StrictSlash(true)
-	s := r.Path("/v1").Subrouter()
+func (suite *APITestSuite) TestGetEnvironmentReturnsError() {
+	name := "testEnv"
+	err := errors.New("Error from GetEnvironment")
+	suite.environment.EXPECT().GetEnvironment(gomock.Any(), name).Return(nil, err)
+	request, err := http.NewRequest("GET", "/v1/environments/"+name, nil)
```
poojamaiya: I'd move this to a separate function since it's being used in multiple functions and any changes to it later will not be hard -
func (suite *APITestSuite) generateGetEnvRequest(t *testing.T, name string) *http.Request {
    request, err := http.NewRequest("GET", "/v1/environments/"+name, nil)
    assert.Nil(suite.T(), err, "Unexpected error")
    return request
}

daemon-scheduler/pkg/api/v1/api_test.go
kiranmeduri: I don't like repeating environment multiple times esp. given that parameters are of environment type.

daemon-scheduler/pkg/api/v1/api_test.go
kiranmeduri: done

daemon-scheduler/pkg/api/v1/api_test.go
poojamaiya: It's just that this the API package and we may want to assert deployments later on. But we can deal with it when we have to later on I guess.

PR #75, opened 2016-12-05
Fix `make generate` for the daemon scheduler and cluster state service

samuelkarp:
#### Summary
Fixes `make generate` for the daemon scheduler so `make` actually works.

#### Implementation details
* Set the shebang to `/bin/bash` instead of `/bin/sh` so it uses Bash rather than another shell.
* Fixed invalid JSON in `swagger.json` for the daemon scheduler.
* Fixed invalid JSON in `swagger.json` for the cluster state service.

#### Testing

- [x] Binary built locally and unit-tests pass (`make`)  
- [x] Build in docker succeeds (`make release`)

None

#### Licensing

### DIFF
--- a/cluster-state-service/handler/api/v1/swagger/swagger.json
+++ b/cluster-state-service/handler/api/v1/swagger/swagger.json
@@ -229,7 +229,7 @@
         "registeredResources",
         "remainingResources",
         "status",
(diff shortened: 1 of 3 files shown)


kylbarnes:
Merging swagger json fixes.

PR #77, opened 2016-12-05
Add kinesis as a consumer.

jhspaybar:
#### Summary
Adds Kinesis as a consumer type

#### Implementation details
additional command line arguments uses kinesis instead of sqs

#### Testing
- [x ] Binary built locally and unit-tests pass (`make`)  
- [ x] Build in docker succeeds (`make release`)

New tests cover the changes: no.  They will be added once this approach and feature is determined to be wanted.

Pass --stream instead of --queue to listen to a Kinesis stream for ECS events.

#### Licensing


cluster-state-service/handler/run/run.go
```diff
@@ -92,15 +92,27 @@ func StartClusterStateService(queueName string, bindAddr string, etcdEndpoints [
...
-		return errors.Wrapf(err, "Could not start the consumer")
-	}
+		// start event consumer
+		consumer, err := event.NewSQSConsumer(sqsClient, processor, queueName)
+		if err != nil {
+			return errors.Wrapf(err, "Could not start the consumer")
+		}
+
+		go consumer.PollForEvents(ctx)
+	} else {
```
dbenhur: Should check for `len(streamName) > 0` for the kinesis path; else should return a "queue or stream required" error

cluster-state-service/cmd/root.go
```diff
@@ -46,6 +47,7 @@ a localized data store, which provides you a near-real-time view of your cluster
 	}
 	// TODO: Fix the description
 	rootCmd.PersistentFlags().StringVar(&config.SQSQueueName, sqsQueueNameFlag, "", "SQS queue name")
+	rootCmd.PersistentFlags().StringVar(&config.KinesisStreamName, kinesisStreamNameFlag, "", "Kinesis stream name")
```
dbenhur: As in interface decision, I'm not sure I like adding `--stream` as this may proliferate flag names as we consider additional facilities for acquiring events and the distinction between `queue` and `stream` is muddy at best. Perhaps a `--queue-type` which can have values of `sqs` and `kinesis`, defaulting to `sqs` to preserve prior behavior?

jhspaybar: I prefer the URI style approach. I'll go this direction. It also sounds like this is a change that you'd like to accept given it's improved, so I'll add tests and update the pull request. Should I flatten my commits as I go, or only at the end when you're ready to merge?  Not sure the preferred style for this repo.

jhspaybar:
@dbenhur updated, includes tests as well

cluster-state-service/cmd/root.go
narehayrapetyan: Doesn't matter how you treat your commits before merging but we usually squash into one commit (as long as everything belongs together logically) and rebase when merging.

cluster-state-service/handler/event/kinesis_consumer.go
```diff
@@ -0,0 +1,103 @@
...
+func (kinesisConsumer *kinesisEventConsumer) pollForMessages() {
+	log.Infof("In pollForMessages")
+	if kinesisConsumer.iterator == nil {
+		log.Infof("First pass, getting iterator")
+		iteratorRequest := &kinesis.GetShardIteratorInput{
+			ShardId: aws.String("shardId-000000000000"),
+			ShardIteratorType: aws.String("TRIM_HORIZON"),
+			StreamName: aws.String(kinesisConsumer.streamName),
+		}
+		iteratorResponse, err := kinesisConsumer.kinesis.GetShardIterator(iteratorRequest);
```
narehayrapetyan: no ';' necessary ;)

cluster-state-service/handler/event/kinesis_consumer.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+	}
+}
+
+func (kinesisConsumer *kinesisEventConsumer) pollForMessages() {
+	log.Infof("In pollForMessages")
+	if kinesisConsumer.iterator == nil {
+		log.Infof("First pass, getting iterator")
+		iteratorRequest := &kinesis.GetShardIteratorInput{
+			ShardId: aws.String("shardId-000000000000"),
```
narehayrapetyan: where do you configure how many shards you have? Did you choose the name or is it provided by kinesis? Can you move it to a const for now (later we can move it into config)

cluster-state-service/handler/event/kinesis_consumer.go
```diff
@@ -0,0 +1,103 @@
...
+			ShardIteratorType: aws.String("TRIM_HORIZON"),
```
narehayrapetyan: move type to const. AFAIK trim_horizon means read from oldest untrimmed, do we trim it ourselves or does kinesis? How do we handle multiple threads reading the same record? Are we going to rely on etcd to handle duplicate records?

cluster-state-service/handler/event/kinesis_consumer_test.go
```diff
@@ -0,0 +1,139 @@
...
+)
+
+const (
+	streamName       = "test"
+	kinesisMessageBody1 = "messageBody"
+	kinesisMessageBody2 = "messageBody2"
+)
+
+type consumerMockKinesisContext struct {
+	mockCtrl             *gomock.Controller
```
poojamaiya: Nit - The formatting here looks weird. [gofmt](https://atom.io/packages/gofmt) should help with this.

narehayrapetyan: did you run gofmt? the indentation looks off

cluster-state-service/handler/event/kinesis_consumer.go
```diff
@@ -0,0 +1,103 @@
...
+		case <-ctx.Done():
+			return
+		default:
+			kinesisConsumer.pollForMessages()
+		}
+	}
+}
+
+func (kinesisConsumer *kinesisEventConsumer) pollForMessages() {
+	log.Infof("In pollForMessages")
```
poojamaiya: This log message feels like noise. Can we remove it?

cluster-state-service/handler/event/kinesis_consumer.go
poojamaiya: Can we add the shard id as a constant? The iterator type below too.

cluster-state-service/handler/event/kinesis_consumer_test.go
```diff
@@ -0,0 +1,139 @@
...
+	if err == nil {
+		t.Error("Expected an error when stream name is empty")
+	}
+}
+
+func TestPollForKinesisEventsReceiveTwoMessages(t *testing.T) {
+	mockContext := NewConsumerMockKinesisContext(t)
+	defer mockContext.mockCtrl.Finish()
+
+	mockContext.kinesisClient.EXPECT().GetShardIterator(gomock.Eq(mockContext.getShardIteratorInput)).Return(mockContext.getShardIteratorOutput, nil)
```
narehayrapetyan: I'd also add tests verifying that 

* if the iterator is not null this call does not happen
* getRecords throws an error
* getRecords does not return any records

cluster-state-service/handler/event/kinesis_consumer.go
poojamaiya: I'm not very familiar with Kinesis. But do you think using the iterator type `AFTER_SEQUENCE_NUMBER` will be useful if the streaming breaks for some reason and we have to restart from the point after the last sequence number that was read?

cluster-state-service/handler/event/kinesis_consumer.go
```diff
@@ -0,0 +1,103 @@
...
+	if kinesisConsumer.iterator == nil {
```
poojamaiya: Why aren't we doing this within the initialization function `NewKinesisConsumer()`?

cluster-state-service/handler/event/kinesis_consumer.go
```diff
@@ -0,0 +1,103 @@
...
+		iteratorResponse, err := kinesisConsumer.kinesis.GetShardIterator(iteratorRequest);
+		if err != nil {
+			log.Errorf("%+v", errors.Wrapf(err, "Could not get shard iterator"))
+			return
+		}
+		kinesisConsumer.iterator = iteratorResponse.ShardIterator
+	}
+
+	recordsRequest := &kinesis.GetRecordsInput{
+		Limit: aws.Int64(100),
```
poojamaiya: Why 100? Also, can we make this a constant?

cluster-state-service/handler/event/kinesis_consumer_test.go
```diff
@@ -0,0 +1,139 @@
...
+	c, err := NewKinesisConsumer(mockContext.kinesisClient, mockContext.processor, streamName)
+
+	if err != nil {
+		t.Errorf("Unexpected error when calling NewConsumer: %+v", err)
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	pollCount := 0
+
+	mockContext.kinesisClient.EXPECT().GetRecords(mockContext.getRecordsInput).Return(mockContext.getRecordsOutput, nil)
```
poojamaiya: Can we also tests to verify error handling when each of these mocks returns errors?

cluster-state-service/handler/run/run.go
```diff
@@ -92,15 +95,28 @@ func StartClusterStateService(queueName string, bindAddr string, etcdEndpoints [
...
 	processor := event.NewProcessor(stores)
 
-	sqsClient := clients.NewSQSClient(awsSession)
 
-	// start event consumer
-	consumer, err := event.NewConsumer(sqsClient, processor, queueName)
-	if err != nil {
-		return errors.Wrapf(err, "Could not start the consumer")
-	}
+	if strings.HasPrefix(queueNameURI, kinesisPrefix) {
```
jhspaybar: This doesn't seem necessary.  The existing behavior seemed to validate sqs inside the sqs consumer constructor and to maintain backwards behavior, assuming an sqs queue seemed correct.  It seems that just looking for the kinesis prefix otherwise handing off to sqs is the desired behavior.

cluster-state-service/handler/event/kinesis_consumer.go
```diff
@@ -0,0 +1,103 @@
...
+		streamName: streamName,
+		processor: processor,
+	}, nil
+}
+
+func (kinesisConsumer *kinesisEventConsumer) PollForEvents(ctx context.Context) {
+	log.Infof("Starting to poll for events from Kinesis")
+	for {
+		select {
+		case <-ctx.Done():
```
jhspaybar: I thought about this, but there doesn't seem to be an equivalent style API to call on Kinesis.  There is an API to enable advanced monitoring on a stream, but it seems that it's optional and that those metrics would be viewed through cloudwatch metrics APIs instead.

cluster-state-service/handler/event/kinesis_consumer.go
jhspaybar: It seems that shardId-000000000000 is some sort of magic value that you are supposed to start with to go to the beginning of a stream.  I'll turn this into a const for now since it's unlikely someone would be passing an alternate value since shard IDs are only guaranteed valid for something like 300 seconds.

cluster-state-service/handler/event/kinesis_consumer.go
jhspaybar: will move to const, kinesis will trim it based on the policy configured on the stream in kinesis.  Multiple threads(I would assume rather multiple consumers as the consumer in this project seems to be a single go-routing by design) won't cause any issues.  Kinesis allows all consumers to consume at independent paces.  As for duplicates, yes, my expectation is for the data store to no-op the addition if it is older than the current version.  This seems like a pretty hard promise since messages can get re-ordered so I was leveraging it.

cluster-state-service/handler/event/kinesis_consumer_test.go
jhspaybar: nope, will do!

cluster-state-service/handler/event/kinesis_consumer.go
jhspaybar: There are some error scenarios where you are expected to start over reading the stream, so it's possible for the iterator to return to a nil state after these errors occur.

cluster-state-service/handler/event/kinesis_consumer.go
jhspaybar: 100 seemed like a nice number :)  will make it a const though.

jhspaybar:
@narehayrapetyan I have not run the cucumber tests(new to go, didn't even know what those were).  Is there some sort of setup or instruction on what to do?  I didn't see info on this in the CONTRIBUTING.md and in the CONTRIBUTING file it said to run `make` which is what I had been using, so unless this runs cucumber I hadn't done so.

jhspaybar:
Updated code, things still noticeably missing

1) CFN template for kinesis
2) Running of cucumber tests(though I have manually verified things work against a hand made cluster/kinesis queue)
3) A test for the empty message case, if a test takes longer than 1 second it fails, is there a way to override this for a single test?

narehayrapetyan:
Doesn't look like there is a way to override the timeout for individual tests: https://github.com/golang/go/issues/10529. Ideally we don't want to sleep in unit tests. Can you instead make the timeout configurable and pass in 0 in tests?

We decided not to make running integration tests mandatory for PRs: https://github.com/blox/blox/pull/85. We will run them for you before we merge. If you want to run them yourself, however:

* cluster-state-service end-to-end tests: (setup details in cluster-state-service/internal/Readme.md) 
cd cluster-state-service; make e2e-tests

* daemon-scheduler integration tests: (setup details in daemon-scheduler/internal/README.md) 
cd daemon-scheduler; make integ-tests

* daemon-scheduler end-to-end tests: (setup details in daemon-scheduler/internal/README.md)
cd daemon-scheduler; make e2e-tests

jhspaybar:
I'd like to skip CFN for now to close this out. I'll make an issue and play around with making that work and seeing if I can run the cucumber tests now that you have instructions.

narehayrapetyan:
I ran the integration and end-to-end tests with sqs to ensure existing functionality was not broken. We need to also run the tests with kinesis which will be easier to set up once the cloudformation template is done: https://github.com/blox/blox/issues/95

PR #82, opened 2016-12-07
Backend implementation of filtering environments by cluster name

poojamaiya:
#### Summary
We had backend support to filter environments by cluster ARNs. Adding filtering by cluster names.

#### Implementation details
Check if the cluster filter value is an ARN or a name and filter accordingly.

#### Testing

- [X] daemon-scheduler binary built locally and unit-tests pass (`cd daemon-scheduler; make; cd ../`)
- [X] daemon-scheduler build in Docker succeeds (`cd daemon-scheduler; make release; cd ../`)

New tests cover the changes: Yes

#### Licensing


daemon-scheduler/pkg/regex/regex_consts.go
```diff
@@ -0,0 +1,21 @@
...
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package regex
+
+const (
+	clusterNameRegexWithoutStart = "[a-zA-Z][a-zA-Z0-9_-]{1,254}$"
+	ClusterNameRegex             = "^" + clusterNameRegexWithoutStart
+	ClusterARNRegex              = "^(arn:aws:ecs:)([\\-\\w]+):[0-9]{12}:(cluster)/" + clusterNameRegexWithoutStart
```
dbenhur: Why are there capture groups in this regexp that are never used?  Is and will account-id always be exactly 12 digits? Can regions really start with `-`? If this ties to a published spec [eg: aws-arns-and-namespaces](http://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#arn-syntax-ecs), it would be good to link it in a comment.

daemon-scheduler/pkg/regex/validate.go
```diff
@@ -0,0 +1,34 @@
...
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package regex
```
dbenhur: Why name this package after its implementation technique?  Consumer of these functions cares that they validate or recognize particular string forms, not that they use regexps to do it (in fact they shouldn't know that).  Perhaps "validate", "sense", "recognize", or "awsnames" would be a more functional name for the module?

daemon-scheduler/pkg/regex/regex_consts.go
poojamaiya: I pulled this regex off of the error message the ECS gives when you try to create a cluster with an invalid name - `...Cluster name must match ^[a-zA-Z0-9\-_]{1,255}$...`. We've used this same regex in the cluster state manager as well and wanted to keep it consistent.

daemon-scheduler/pkg/regex/validate.go
poojamaiya: I thought about it, but decided to name it 'regex' because we may want to do other things using the regexes other than just validation. For example, in the cluster state service, we use regexes to extract the cluster name from the cluster ARN.

daemon-scheduler/pkg/regex/regex_consts.go
kylbarnes: These regexes look good to me. I agree that the capture groups on ClusterARNRegex are unnecessary and don't seem to be used anywhere. But i'm also a fan of keeping our regexes consistent, so this isn't a worry to me.

poojamaiya:
We had this discussion for the Cluster State Service and decided to allow querying using both cluster names and ARNs. I agree that there is a risk of getting cross account data. But today CSS works with only a single account. With any future changes to allow CSS to work with multiple accounts, we'll have to make changes in the CSS. I assumed we'd want querying by cluster names in the daemon scheduler for consistency. 

>> API layer need to normalize to arns by calling ECS.DescribeCluster api
How does this mitigate the risk of getting cross account data  (assuming we allow multiple account data ingress in the future)? We still have the same info that is stored locally. Plus, I'm not very convinced about using ECS for any data disambiguation. I'd rather not support this feature in that case. What do you think?

narehayrapetyan:
@poojamaiya what do you mean by " I assumed we'd want querying by cluster names in the daemon scheduler for consistency."?

poojamaiya:
@narehayrapetyan - since we allow querying by cluster names in CSS, I thought we'd want to allow the same in the daemon scheduler.

kiranmeduri:
@poojamaiya yes you are right that ECS.Describe will not solve this. I believe other than forcing users to use ARN we cannot avoid cross-account problems. And forcing to use ARN is not good for UX. I am okay with this change, but lets add a tracking issue to solve multiple-account use-case.

Thanks for clarifying

daemon-scheduler/pkg/regex/validate.go
kiranmeduri: I agree with @dbenhur here. It makes sense to have 'validate' use 'regex' but not be part of 'regex'. And 'regex' is a means to get to what you want but not the only way. So in future if we may want to get rid of 'regex' but not 'validate'.

poojamaiya: I'm still not convinced. We use these regexes for more than just validaing. For example, in CSS, we use them to define router paths. It'll be weird to use validate.ClusterARNRegex in the router because we're not validating anything. I'd be happy to change this to something more generic than 'validate'. Not sure I can think of anything though.

poojamaiya:
@kiranmeduri created an [issue](https://github.com/blox/blox/issues/86) for mutiple account use case

daemon-scheduler/pkg/deployment/environment.go
```diff
@@ -161,14 +163,40 @@ func (e environment) FilterEnvironments(ctx context.Context, filterKey string, f
 }
 
 func (e environment) filterEnvironmentsByCluster(ctx context.Context, cluster string) ([]types.Environment, error) {
+	if validate.IsClusterARN(cluster) {
+		return e.filterEnvironmentsByClusterARN(ctx, cluster)
+	}
+	if validate.IsClusterName(cluster) {
+		return e.filterEnvironmentsByClusterName(ctx, cluster)
+	}
+	return nil, errors.Errorf("'%s' is neither is a cluster name nor a cluster ARN", cluster)
```
narehayrapetyan: is neither is -> is neither

daemon-scheduler/pkg/deployment/environment.go
```diff
@@ -161,14 +163,40 @@ func (e environment) FilterEnvironments(ctx context.Context, filterKey string, f
 }
 
 func (e environment) filterEnvironmentsByCluster(ctx context.Context, cluster string) ([]types.Environment, error) {
```
narehayrapetyan: what is the use case for the daemon scheduler to retrieve clusters across environments? We should never be deploying to multiple environments at once.

daemon-scheduler/pkg/deployment/environment.go
poojamaiya: Thanks! fixing.

PR #84, opened 2016-12-07
Fixing service name in logs

narehayrapetyan:
#### Summary

Fixes the service name in a log line

#### Testing

#### Licensing


cluster-state-service/main.go
```diff
@@ -36,15 +36,15 @@ func main() {
 		fmt.Printf("Could not initialize logger: %+v", err)
 	}
 	if err := cmd.RootCmd.Execute(); err != nil {
-		log.Criticalf("Error executing: %v", err)
+		log.Criticalf("Error executing: %+v", err)
```
poojamaiya: As a golang nube, afaik, the '+' adds field names for structs. Whats the advantage here?

narehayrapetyan: it prints out the stack trace: https://dave.cheney.net/2016/06/12/stack-traces-and-the-errors-package. Even though in that article it talks about fmt.Formatter to be honest I haven't tested if seelog prints out the stack trace without the '+' but it doesn't hurt having it there.

PR #85, opened 2016-12-08
Adding integ and end to end test requirements to the PR template

narehayrapetyan:
#### Summary

Modifies the PR template

#### Licensing


kylbarnes:
Should we require e2e and integration tests be run for submitting a PR? These tests require additional ECS resources and performs administrative actions against a live ECS cluster. In my opinion, these should not be a requirement for submitting a PR.

aaithal:
+1 I don't think these should be prereqs for submitting a PR because of the reasons mentioned by @kylbarnes

poojamaiya:
I understand the points raised here. But how would we ensure regression with respect to functionality? Does this mean that till our functionality testing is completely local, we run the tests before merging in any pull requests?

kylbarnes:
We should setup automated CI using Travis or Jenkins, and run the e2e/integration tests on our side whenever pull requests get submitted/merged. That way we control and get billed for the ECS resources, and not require everyone to have a secondary ECS cluster for these tests.

narehayrapetyan:
Before we set up integration testing on our side we cannot accept PRs without making sure these tests have been run. We can either prioritize setting up a pipeline or add these to the template temporarily.

aaithal:
I'd vote for investing the time to build the pipeline as it doesn't seem fair to ask contributors to run these tests, issues being:
* Resource cost
* Setup for tests isn't automated

How about adding a note in the PR that we'll run the tests for them if needed and filling the 'integ-test' section ourselves? The process would be manual until we have the automation built-in, which should motivate us more towards building the automation :)

narehayrapetyan:
@aaithal I agree. I'll modify the template to remind us to run the tests before we merge. We should also double check who's allowed to merge into the dev/master branches.

dbenhur:
I'm in agreement with @kylbarnes and @aaithal . It's not good form to require collaborators to incur AWS charges and manage non-automated setups to run integration tests.  The burden should be shouldered by the Blox team until we come up with a more automatic solution.

Any automated solution for integration testing this kinda system, which requires network access and permission to external resources, is problematic.  If we automatically trigger integ-tests for PR's, we are executing arbitrary externally-submitted code in an environment with network access and access to provisioned resources and credentials -- there's all sorts of opportunity for abuse there, from spam emission and DDOS to simple exploitive consumption of AWS capacity.   We will need to think about how we can sandbox and constrain the integration test environment to mitigate these vulnerabilities.  It's worth exploring how other OSS projects with automated build and test regimes address this.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -23,3 +23,9 @@ for inclusion in changelog.
 #### Licensing
 
 This contribution is under the terms of the Apache 2.0 License: Yes
+
+#### Before merging
+<!-- Run integration and end-to-end tests before merging -->
+- [ ] cluster-state-service end-to-end tests pass. Required setup details in cluster-state-service/internal/Readme.md (`cd cluster-state-service; make e2e-tests; cd ../`)
```
poojamaiya: `make e2e tests` will not pass without setting the right environment variables (`ECS_CLUSTER`, etc.). We should probably mention that or not provide the command at all. Same with the 2 commands below.

narehayrapetyan: I don't want to duplicate the giant setup section from Readme. I think it's enough to redirect them to Readme

poojamaiya: If thats the case, I'd prefer to remove the commands altogether.. What do you think?

narehayrapetyan: sure

PR #90, opened 2016-12-12
Fixing CSS naming in the tests readme file

poojamaiya:
#### Summary
Renaming Event Stream Handler to Cluster State Service

#### Licensing


dbenhur:
Just a note that I find the initialism **CSS** for this service unfortunate, since it's already bound to a much older and widely used technology Cascading Style Sheets.  Perhaps we cold use something distinct so as not to trigger my (and others) cognitive overload. **CluSS** might work.

PR #103, opened 2016-12-15
Supporting instance filtering by both cluster and status

poojamaiya:
#### Summary
Supporting instance filtering by both cluster and status

#### Implementation details
Backend now deals with a map of filter keys to values instead of a single filter key and value.

#### Testing

- [X] cluster-state-service e2e tests succeed (`cd cluster-state-service; ECS_CLUSTER=SchedulerTest ECS_CLUSTER_ASG=scheduler-test AWS_PROFILE=event-handler AWS_REGION=us-east-1 make e2e-tests; cd ../`)
- [X] Manually tests with the curl command "curl -v "http://localhost:3000/v1/instances?status=<status>&cluster=<cluster>"" and the responses are valid
- [X] Also wrote a script and tested the behavior with swagger client and the responses are valid

New tests cover the changes: Only unit tests. Adding integ tests is a little complicated because there is no automation right now and this would mean that we'll require more prerequisite conditions.

Supporting instance filtering by both cluster and status

#### Licensing


cluster-state-service/handler/store/instancestore.go
```diff
@@ -130,19 +134,51 @@ func (instanceStore eventInstanceStore) ListContainerInstances() ([]types.Contai
...
+	filters := make([]string, 0, len(filterMap))
+	for k := range filterMap {
+		filters = append(filters, k)
+	}
+
+	if !instanceStore.areFiltersValid(filters) {
+		return nil, errors.Errorf("At least one of the provided filters '%v' is not supported.", filters)
+	}
+
+	for _, f := range filters {
```
kylbarnes: Could this be simplified as `for k,v := range filterMap` and check if `v == ""`? Do you need both filters and filterMap?

cluster-state-service/handler/store/instancestore.go
```diff
@@ -130,19 +134,51 @@ func (instanceStore eventInstanceStore) ListContainerInstances() ([]types.Contai
...
+			return instanceStore.filterContainerInstancesByStatus(filterValue)
+		case filterKey == instanceClusterFilter:
+			return instanceStore.filterContainerInstancesByCluster(filterValue)
+		}
+	}
+
+	// Currently the only combination of filters that is supported are status & cluster
+	if len(filters) == 2 {
+		status, ok := filterMap[instanceStatusFilter]
+		if !ok {
```
kylbarnes: Could this be simplified as `cluster, ok2 := filterMap[instanceClusterFilter]` and combining the two checks into one `if !ok || !ok2 {`?

cluster-state-service/handler/store/instancestore.go
poojamaiya: Nope, don't need both. Using filterMap is probably cleaner although it may not buy us anything else. Will change.

cluster-state-service/handler/store/instancestore.go
poojamaiya: Changing..

cluster-state-service/handler/store/instancestore.go
```diff
@@ -130,19 +134,48 @@ func (instanceStore eventInstanceStore) ListContainerInstances() ([]types.Contai
...
+		return nil, errors.Errorf("At least one of the provided filters '%v' is not supported.", filters)
+	}
+
+	for key, val := range filterMap {
+		if val == "" {
+			return nil, errors.Errorf("Filter value for filter '%s' is empty", key)
+		}
+	}
+
+	if len(filters) == 1 {
```
kylbarnes: Could these two len(filters) checks be simplified into something like:
```
status, statusFilterExists := filterMap[instanceStatusFilter]
cluster, clusterFilterExists := filterMap[instanceClusterFilter]
switch {
  case statusFilterExists && statusFilterExists:
    return instanceStore.filterContainerInstancesByStatusAndCluster(status, cluster)
  case statusFilterExists:
    return instanceStore.filterContainerInstancesByStatus(filterValue)
  case clusterFilterExists:
    return instanceStore.filterContainerInstancesByCluster(filterValue)
  default:
    return nil, errors.Errorf("Unsupported filter combination '%v'", filters)
}
```

Issue #80, opened 2016-12-05
Generating `license.go` is non-deterministic


samuelkarp:
It looks like generating `license.go` is non-deterministic; I get a different order when running on my Ubuntu laptop than what is checked in here.  `license.sh` does [shell globbing](https://github.com/blox/blox/blob/3f58577/daemon-scheduler/scripts/license.sh#L50) which can differ between machines.

kylbarnes:
Logged as bug. Thanks for finding this.

kylbarnes:
This has been resolved in #107.

PR #107, opened 2016-12-17
Fix bug that causes diff in license.go file when building daemon scheduler

tuedtran:
#### Summary
Fix a bug that causes diff in license.go file when building daemon scheduler.
Related issue: https://github.com/blox/blox/issues/80

#### Implementation details
_ Set environment language to C to keep sorting consistent across different environments.
_ Strip out Windows-styled end-of-line character CRLF when generating license.go.
_ Make generator scripts for both DS and CSS identical.

#### Testing

Fix a bug that causes diff in license.go file when building daemon scheduler.
Related issue: https://github.com/blox/blox/issues/80

#### Licensing


samuelkarp:
> Remove Windows-styled end-of-line character CRLF from LICENSE.txt in package cihub/seelog.

It's a bad practice to change a vendored source file.  It introduces a change from upstream which is not tracked anywhere but here and it will get blown away next time the dependency gets updated.  It would be better to change the generator to strip out CRLF from all the license files instead.

kylbarnes:
I agree with @samuelkarp. We shouldn't change the actual vendor source, as doing a fresh pull would revert our local change. Can you strip out windows carriage returns in our shell script?

samuelkarp:
@tuedtran Your new commit is still manually changing the line-endings, so the next time someone runs `make generate` your changes might be reverted.  Can you adjust the generator itself to pick a consistent line-ending?

tuedtran:
Adjusted the generator to remove Carriage Returns from file as suggested. Also made both generators from CSS and DS identical.

daemon-scheduler/scripts/license.sh
```diff
@@ -52,6 +55,7 @@ for repo in ./../../vendor/gopkg.in/* ./../../vendor/google.golang.org/*; do
 done;
 
 inputfile+="
+***
```
narehayrapetyan: what's this?

licenses/license.go
```diff
@@ -6049,6 +6049,7 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+***
```
narehayrapetyan: here as well

daemon-scheduler/scripts/license.sh
kylbarnes: `***` is the space separator between individual licenses in license.go.

narehayrapetyan: oh I see. Thanks, @kylbarnes

tuedtran:
I've tested it with multiple environments, it worked fine.
Since I can't merge this myself, whoever does please squash these commits before merging.

cluster-state-service/scripts/license.sh
```diff
@@ -76,7 +79,7 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 "
 
-cat << EOF > "${outputfile}"
+cat << EOF | tr -d '\r' > "${outputfile}"
```
dbenhur: -1 for [gratuitous use of cat](http://porkmail.org/era/unix/award.html#cat).

tuedtran: Issue fixed: https://github.com/blox/blox/pull/121

PR #108, opened 2016-12-20
Fix Makefiles to not generate licenses every time it builds.

tuedtran:
#### Summary
Fix Makefiles to not regenerate licenses every time it builds.

#### Implementation details
_ Move regenerate script in DS Makefile from clean task to build task.
_ Remove generate licenses from regenerate script in DS Makefile.
_ Add echo messages.

#### Testing

Fix Makefiles to not regenerate licenses every time it builds.

#### Licensing

### DIFF
--- a/cluster-state-service/Makefile
+++ b/cluster-state-service/Makefile
@@ -29,11 +29,12 @@ build:	$(LOCAL_BINARY)
 
 $(LOCAL_BINARY): $(SOURCES)
 	./scripts/build_binary.sh ./out
-	@echo "Built cluster state service"
+	@echo "Built Cluster State Service"
 
 .PHONY: generate
 generate: generate-models $(SOURCES)
 	PATH="$(ROOT)/scripts:${PATH}" go generate ./licenses/... ./copyright_gen/...
+	@echo "Generate licenses and copyrights"
 
 .PHONY: generate build-in-docker
 build-in-docker:
--- a/cluster-state-service/Makefile
+++ b/cluster-state-service/Makefile
@@ -17,7 +17,7 @@ LOCAL_BINARY=out/cluster-state-service
 ROOT := $(shell pwd)
 
 .PHONY: all
-all: clean build unit-tests
+all: clean generate build unit-tests
 
 .PHONY: generate-models
 generate-models:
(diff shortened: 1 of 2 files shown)


cluster-state-service/Makefile
```diff
@@ -29,11 +29,12 @@ build:	$(LOCAL_BINARY)
 
 $(LOCAL_BINARY): $(SOURCES)
 	./scripts/build_binary.sh ./out
-	@echo "Built cluster state service"
+	@echo "Built Cluster State Service"
 
 .PHONY: generate
 generate: generate-models $(SOURCES)
 	PATH="$(ROOT)/scripts:${PATH}" go generate ./licenses/... ./copyright_gen/...
+	@echo "Generate licenses and copyrights"
```
poojamaiya: Nit - Generated (here and below)

tuedtran:
Nit fixed

kylbarnes:
So the change here is to only generate the licenses file when we manually run `make generate`? What's the logic behind this change? Is it to save time when just running `make`, or are there other benefits as well?

tuedtran:
No other benefits. It saves the build 2 seconds.

narehayrapetyan:
I'm worried that we're adding an additional manual step (run make generate every time you have a new dependency) that's going to be easy to forget, especially as new people join the team.

kylbarnes:
I agree that it's nice to only run this as needed, such as when adding a new dependency. It's a pointless step 99% of the time... But if it only saves 2 seconds during the build time, with no other benefits, I don't really see the point in this change personally. I can see us forgetting to run this manually when adding a new dependency.

tuedtran:
Ok. I will add :generate to :all task.

tuedtran:
I added a new commit yesterday. Does Github send you guys notifications for new commits? Or only for new comments?

daemon-scheduler/Makefile
```diff
@@ -34,6 +34,11 @@ $(LOCAL_BINARY): $(SOURCES)
 	./scripts/build_binary.sh ./out
 	@echo "Built Blox Daemon Scheduler"
 
+$(SWAGGER_GEN): $(SWAGGER_SPEC)
+	./scripts/generate_swagger_artifacts.sh
+	PATH="$(ROOT)/scripts:${PATH}" go generate ./copyright_gen/...
```
kylbarnes: Is this line even needed? This seems to be a duplicate of the line in the `generate` target, and is only called by the `generate` target.

tuedtran:
Unnecessary line removed

PR #112, opened 2016-12-22
Adding pending deployment events and in progress monitor unit tests

narehayrapetyan:
#### Summary

Creates events for pending deployments. The events are not being acted on yet (coming in a diferent PR)

#### Testing

- [x] daemon-scheduler end to end tests pass

New tests cover the changes:  yes

#### Licensing


daemon-scheduler/pkg/engine/dispatcher_test.go
```diff
@@ -30,8 +30,7 @@ import (
 )
 
 const (
-	clusterArn      = "cluster-arn"
-	environmentName = "env-name"
+	clusterArn = "cluster-arn"
```
poojamaiya: Can we change this to clusterARN if it isnt too much work?

daemon-scheduler/pkg/engine/monitor.go
```diff
@@ -78,3 +110,22 @@ func (m monitor) runOnce() error {
 
 	return nil
 }
+
+func (m monitor) runPendingOnce() error {
+	environments, err := m.environment.ListEnvironments(m.ctx)
+	if err != nil {
+		return errors.New("Could not retrieve environments")
```
poojamaiya: Adding context here might be useful.. `Could not retrieve environments while running the pending monitor`.

daemon-scheduler/pkg/engine/monitor_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	mocks "github.com/blox/blox/daemon-scheduler/pkg/mocks"
+	"github.com/blox/blox/daemon-scheduler/pkg/types"
+	"github.com/golang/mock/gomock"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+)
+
+type MonitorTestSuite struct {
```
poojamaiya: Can we also add tests to check the cts.Done() cases for both the monitors?

daemon-scheduler/pkg/engine/monitor.go
```diff
@@ -48,13 +50,18 @@ func NewMonitor(
 	}
 }
 
-func (m monitor) InProgressMonitorLoop() {
-	ticker := time.NewTicker(inProgressMonitorTickerDuration)
+func (m monitor) InProgressMonitorLoop(tickerDuration time.Duration) {
```
poojamaiya: Why this change when you're passing in the same constant in run.go?

narehayrapetyan: because if the ticker duration is hard coded it cannot be modified in unit tests and they end up taking 10 secs per test, which is unnecessary and makes build very long.

daemon-scheduler/pkg/engine/monitor.go
narehayrapetyan: sure

daemon-scheduler/pkg/engine/monitor_test.go
narehayrapetyan: The tests actually do hit ctx.Done() because that's how you break out of the for loop. If they didn't, the for loop would never break and the tests would not finish

PR #119, opened 2016-12-23
Refactoring swagger artifact generation

poojamaiya:
#### Summary
Refactors swagger artifact generation to check in the client along with the software. This way, we don't have to explicitly generate the client for the tests. Also changed the directory holding the generated artifacts from ./handler/v1/api to ./swagger/v1/generated - Addressing issue #115.

#### Implementation details
Check in client (in addition to the models) with the software and change tests to use this client.

#### Testing

- [X] cluster-state-service binary built locally and unit-tests pass (`cd cluster-state-service; make; cd ../`)
- [X] cluster-state-service build in Docker succeeds (`cd cluster-state-service; make release; cd ../`)
- [X] daemon-scheduler binary built locally and unit-tests pass (`cd daemon-scheduler; make; cd ../`)
- [X] daemon-scheduler build in Docker succeeds (`cd daemon-scheduler; make release; cd ../`)

New tests cover the changes: Yes

#### Description for the changelog
Refactoring swagger artifact generation to check in the client along with the software

#### Licensing

This contribution is under the terms of the Apache 2.0 License: Yes

#### Before merging

- [X] cluster-state-service end-to-end tests pass. Required setup details in cluster-state-service/internal/Readme.md
- [X] daemon-scheduler integration tests pass.
- [X] daemon-scheduler end-to-end tests pass.

Issue #116, opened 2016-12-22
Add swagger and API support to filter tasks by started_by


poojamaiya:
Backend implementation is complete, just have to add fronted and swagger support.

poojamaiya:
Completed in #122. Closing.

PR #122, opened 2016-12-23
Adding API support to filter tasks by startedBy

poojamaiya:
#### Summary
Adds API support to filter tasks by startedBy. Addresses part of issue #116. WIll add swagger support once my PR #119 is merged in to avoid rework.

#### Implementation details
Added startedBy as an acceptable filter on the API level. Moving verification of supported filter combination to the backend in order to avoid validations at multiple levels.

#### Testing

- [X] Manually tested using the following curl commands
* curl "http://localhost:3000/v1/tasks?status=stopped&startedBy=someone" -> Invalid filter combination
* curl "http://localhost:3000/v1/tasks?startedBy=someone" -> works as expected

New tests cover the changes: Yes

Adding API support to filter tasks by startedBy

#### Licensing

#### Before merging

- [X] cluster-state-service end-to-end tests pass. Required setup details in cluster-state-service/internal/Readme.md


cluster-state-service/handler/api/v1/taskapis.go
```diff
@@ -24,19 +24,22 @@ import (
...
 
 const (
 	taskARNKey     = "arn"
 	taskClusterKey = "cluster"
 
-	taskStatusFilter  = "status"
-	taskClusterFilter = "cluster"
+	taskStatusFilter    = "status"
+	taskClusterFilter   = "cluster"
+	taskStartedByFilter = "startedBy"
```
dbenhur: I would suggest that requiring mixed case query string parameters is an invitation for error. While the w3c specs say query strings are case-sensitive, URLs are often typed or read and transcribed, so avoiding case-sensitivity is a kindness to the user.

kylbarnes: @dbenhur what are you proposing? I typically see either camel case like this or hyphenated. I'm okay with either way, but we should be consistent. I don't see all lowercase without hyphens very often, if that is your recommendation.

dbenhur: @kylbarnes Off the top of my head, I know Facebook and Yelp uniformly use snake_case for query param keys

dbenhur: Sadly, [Docker](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.24/) isn't consistent in the query param convention (eg: `ps_args` in `/containers/{id}/top` vs `detachKeys` in `/containers/{id}/start`.  [Kubernetes](http://kubernetes.io/docs/api-reference/v1/operations/) consistently uses CamelCase for query parameters and mashedwords for path segments (eg `/api/v1/componentstatuses`).  So, there's no conventional guidance from prominent golang operational tooling.

kylbarnes:
The logic looks good. Verified filtering by startedBy works as expected. We should come to a consensus on URI path and query string naming conventions before merging this in.

cluster-state-service/handler/store/taskstore.go
```diff
@@ -141,43 +141,43 @@ func (taskStore eventTaskStore) FilterTasks(filterMap map[string]string) ([]type
...
 		switch {
-		case statusFilterExists:
+		case status != "":
 			return taskStore.filterTasks(isTaskStatus, status)
-		case startedByFilterExists:
+		case startedBy != "":
 			return taskStore.filterTasks(isTaskStartedBy, startedBy)
-		case clusterFilterExists:
+		case cluster != "":
 			return taskStore.filterTasksByCluster(cluster)
```
narehayrapetyan: can this follow filterTasks(isCluster, ..) as well?

cluster-state-service/handler/store/taskstore.go
```diff
@@ -141,43 +141,43 @@ func (taskStore eventTaskStore) FilterTasks(filterMap map[string]string) ([]type
...
 
 	// Currently the only filter combination that is supported is status & cluster
 	if len(filters) == 2 {
-		if statusFilterExists && clusterFilterExists {
+		if status != "" && cluster != "" {
 			return taskStore.filterTasksByClusterAndStatus(cluster, status)
 		}
 	}
 
 	if len(filters) == 1 {
```
narehayrapetyan: should we add tests for each of the filter value cases?

cluster-state-service/handler/store/taskstore.go
poojamaiya: filterTasks(isSomeFilter, value) lists all tasks using the general ecs/task prefix and does a value comparison for filtering. Filtering by cluster on the other uses the ecs/task/cluster prefix and does no comparison. So we can't really use the same function pointer unless we're ok with listing everything and comparing fields.

cluster-state-service/handler/store/taskstore.go
poojamaiya: What do you mean? I thought we did have all tests..

narehayrapetyan: oh, I just looked at the modified test and it didn't look like it was testing all the filter combinations

PR #136, opened 2017-01-13
Adding deployment worker unit tests

narehayrapetyan:
#### Summary

Adding deployment worker unit tests

#### Testing

#### Licensing

#### Before merging

### DIFF
--- a/daemon-scheduler/pkg/deployment/deployment_worker.go
+++ b/daemon-scheduler/pkg/deployment/deployment_worker.go
@@ -20,6 +20,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/ecs"
 	"github.com/blox/blox/daemon-scheduler/pkg/facade"
 	"github.com/blox/blox/daemon-scheduler/pkg/types"
+	log "github.com/cihub/seelog"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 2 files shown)


daemon-scheduler/pkg/deployment/deployment_worker.go
```diff
@@ -78,38 +78,28 @@ func (d deploymentWorker) UpdateInProgressDeployment(ctx context.Context,
...
-			*updatedDeployment, environment.Name)
-	}
-
 	return updatedDeployment, nil
 }
 
-func (d deploymentWorker) updateDeployment(environment *types.Environment,
-	deployment *types.Deployment) (*types.Deployment, error) {
+func (d deploymentWorker) checkDeploymentProgress(environment *types.Environment,
+	deployment *types.Deployment) (*ecs.DescribeTasksOutput, error) {
```
poojamaiya: Feels weird to me that we're using ecs's types internally here

daemon-scheduler/pkg/deployment/deployment_worker.go
```diff
@@ -78,38 +78,28 @@ func (d deploymentWorker) UpdateInProgressDeployment(ctx context.Context,
...
-		return nil, errors.Wrapf(err, "Error updating the deployment %v in the environment %v",
-			*updatedDeployment, environment.Name)
-	}
-
 	return updatedDeployment, nil
 }
 
-func (d deploymentWorker) updateDeployment(environment *types.Environment,
-	deployment *types.Deployment) (*types.Deployment, error) {
+func (d deploymentWorker) checkDeploymentProgress(environment *types.Environment,
```
poojamaiya: What is this  function doing exactly? It is named `checkDeploymentProgress` but is returning tasks..

daemon-scheduler/pkg/deployment/deployment_worker.go
```diff
@@ -120,6 +110,41 @@ func (d deploymentWorker) updateDeployment(environment *types.Environment,
...
+	}
+
+	// retrieve in-progress again to make sure it has not been updated by another process
+	// TODO: wrap the in-progress check and updateDeployment in a transaction
+	deployment, err = d.deployment.GetInProgressDeployment(ctx, environment.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	if deployment == nil || deployment.ID != updatedDeployment.ID {
```
poojamaiya: Can you remind me what updatedDeployment is?

daemon-scheduler/pkg/deployment/deployment_worker.go
narehayrapetyan: it's checking progress on the deployment's tasks

daemon-scheduler/pkg/deployment/deployment_worker.go
narehayrapetyan: there's a todo to switch to cluster state list tasks so I don't see a point in adding new types temporarily

daemon-scheduler/pkg/deployment/deployment_worker.go
narehayrapetyan: the updated deployment object, status changed to completed, failed and succeeded tasks updated

daemon-scheduler/pkg/deployment/deployment_worker.go
poojamaiya: Can we either add this as a comment or change the function name to `checkDeploymentTasksProgress`?

daemon-scheduler/pkg/deployment/deployment_worker.go
poojamaiya: Can we also add a todo here to replace these types to internal types later then? I'm guessing we wont want to use CSS's types either..

daemon-scheduler/pkg/deployment/deployment_worker.go
poojamaiya: So we don;t do anything with it if the id doesn't match the in-progress deployment? Where then do we log/update the task info?

daemon-scheduler/pkg/deployment/deployment_worker.go
```diff
@@ -134,6 +159,10 @@ func (d deploymentWorker) updateDeployment(environment *types.Environment,
 }
 
 func (d deploymentWorker) deploymentCompleted(tasks []*ecs.Task, failures []*ecs.Failure) bool {
+	if len(tasks) == 0 {
```
poojamaiya: So a deployment can be completed but can have new sub-deployments later? Like when a new instance is added to the cluster?

daemon-scheduler/pkg/deployment/deployment_worker.go
narehayrapetyan: sure I'll change the name

daemon-scheduler/pkg/deployment/deployment_worker.go
narehayrapetyan: Actually, I'm not sure adding additonal types internally is preferable. We only use the types to do internal processing and don't store it anywhere in that format. If we did create internal types we would just be copying the same object into another object. I think we want to be careful not to use these types in interfaces or as our data objects because then we are creating dependencies within our own software that we can't control, ie if we want to change the way we store data or we need to pass more information from one interface to the other, we can't. But we're only using these types to figure out task and failure status.

I think one place where we can change is where we store failed instances in the database struct as ecs.Failures.

daemon-scheduler/pkg/deployment/deployment_worker.go
narehayrapetyan: This is in the case where the in-progress deployment has already been modified. I can add a log statement here.

daemon-scheduler/pkg/deployment/deployment_worker.go
narehayrapetyan: Yes, that's correct

PR #143, opened 2017-01-24
Added versioned streaming to cluster-state-service APIs.

kylbarnes:
#### Summary
Added versioned streaming to cluster-state-service APIs.

#### Implementation details
* Updated all cluster-state-service API response models to include entityVersion and new structure.
* Added new 'entityVersion' query string parameter to both cluster-state-service streaming API methods.

#### Testing

Added versioned streaming to cluster-state-service APIs.

#### Licensing

#### Before merging


cluster-state-service/swagger/v1/swagger.json
```diff
@@ -248,8 +264,30 @@
...
+        "entityVersion": {
+          "type": "string"
+        }
+      }
+    },
     "ContainerInstance": {
       "type": "object",
+      "properties": {
+        "metadata": {
+          "$ref": "#/definitions/Metadata"
```
poojamaiya: Do you think we should keep the metadata object of the container instance independent of that of task? It might be a useful thing to do in case we want to add different metadata information to the two of them. Might be an overkill right niw, not sure!

cluster-state-service/handler/api/v1/translate.go
```diff
@@ -87,50 +88,55 @@ func toContainerInstanceResource(r *types.Resource) *models.ContainerInstanceRes
 	return resource
 }
 
-// ToContainerInstance tranlates a container instance represented by the internal structure (types.ContainerInstance) to it's external representation (models.ContainerInstance)
-func ToContainerInstance(instance types.ContainerInstance) (models.ContainerInstance, error) {
-	err := validateContainerInstance(instance)
+// ToContainerInstance translates a container instance represented by the internal structure (storetypes.ContainerInstanceWrapper) to it's external representation (models.ContainerInstance)
+func ToContainerInstance(instanceWrapper storetypes.ContainerInstanceWrapper) (models.ContainerInstance, error) {
+	err := validateContainerInstance(instanceWrapper.ContainerInstance)
```
poojamaiya: I would add `c := instanceWrapper.ContainerInstance` before this and use `c` everywhere to make the code look cleaner.

cluster-state-service/handler/regex/validate.go
```diff
@@ -50,3 +53,13 @@ func IsInstanceARN(instanceARN string) bool {
 	}
 	return false
 }
+
+// IsEntityVersion validates an entity version as a positive integer
+func IsEntityVersion(entityVersion string) bool {
+	if value, err := strconv.ParseInt(entityVersion, 10, 64); err == nil {
```
poojamaiya: We should probably at least log this error here. We might lose info that might be useful since we're not returning the error.

cluster-state-service/handler/store/datastore.go
```diff
@@ -110,13 +112,25 @@ func (datastore etcdDataStore) Get(key string) (map[string]string, error) {
...
 		return nil, errors.New("Key prefix cannot be empty while streaming data from datastore by prefix")
 	}
 
-	kvChan := make(chan map[string]string) // go routine datastore.stream() handles closing of this channel
-	go datastore.stream(ctx, keyPrefix, kvChan)
+	// If entity version is specified, verify that it is not out of range.
+	if entityVersion != "" {
+		var revision int64
+		var err error
+		if revision, err = strconv.ParseInt(entityVersion, 10, 64); err != nil {
```
poojamaiya: Can we not use regex.IsEntityVersion() here?

cluster-state-service/handler/store/datastore.go
```diff
@@ -110,13 +112,25 @@ func (datastore etcdDataStore) Get(key string) (map[string]string, error) {
...
+			return nil, err
+		}
+		if _, err = datastore.etcdInterface.Get(ctx, keyPrefix, clientv3.WithRev(revision)); err != nil {
```
poojamaiya: Is this an extra read call just to make sure that the revision is valid? How does etcd's stream handle a non-existent/invalid revision? Also, can the revision in etcd be modified between the get here and the stream that follows? If yes, how is that handled?

cluster-state-service/handler/store/datastore.go
```diff
@@ -158,7 +181,17 @@ func (datastore etcdDataStore) stream(ctx context.Context, keyPrefix string, kvC
 			}
 			resetStreamIdleTimer(streamIdleTimer)
 			for _, ev := range event.Events {
-				kv := map[string]string{string(ev.Kv.Key): string(ev.Kv.Value)}
+				// Skip empty events, such as Etcd deletes.
```
poojamaiya: Shouldn't a stream technically notify user of deletes? I'm not sure how yet, but it may be useful info to emit to users. What do you think? Maybe add a TODO to deal with it?

cluster-state-service/handler/store/datastore.go
```diff
@@ -179,15 +212,20 @@ func resetStreamIdleTimer(t *time.Timer) {
...
 	if resp == nil || resp.Kvs == nil {
 		return kv
 	}
 
 	for _, response := range resp.Kvs {
-		kv[string(response.Key)] = string(response.Value)
+		entity := types.Entity{
+			Key: string(response.Key),
+			Value: string(response.Value),
+			Version: strconv.FormatInt(response.ModRevision, 10),
```
poojamaiya: Can you maybe add a comment or an etcd link that explains what the ModRevision is?

cluster-state-service/handler/store/datastore.go
kylbarnes: The reason I can not use regex.IsEntityVersion() here is because I actually need the int returned by the ParseInt call, whereas regex.IsEntityVersion just returns a boolean.

cluster-state-service/handler/store/datastore.go
kylbarnes: The problem is that the value returned tries to get deserialized into a task or container instance object in the calling method. When no value is returned, it couldn't be deserialized, and it was throwing an error and killing the stream. Not sure how we want to handle this. What should the stream send out on this scenario?

cluster-state-service/swagger/v1/swagger.json
kylbarnes: I've gone back and forth on this. Right now, the metadata is the same for both objects, as it's just entityVersion. If you have a strong opinion for separating them out, i'll make different objects for each.

cluster-state-service/handler/store/datastore.go
poojamaiya: I'm not sure.. Maybe introduce a status? Lets add a TODO for this and come back to it. We may want to ignore it like this code is already doing. But lets revisit later.

cluster-state-service/handler/store/datastore.go
poojamaiya: I would add it to regex.extract maybe? Would be nice to keep it all in a single place.

cluster-state-service/swagger/v1/swagger.json
poojamaiya: Nope, a single object is probably fine for now.

kylbarnes: Ok, keeping single Metadata object for now. Should be an easy change in the future if we want.

cluster-state-service/handler/api/v1/instanceapis.go
```diff
@@ -122,7 +125,7 @@ func (instanceAPIs ContainerInstanceAPIs) ListInstances(w http.ResponseWriter, r
 		}
 	}
 
-	var instances []types.ContainerInstance
+	var instances []storetypes.ContainerInstanceWrapper
```
kylbarnes: Good idea, I like those names better. I've changed them to `VersionedContainerInstance` and `VersionedTask`.

cluster-state-service/handler/api/v1/instanceapis.go
```diff
@@ -172,8 +175,24 @@ func (instanceAPIs ContainerInstanceAPIs) StreamInstances(w http.ResponseWriter,
...
+	if entityVersion != "" {
+		if !regex.IsEntityVersion(entityVersion) {
+			http.Error(w, invalidEntityVersionClientErrMsg, http.StatusBadRequest)
+			return
+		}
+	}
+
+	instanceRespChan, err := instanceAPIs.instanceStore.StreamContainerInstances(ctx, entityVersion)
 	if err != nil {
+		if err == rpctypes.ErrCompacted || err == rpctypes.ErrFutureRev {
```
kylbarnes: Much cleaner way of handling this. I've wrapped these errors in a new custom `OutOfRangeEntityVersion` error object, and detected them here.

cluster-state-service/handler/api/v1/instanceapis_test.go
```diff
@@ -462,13 +468,13 @@ func (suite *InstanceAPIsTestSuite) TestListInstancesWithRedundantFilters() {
 }
 
 func (suite *InstanceAPIsTestSuite) TestStreamInstancesReturnsInstances() {
-	instanceRespChan := make(chan storetypes.ContainerInstanceErrorWrapper)
-	suite.instanceStore.EXPECT().StreamContainerInstances(gomock.Any()).Return(instanceRespChan, nil)
+	instanceRespChan := make(chan storetypes.ContainerInstanceWrapper)
+	suite.instanceStore.EXPECT().StreamContainerInstances(gomock.Any(), gomock.Any()).Return(instanceRespChan, nil)
```
kylbarnes: Changed.

cluster-state-service/handler/api/v1/translate.go
kylbarnes: Good idea. Makes the code much more readable. Changed.

cluster-state-service/handler/regex/validate.go
kylbarnes: Logging error here now.

cluster-state-service/handler/store/datastore.go
kylbarnes: Good idea. I've changed to using `regex.GetEntityVersion` on these cases.

cluster-state-service/handler/store/datastore.go
kylbarnes: Yup, I needed a way to make sure the revision being specified has not been compacted and does not exceed the latest revision in Etcd. The Etcd watch stream does not throw a compacted or future revision error message unfortunately. It is possible, although fairly unlikely, that a compaction could happen between this check and the start of the stream. However, the bigger issue is that if you specify a future revision, the stream blocks until that revision is created in the future. I've added a TODO to look into a better way of handling this.

cluster-state-service/handler/store/datastore.go
kylbarnes: Cool. I've added a TODO to revisit.

cluster-state-service/handler/store/datastore.go
kylbarnes: Added a comment about what ModRevision is.

kylbarnes:
Addressed PR comments.

poojamaiya:
LGTM. Let's make the daemon scheduler changes before merging this in tt make sure things don't break.

cluster-state-service/internal/features/e2e_stream_instances.feature
```diff
@@ -4,5 +4,14 @@ Feature: Stream Instances
...
+    And I get instance where the task was started
     Then the stream instances response contains at least 1 instance
-    And the stream instances response contains the cluster where the task was started
\ No newline at end of file
+    And the stream instances response contains the instance where the task was started
+
+  Scenario: Get instance stream with past entity version
+    Given I start 1 task in the ECS cluster
+    And I get instance where the task was started
+    And I stop the 1 tasks in the ECS cluster
```
poojamaiya: Nit: task -> tasks.

Changing `^I stop the (\d+) tasks in the ECS cluster$` to `^I stop the (\d+) task(?:|s) in the ECS cluster$` [here](https://github.com/blox/blox/blob/dev/cluster-state-service/internal/features/e2etasksteps/list_tasks_steps.go#L109) should help.

cluster-state-service/internal/features/e2einstancesteps/get_instance_steps.go
```diff
@@ -66,6 +67,26 @@ func init() {
 		cssContainerInstanceList = append(cssContainerInstanceList, *cssInstance)
 	})
 
+	When(`^I get instance where the task was started$`, func() {
```
poojamaiya: I know this is a get call, but should this be part of the stream_instances_steps.go file? We don't use it in the get tests right?

cluster-state-service/internal/features/e2etasksteps/common_steps.go
```diff
@@ -29,7 +29,7 @@ type Exception struct {
 
 var (
 	// Lists to memorize results required for the subsequent steps
-	ecsTaskList   = []ecs.Task{}
+	EcsTaskList   = []ecs.Task{}
```
poojamaiya: Instead of doing this, let's just add the var ecsTaskList to e2einstancesteps/common_steps.go and use that for instance tests. The idea is to keep the instance and task feature step implementations separate.

kylbarnes: Not sure what you mean. Unfortunately the instance steps are already coupled to the task steps, as you need to start a task to get an instance event in the event stream.

poojamaiya: Ah right. But I'd still create the ecsTaskList variable in e2einstancesteps/common_steps.go. But I dont feel strongly because of the reason you mentioned.

cluster-state-service/internal/features/e2e_stream_instances.feature
kylbarnes: Changed to `task(?:|s)`.

cluster-state-service/internal/features/e2einstancesteps/get_instance_steps.go
kylbarnes: Good point. Moved this to stream_instances_steps.go.

cluster-state-service/internal/features/e2etasksteps/common_steps.go
kylbarnes: It's not just a variable name. I'm using the state assigned to EcsTaskList in the task steps within the instance steps, due to the instance streaming steps being dependent on the task steps. Since the instance/task steps are already coupled, I don't think it's worth the effort to completely refactor this variable usage.

cluster-state-service/internal/features/e2e_stream_instances.feature
poojamaiya: Can we also change task to tasks in the feature definition?

poojamaiya: tasks to task*

poojamaiya:
merge conflict fixes and daemon scheduler changes and this PR will be good to go :)

kylbarnes:
Added daemon-scheduler changes and fixed merge conflicts.

PR #150, opened 2017-01-31
Adding a travis.yml file

narehayrapetyan:
#### Licensing

Validated that syntax is valid using http://lint.travis-ci.org/
Not sure how to test that it will correctly run the commands before pushing.


.travis.yml
```diff
@@ -0,0 +1,10 @@
+language: go
+sudo: false
+go:
+    - 1.7
+install:
+    - cd cluster-state-service; make get-deps; cd ../
+    - cd daemon-scheduler; make get-deps; cd ../
+script:
+    - cd cluster-state-service; make; make release; cd ../
```
aaithal: I don't think `release` is the right target here. It has a dependency on `docker`. `unit-tests` would be the right one I think, also I think you should `&&` instead of `;` so that the exit code is captured correctly.

.travis.yml
```diff
@@ -0,0 +1,10 @@
...
+    - cd daemon-scheduler; make; make release; cd ../
```
aaithal: Same as above. Can you change `release` to `unit-tests`? If `make` already covers that, you wouldn't need the second `make` anyway. Also, why is `cd ../` required here?

.travis.yml
```diff
@@ -0,0 +1,10 @@
+language: go
+sudo: false
+go:
+    - 1.7
+install:
+    - cd cluster-state-service; make get-deps; cd ../
```
aaithal: You'd need to replace `;` with `&&` here as well.

.travis.yml
```diff
@@ -0,0 +1,10 @@
...
+    - cd daemon-scheduler; make get-deps; cd ../
```
aaithal: Do you need to do a `cd ../` here?

samuelkarp:
You probably want to add `go_import_path: github.com/blox/blox` so that this can build on forks as well.

.travis.yml
poojamaiya: Not very familiar with travis. But should this include a target that actually runs the unit tests? I don't think make release runs unit tests.

narehayrapetyan: make runs unit tests, do we need a separate unit-test target? Oh, I didn't know about the exit code, does && actually do a boolean and on the exit code?

.travis.yml
narehayrapetyan: I copied "cd ../" from our PR template, do you know if each line on the script takes you back to root?

narehayrapetyan:
@samuelkarp is it going to autobuild on forks? If you're building on forks don't you want it to build the code in the fork? Or am I misunderstanding what import path does?

.travis.yml
narehayrapetyan: Doesn't make run unit tests?

samuelkarp:
@narehayrapetyan [`go_import_path`](https://docs.travis-ci.com/user/languages/go#Go-Import-Path) sets the directory where the code will be placed during the Travis build.  This ensures that forks are buildable with Travis and makes it so contributors can set up Travis on their forks to ensure that their commits build properly prior to sending a pull request.

.travis.yml
aaithal: As I'd mentioned in the next comment, if make already runs unit tests, then that alone should be sufficient. Thee only reason I don't like invoking `make unit-tests` explicitly is that if the `Makefile` is changed over time so that `all` doesn't depend on `unit-tests`, we'd lose this functionality. 

> does && actually do a boolean and on the exit code?

`&&` helps when commands are being chained in that if there's any error with the command preceding `&&`, anything proceeding it will not be executed. With ';' you'd silently consume it. Also see [this section in travis docs](https://docs.travis-ci.com/user/customizing-the-build#Customizing-the-Build-Step) for an example of the same.

.travis.yml
aaithal: I do not. If each line in that section gets it's own shell, we wouldn't need `cd ../`. Also, the [build matrix](https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix) is something that could be of help here. Here's a [blog](https://lord.io/blog/2014/travis-multiple-subdirs/) that illustrates the same.

.travis.yml
kylbarnes: `make` will run unit tests, but will also build the binary. And make release will build the docker image. Do we care about having Travis verify it can build the binary and docker image, or just run the unit tests?

kylbarnes:
How do you intend to test this before merging? Can you get Travis to run these against your fork, or will it only work on the main Blox repo? What branches will Travis run against? What is the Travis trigger? Does it run on PR creation, merging, every commit in your fork?

narehayrapetyan:
tested in my fork: https://travis-ci.org/narehayrapetyan/blox/builds/197112823

@kylbarnes a build is kicked off anytime a new commit is pushed. If you want to build in your fork, you need to enable it in your fork from settings/services

aaithal:
@narehayrapetyan is it possible to do a negative test as well? A deliberate test failure to ensure that everything's setup properly and that build status is shown as failed when there's actually an issue? A temporary commit that can be reverted should help us quickly validate this.

narehayrapetyan:
@aaithal good idea. Exits with 1: https://travis-ci.org/narehayrapetyan/blox/builds/197125837

aaithal:
Awesome, thanks!

poojamaiya:
I built the code base locally using make build and all the swagger related changes were undone. I know we've merged these changes in already but  @narehayrapetyan can you please verify?

narehayrapetyan:
@poojamaiya you need to update your swagger. Run make get-deps

aaithal:
@poojamaiya might also need to make sure that the swagger executable in your `$PATH` is from `$GOPATH/bin` and not from somewhere else as sometimes that can mess up this process as well.

PR #159, opened 2017-02-05
Updated democli scripts to work with new URIs.

kylbarnes:
#### Summary
Updated democli scripts to work with new URIs

#### Implementation details
Updated the URI and query string parameters in the democli python scripts for the cluster-state-service and daemon-scheduler.

Updated democli scripts to work with new URIs.

#### Licensing


deploy/demo-cli/README.md
```diff
@@ -9,15 +9,19 @@ $ ./blox-list-environments.py --help
 == Blox Demo CLI - List Blox Environments ==
 
 usage: blox-list-environments.py [-h] [--region REGION] [--apigateway]
-                                 [--host HOST]
+                                 [--host HOST] [--environment ENVIRONMENT]
```
poojamaiya: Not sure I understand. ListEnvironments doesn't take environment as a parameter does it?

kylbarnes: This is to get an environment by name. `GET /environments/{name}`.

narehayrapetyan: get environment uses the same cli command as list environments you just provide an environment name?

deploy/demo-cli/blox-create-deployment.py
```diff
@@ -28,7 +28,11 @@ def run_apigateway(params):
 	command = ["cloudformation", "describe-stack-resource", "--stack-name", params.stack, "--logical-resource-id", "ApiResource"]
 	restResource = common.run_shell_command(params.region, command)
 
-	command = ["apigateway", "test-invoke-method", "--rest-api-id", restApi['StackResourceDetail']['PhysicalResourceId'], "--resource-id", restResource['StackResourceDetail']['PhysicalResourceId'], "--http-method", "POST", "--headers", "{}", "--path-with-query-string", "/v1/environments/%s/deployments?deploymentToken=%s" % (params.environment, params.token), "--body", ""]
+	uri = '/v1/environments/%s/deployments' % params.environment
```
narehayrapetyan: did we ever update v1 or are we sticking to it for now even though the APIs have changed?

deploy/demo-cli/README.md
kylbarnes: Correct. I didn't think it was worth it to build a completely new CLI script, when this functionality could be supported by passing in the environment to the list command.

deploy/demo-cli/blox-create-deployment.py
kylbarnes: We are sticking with v1 for now. I think we decided that the v1 APIs are going to be mutable until probably after v0.3.0.

PR #167, opened 2017-02-17
Fixed cluster-state-service reconciler to not delete stopped tasks

kylbarnes:
#### Summary
Fixed cluster-state-service reconciler to not delete stopped tasks.

#### Implementation details
The reconciler is now calling ECS ListTasks for a desired status of both running and stopped.

#### Testing

Fixed cluster-state-service reconciler to not delete stopped tasks.

#### Licensing

#### Before merging


cluster-state-service/handler/reconcile/loader/instanceloader.go
```diff
@@ -124,6 +124,7 @@ func (loader instanceLoader) loadLocalClusterStateFromStore() (clusterARNsToInst
 }
 
 // getContainerInstancesFromECS gets a list of container instances from ECS for the specified cluster.
+// The ECS ListContainerInstances method returns active and draining container instances. It does not return inactive container instances.
```
poojamaiya: Are both returned by default? Or just active? If it's the latter, can we update the comment to clarify it?

cluster-state-service/handler/reconcile/loader/taskloader.go
```diff
@@ -148,6 +153,31 @@ func (loader taskLoader) getTasksFromECS(cluster *string) ([]types.Task, error)
 	return tasks, nil
 }
 
+// getTaskARNsFromECS gets a list of the task ARNs from ECS for both running and stopped tasks.
+// ECS currently stores stopped tasks for about 1 hour.
```
poojamaiya: How do we know this is stored for 1 hour? If we know this by experimentation and ECS does not garantee this anywhere in it's documentation, this note here can get outdated easily.

cluster-state-service/handler/reconcile/loader/taskloader.go
```diff
@@ -148,6 +153,31 @@ func (loader taskLoader) getTasksFromECS(cluster *string) ([]types.Task, error)
...
+func (loader taskLoader) getTaskARNsFromECS(cluster *string) ([]*string, error) {
+	taskARNs := make([]*string, 0)
```
poojamaiya: Out of curiosity, how is this different from `make([]*string)`?

cluster-state-service/handler/reconcile/loader/instanceloader.go
kylbarnes: Active and draining container instances are both returned by default.

cluster-state-service/handler/reconcile/loader/taskloader.go
kylbarnes: The API documentation at http://docs.aws.amazon.com/cli/latest/reference/ecs/list-tasks.html says `Recently-stopped tasks might appear in the returned results. Currently, stopped tasks appear in the returned results for at least one hour`. But I agree, since it is non-deterministic, I have removed the comment.

cluster-state-service/handler/reconcile/loader/taskloader.go
kylbarnes: The make function requires a length parameter when creating slices. It is optional when creating maps or channels.

PR #178, opened 2017-02-28
Automate the setup of CSS E2E test

tuedtran:
#### Summary
Automate the setup of CSS E2E test

#### Implementation details
Automate the setup of the environment for the CSS E2E test
#### Testing

Automate the setup of CSS E2E test

#### Licensing

#### Before merging


cluster-state-service/internal/features/wrappers/ec2_wrapper.go
```diff
@@ -52,6 +54,43 @@ func newAWSEC2Session() *session.Session {
...
+func (slice Images) Less(i int, j int) bool {
+	return *slice[i].CreationDate < *slice[j].CreationDate
+}
+
+func (slice Images) Swap(i int, j int) {
+	slice[i], slice[j] = slice[j], slice[i]
+}
+
+func (EC2Wrapper EC2Wrapper) getLatestECSOptimizedImageID() (string, error) {
+	owner := "amazon"
```
samuelkarp: Please do not do this.  We do not guarantee that the most recently created image published by Amazon with the string "ecs-optimized" is the right image to use.  Relying on this will make your tests brittle.  Hard-coding expected AMI IDs would be a better choice here.

cluster-state-service/internal/features/_test/gucumbertest__.go
```diff
@@ -0,0 +1,27 @@
+
```
samuelkarp: (multiple occurrences) It looks like you're missing license headers on these files.

cluster-state-service/internal/features/wrappers/ec2_wrapper.go
poojamaiya: @samuelkarp Wouldn't we want to run the tests with the latest image always though?

>> We do not guarantee that the most recently created image published by Amazon with the string "ecs-optimized" is the right image to use
What do you mean by 'right' image here?

samuelkarp: What do you mean by "latest"?  Do you mean "highest version number" or "most recently released"?  Those are different things and we do not guarantee that they'll always be the same.

cluster-state-service/internal/features/wrappers/ec2_wrapper.go
```diff
@@ -0,0 +1,151 @@
...
+
+	_, err := ec2Wrapper.client.TerminateInstances(&in)
+	if err != nil {
+		return errors.Wrapf(err, "Failed to terminate EC2 instances with instance IDs '%v'. ", instanceIDs)
+	}
+
+	return nil
+}
+
+func (ec2Warpper EC2Wrapper) DescribeKeyPairs(keyPairName *string) error {
```
poojamaiya: nit: ec2Warpper -> ec2Wrapper
Here and other places

cluster-state-service/internal/features/e2etasksteps/common_steps.go
```diff
@@ -34,61 +35,122 @@ var (
...
-	Before("@task|@stream-instances", func() {
-		clusterName, err := wrappers.GetClusterName()
+		err = createInstanceProfile(iamWrapper)
 		if err != nil {
 			T.Errorf(err.Error())
 			return
 		}
-		err = stopAllTasks(ecsWrapper, clusterName)
+
+		err = ec2Wrapper.DescribeKeyPairs(&testKeyPair)
```
poojamaiya: Does DescribeKeyPairs return a specific error like NotFound or something? We should check the error type before creating a new key pair.

cluster-state-service/internal/features/e2etasksteps/common_steps.go
```diff
@@ -179,3 +233,96 @@ func startNTasks(numTasks int, startedBy string, ecsWrapper wrappers.ECSWrapper)
...
+		"Principal": {
+			"Service": "ec2.amazonaws.com"
+		},
+		"Action": "sts:AssumeRole"
+		}
+		]
+	}`
+
+	err := iamWrapper.GetRole(&roleName)
+	if err != nil {
```
poojamaiya: Same comment as above. Check error type before creating the role.

cluster-state-service/internal/features/e2etasksteps/common_steps.go
```diff
@@ -34,61 +35,122 @@ var (
 	cssTaskList   = []models.Task{}
 	exceptionList = []Exception{}
 
-	taskDefnARN = ""
+	taskDefnARN         = ""
+	testKeyPair         = "E2ETestKeyPair"
+	roleName            = "E2ETestRole"
+	instanceProfileName = "E2ETestInstance"
+	testInstanceIDs     = []*string{}
+	policyArn           = "arn:aws:iam::aws:policy/service-role/AmazonEC2ContainerServiceforEC2Role"
```
poojamaiya: nit: policyArn -> policyARN

cluster-state-service/internal/features/e2etasksteps/common_steps.go
```diff
@@ -179,3 +233,96 @@ func startNTasks(numTasks int, startedBy string, ecsWrapper wrappers.ECSWrapper)
...
+	}
+
+	err = iamWrapper.AttachRolePolicy(&policyArn, &roleName)
+	if err != nil {
+		return err
+	}
+
+	err = iamWrapper.AddRoleToInstanceProfile(&roleName, &instanceProfileName)
+
+	time.Sleep(time.Duration(5) * time.Second)
```
poojamaiya: Why sleep here?

cluster-state-service/internal/features/e2etasksteps/common_steps.go
```diff
@@ -179,3 +233,96 @@ func startNTasks(numTasks int, startedBy string, ecsWrapper wrappers.ECSWrapper)
...
+			return err
+		}
+	}
+
+	err = iamWrapper.AttachRolePolicy(&policyArn, &roleName)
+	if err != nil {
+		return err
+	}
+
+	err = iamWrapper.AddRoleToInstanceProfile(&roleName, &instanceProfileName)
```
poojamaiya: We should act on this error..

cluster-state-service/internal/features/e2etasksteps/common_steps.go
```diff
@@ -179,3 +233,96 @@ func startNTasks(numTasks int, startedBy string, ecsWrapper wrappers.ECSWrapper)
...
+
+	time.Sleep(time.Duration(5) * time.Second)
+
+	return nil
+}
+
+func deleteInstanceProfile(iamWrapper wrappers.IAMWrapper) error {
```
poojamaiya: Just wanted to confirm. Does this return an error if there are running ec2 instances that is using this profile?

cluster-state-service/internal/features/e2etasksteps/common_steps.go
```diff
@@ -179,3 +233,96 @@ func startNTasks(numTasks int, startedBy string, ecsWrapper wrappers.ECSWrapper)
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func terminateAllContainerInstances(ec2Wrapper wrappers.EC2Wrapper, ecsWrapper wrappers.ECSWrapper, clusterName string) error {
+	instanceARNs, err := ecsWrapper.ListContainerInstances(clusterName)
+	if len(instanceARNs) == 0 {
```
poojamaiya: We should first check for err here and return err if it is not nil

cluster-state-service/internal/features/e2etasksteps/common_steps.go
```diff
@@ -179,3 +233,96 @@ func startNTasks(numTasks int, startedBy string, ecsWrapper wrappers.ECSWrapper)
...
+		return nil
+	}
+	err = ecsWrapper.DeregisterContainerInstances(&clusterName, instanceARNs)
+	if err != nil {
+		return errors.Wrapf(err, "Fail to deregister container instances '%v'.", instanceARNs)
```
poojamaiya: nit: Fail -> Failed

cluster-state-service/internal/features/e2etasksteps/common_steps.go
```diff
@@ -179,3 +233,96 @@ func startNTasks(numTasks int, startedBy string, ecsWrapper wrappers.ECSWrapper)
...
+	}
+
+	instanceIDs := make([]*string, len(instanceARNs))
```
poojamaiya: Change this to
`instanceIDs := make([]*string, 0, len(instanceARNs))`
If you default a slice to say length of 10, the first element that gets appended goes into the 11th spot, not the 1st (https://blog.golang.org/go-slices-usage-and-internals).

cluster-state-service/internal/features/e2etasksteps/common_steps.go
```diff
@@ -179,3 +233,96 @@ func startNTasks(numTasks int, startedBy string, ecsWrapper wrappers.ECSWrapper)
...
+	for k, v := range instanceARNs {
+		containerInstance, err := ecsWrapper.DescribeContainerInstance(clusterName, *v)
+		if err != nil {
+			return errors.Wrapf(err, "Fail to describe container instance '%v'.", v)
+		}
+		instanceIDs[k] = containerInstance.Ec2InstanceId
```
poojamaiya: I would change the variable name instanceIDs to ec2InstanceIDs for clarity

cluster-state-service/internal/features/wrappers/ec2_wrapper.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	resp, err := ec2Wrapper.client.RunInstances(&in)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to launch '%d' EC2 instances with "+
+			"container instances registered to ECS cluster with name '%s'. ",
+			*count, *clusterName)
+	}
+
+	instanceIDs := make([]*string, len(resp.Instances))
```
poojamaiya: instanceIDs := make([]*string, 0, len(resp.Instances))

cluster-state-service/internal/features/e2etasksteps/common_steps.go
tuedtran: The role and instance profile need a few seconds before EC2 can find them and launches an instance with them.

cluster-state-service/internal/features/e2etasksteps/common_steps.go
tuedtran: Yes, that would be ErrCodeDeleteConflictException "DeleteConflict"

cluster-state-service/internal/features/e2etasksteps/common_steps.go
poojamaiya: Hmm.. I'd suggest that we add a retry with sleep when launching ec2 instanve nased on the error returned there. What do you think?

cluster-state-service/internal/features/wrappers/ec2_wrapper.go
```diff
@@ -0,0 +1,124 @@
...
+	success := false
+	var err error
+	var resp *ec2.Reservation
+	for i := 0; i < 5; i++ {
+		resp, err = ec2Wrapper.client.RunInstances(&in)
+		if err == nil {
+			success = true
+			break
+		}
+		time.Sleep(1 * time.Second)
```
poojamaiya: We have to make all these magic numbers constants. (I've to do this to the canary too). Either do it in this PR (preferably) or add a TODO and come back to it..

cluster-state-service/internal/features/wrappers/iam_wrapper.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+func NewIAMWrapper() IAMWrapper {
+	awsSession := newAWSIAMSession()
+	return IAMWrapper{
+		client: iam.New(awsSession),
+	}
+}
+
+func newAWSIAMSession() *session.Session {
```
poojamaiya: This piece of code seems duplicated across various aws sessions we have except for the endpoint piece. There may be value in factoring it out.

cluster-state-service/internal/features/e2etasksteps/common_steps.go
```diff
@@ -34,61 +36,108 @@ var (
...
+	ec2Wrapper := wrappers.NewEC2Wrapper()
+	iamWrapper := wrappers.NewIAMWrapper()
 
 	BeforeAll(func() {
-		var err error
+		clusterName := wrappers.GetClusterName()
+
+		err := ecsWrapper.CreateCluster(&clusterName)
+		if err != nil {
+			T.Errorf(err.Error())
```
kylbarnes: I don't think `T` gets initialized until the first test is run. I think `T` will be nil in `BeforeAll`, and probably nil pointer panic on all calls in this function.

cluster-state-service/internal/features/e2etasksteps/common_steps.go
```diff
@@ -179,3 +220,111 @@ func startNTasks(numTasks int, startedBy string, ecsWrapper wrappers.ECSWrapper)
...
+	}
+
+	err = iamWrapper.AttachRolePolicy(&policyARN, &roleName)
+	if err != nil {
+		return err
+	}
+
+	err = iamWrapper.AddRoleToInstanceProfile(&roleName, &instanceProfileName)
+	if err != nil {
+		if awsErr, ok := errors.Cause(err).(awserr.Error); ok && awsErr.Code() != "EntityAlreadyExists" && awsErr.Code() != "LimitExceeded" {
```
tuedtran: When IAM role and profile already existed and is tied to each others. In that case, we can just move on.

cluster-state-service/internal/features/wrappers/ec2_wrapper.go
tuedtran: Agree. But there are a bunch of them over the repo. I'll fix this one and leave the rest for another PR.

cluster-state-service/internal/features/e2etasksteps/common_steps.go
kylbarnes: What are we doing if `err` is not an `awserr.Error` here? Do we need to handle this scenario?

tuedtran: It will print out the error. I don't know what else we can do in that case. It should always work though.

kylbarnes: How will it print out the error? The `if` check here only tests if it is an `awserr.Error`.

cluster-state-service/internal/features/wrappers/ec2_wrapper.go
```diff
@@ -0,0 +1,124 @@
...
+	userData := `#!/bin/bash
+echo ECS_CLUSTER=` + *clusterName + ` >> /etc/ecs/ecs.config
+#cloud-boothook
+#!/bin/sh -x
+echo -e '#!/bin/bash' >> /etc/cron.hourly/shutdowninstance
+echo -e '[ "$(( ` + currentDate + ` - ` + bootTime + ` ))" -gt 3600 ] && /sbin/shutdown -h now' >> /etc/cron.hourly/shutdowninstance
+chmod +x /etc/cron.hourly/shutdowninstance`
+	encodedUserData := base64.StdEncoding.EncodeToString([]byte(userData))
+
+	amiID := "ami-022b9262"
```
kylbarnes: The following URL lists the most up-to-date ECS AMIs by region:
http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-optimized_AMI.html

cluster-state-service/internal/features/e2etasksteps/common_steps.go
```diff
@@ -34,48 +38,109 @@ var (
...
+		if err != nil {
+			fmt.Println(err.Error())
+			os.Exit(1)
 		}
 	})
 
 	Before("@task|@stream-instances", func() {
-		clusterName, err := wrappers.GetClusterName()
+		clusterName := wrappers.GetClusterName()
+		err := stopAllTasks(ecsWrapper, clusterName)
```
kylbarnes: `T` below will not be initialized until after the first test. You probably want to check if `T` is not nil below so that it doesn't panic.

cluster-state-service/internal/features/e2etasksteps/common_steps.go
```diff
@@ -34,48 +38,113 @@ var (
...
 	exceptionList = []Exception{}
 
-	taskDefnARN = ""
+	taskDefnARN         = ""
+	roleName            = "E2ETestRole"
+	instanceProfileName = "E2ETestInstance"
+	policyARN           = "arn:aws:iam::aws:policy/service-role/AmazonEC2ContainerServiceforEC2Role"
 )
 
+const errorCode = 1
```
kylbarnes: I don't see where this variable `errorCode` is used anywhere. I assume it's supposed to be used in the `os.Exit` calls?

tuedtran: Yes, it is. My mistake.

cluster-state-service/internal/features/e2etasksteps/common_steps.go
```diff
@@ -34,48 +38,113 @@ var (
...
 func init() {
 
 	cssWrapper := wrappers.NewCSSWrapper()
 	ecsWrapper := wrappers.NewECSWrapper()
+	ec2Wrapper := wrappers.NewEC2Wrapper()
+	iamWrapper := wrappers.NewIAMWrapper()
 
 	BeforeAll(func() {
-		var err error
+		clusterName := wrappers.GetClusterName()
```
kylbarnes: Can you add a comment above `BeforeAll` that says why you're using `os.Exit` here instead of `T.Errorf`? Maybe something like:

// TODO: Change these os.Exit calls to T.Errorf. Currently unable to do so because T is not initialized until the first test.
// (https://github.com/gucumber/gucumber/issues/28)

cluster-state-service/internal/features/wrappers/utils.go
```diff
@@ -40,3 +67,39 @@ func getECSEndpoint() (string, error) {
...
+	return latestAMIIDByRegion[region], nil
+
+}
+
+// ValidateECSContainerInstance validates that an ECS instance was launched
+// within a cluster 'clusterName'
+func ValidateECSContainerInstance(ecsWrapper ECSWrapper, clusterName string) error {
+	// Takes some time for EC2 instance to come up and for the ECS agent to
+	// register the container instance with ECS.
+	// Retry ECS list call once every minute for 10 minutes
```
kylbarnes: Your comment says `Retry ECS list call once every minute for 10 minutes`, but the sleep below is for 30s. Which one is correct?

cluster-state-service/internal/features/e2etasksteps/common_steps.go
```diff
@@ -179,3 +236,115 @@ func startNTasks(numTasks int, startedBy string, ecsWrapper wrappers.ECSWrapper)
...
+
+	if len(instanceARNs) == 0 {
+		return nil
+	}
+	err = ecsWrapper.DeregisterContainerInstances(&clusterName, instanceARNs)
+	if err != nil {
+		return errors.Wrapf(err, "Failed to deregister container instances '%v'.", instanceARNs)
+	}
+
+	ec2InstanceIDs := make([]*string, len(instanceARNs))
```
poojamaiya: Change this to `ec2InstanceIDs := make([]*string, 0, len(instanceARNs))`

cluster-state-service/internal/features/wrappers/utils.go
```diff
@@ -40,3 +67,39 @@ func getECSEndpoint() (string, error) {
 
 	return endpoint, nil
 }
+
+func GetLatestECSOptimizedAMIID() (string, error) {
+	region := os.Getenv(regionEnvVarName)
+	if region == "" {
+		return "", fmt.Errorf("Empty region. Please specify the AWS region.")
+	}
+	return latestAMIIDByRegion[region], nil
```
poojamaiya: lets do the following to make sure there is nothing weird happening.

amiID, ok := latestAMIIDByRegion[region]
if !ok {
return "", errors.New(...)
}
return amiID, nil

cluster-state-service/internal/features/_test/gucumbertest__.go
```diff
@@ -0,0 +1,40 @@
+// Copyright 2016-2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
kylbarnes: Looks like you committed the gucumber `__.go` files. Can you remove these, as these should not be in source control.

cluster-state-service/internal/features/e2etasksteps/common_steps.go
```diff
@@ -34,48 +38,120 @@ var (
...
 
 func init() {
 
 	cssWrapper := wrappers.NewCSSWrapper()
 	ecsWrapper := wrappers.NewECSWrapper()
+	ec2Wrapper := wrappers.NewEC2Wrapper()
+	iamWrapper := wrappers.NewIAMWrapper()
 
+	// TODO: Change these os.Exit calls to T.Errorf. Currently unable to do so because T is not initialized until the first test.
+	// (gucumber/gucumber#28)
```
kylbarnes: Somehow it removed my full url. Can you prepend these with `https://github.com/` so that it's the absolute url?

cluster-state-service/internal/features/e2etasksteps/common_steps.go
kylbarnes: I'd probably also recommend `ec2InstanceIDs = append(ec2InstanceIDs, containerInstance.Ec2InstanceId)` over `ec2InstanceIDs[k] = containerInstance.Ec2InstanceId` down below if you make this change. Not sure `k` serves much purpose here since you're just appending in order.

cluster-state-service/internal/features/wrappers/utils.go
tuedtran: Fixed. 1 minute is correct.

cluster-state-service/internal/Readme.md
```diff
@@ -5,53 +5,56 @@ These tests call the cluster-state-service (CSS) APIs using the swagger generate
 
 ### What are the assumptions made?
 * CSS server is running locally.
-* An ECS cluster has already been created with at least one Container Instance registered to it. You can specify the cluster name by using the `ECS_CLUSTER` environment variable (This step can be automated in the future versions of the test suite by creating a test cluster and launching an EC2 instance with user data enabling to register itself to the created cluster).
+* The test has automatically set up the environment for you. By default, one EC2 instance is launced with the latest ECS-optimized AMI and without a key pair. The instance is set to auto-terminate itself after 1 hour in case the test fails to clean up. The instance is then registered to an ECS cluster named `E2EETestCluster`.
```
kylbarnes: Nit: `The test will automatically`

`E2EETestCluster` -> `E2ETestCluster`

cluster-state-service/internal/Readme.md
```diff
@@ -5,53 +5,56 @@ These tests call the cluster-state-service (CSS) APIs using the swagger generate
...
 
 ### How to run the test suite?
 From a level above the ./internal directory, use the following commands depending on test suite you want to run.
 
+## Customization
```
kylbarnes: Can you format these as a list by adding a `*` to the beginning of each line?

cluster-state-service/internal/Readme.md
```diff
@@ -5,53 +5,56 @@ These tests call the cluster-state-service (CSS) APIs using the swagger generate
 
 ### What are the assumptions made?
 * CSS server is running locally.
-* An ECS cluster has already been created with at least one Container Instance registered to it. You can specify the cluster name by using the `ECS_CLUSTER` environment variable (This step can be automated in the future versions of the test suite by creating a test cluster and launching an EC2 instance with user data enabling to register itself to the created cluster).
+* The test will automatically set up the environment for you. By default, one EC2 instance is launced with the latest ECS-optimized AMI and without a key pair. The instance is set to auto-terminate itself after 1 hour in case the test fails to clean up. The instance is then registered to an ECS cluster named `E2ETestCluster`.
```
kylbarnes: nit: typo. launced -> launched.

PR #202, opened 2017-07-04
Adding daemon design doc

narehayrapetyan:


ajosuarez (approved):
lgtm.

docs/daemon_design.md
```diff
@@ -0,0 +1,685 @@
...
+            }
+    }
+```
+
+If the controller instance dies in the middle of starting a manager, we need to be able to pick up the work where it was left off and continue. To achieve this we need to save state after each operation so the next worker knows where the work was interrupted, i.e. a workflow. We will use AWS Stepfunctions to create a workflow: state handling and retries will be handled by the stepfunctions service.
+
+##### Monitoring and Lifecycle
+
+We might get into situations where there are multiple managers running for an environment, for example, if a manager cannot send a heartbeat but is still processing deployments. The managers will "lock" the deployment when they pick it up for processing by updating the environment record to "being-processed" using dynamodb conditional writes to avoid race conditions. This will ensure that even if there are multiple managers running for an environment, they don't do duplicate work. The manager might die before releasing the lock. We will need a monitor to go through all deployments and release the lock if it's been held for too long. We also need a monitor to sweep for duplicate managers and clean them up.
```
jhspaybar, line 458: Locks and a background monitor to release them can be a bit tough to operate.  I've had good luck following a "distributed lease" model where a worker attempts to increment a version number with a conditional write along with a TTL.  All other workers start a local timer every time they see the version number increase and only try to acquire it if the timer has expired.  There are a couple libraries out there that make this easy, and I think @Yanacek may have had one under aws-labs before?

docs/daemon_design.md
```diff
@@ -0,0 +1,685 @@
...
+```
+
+## Design
+### Architecture
+![Architectural Diagram](images/diagram.png)
+
+### Service Descriptions
+
+#### Blox Frontend
+Responsible for authentication, authorization, throttling, input validation, user facing API. It interacts with the data service to create environment and deployment objects in the database.
```
jhspaybar, line 218: Does this expose the same API as AWS ECS for these APIs?  It sounds like these components will be used in hosted ECS?  I'm mostly wondering if I host this on my own, or was doing local dev, etc, could I use the same AWS CLI that works against the ECS endpoints?

docs/daemon_design.md
narehayrapetyan: Awesome, thanks for the pointer, Will. I'll check it out.

docs/daemon_design.md
narehayrapetyan: @jhspaybar These will be new APIs but they will be available as part of the AWS SDK. If you were hosting this "locally", i.e. in your account, you will be able to set up the APIs in your API gateway but I'm not sure if you'll be able to override the endpoints in the AWS SDK to point to your API gateway. You might be able to set the endpoint on the sdk to point to a non-aws service but I'm not sure what validation they do and if that will work. Let me get back to you on that.

PR #211, opened 2017-08-01
Make single Blox repo with subprojects for components

wjbuys:
The new repository structure for Blox will be a single Gradle multi-project build. The structure will be something like:

```
blox/blox
├── blox-data (activity implementations for dataservice)
│   └── build.gradle
├── blox-data-api (POJOs and interfaces for dataservice)
│   └── build.gradle
├── blox-frontend (implementations of FE APIs)
│   └── build.gradle
├── blox-frontend-api (FE swagger model)
│   └── build.gradle
├── blox-frontend-client (generated FE client)
│   └── build.gradle
├── blox-integ-tests (end to end integ tests)
│   └── build.gradle
├── blox-scheduling-manager (scheduling manager implementation)
│   └── build.gradle
├── build.gradle
├── README.md (from current repository)
└── settings.gradle
```

This will allow us to easily build the whole stack, and make changes across multiple projects where necessary in the early stages of development. We can potentially split the subprojects out into separate repositories later.

PR #213, opened 2017-08-01
Set up initial project skeleton for managed Blox

wjbuys:
#### Summary
This sets up an empty multi-project Gradle build for Blox (see #211 for details).

#### Implementation details
This is what the project looks like:
```
------------------------------------------------------------
Root project - Blox: Open Source schedulers for Amazon ECS
------------------------------------------------------------

Root project 'blox' - Blox: Open Source schedulers for Amazon ECS
+--- Project ':buildSrc' - Custom build logic for building Blox
+--- Project ':data-model' - Shared POJOs and interfaces for the Blox DataService
+--- Project ':data-service' - Handler implementations for the Blox DataService
+--- Project ':frontend' - Lambda-backed API Gateway API for the Blox Frontend API
+--- Project ':frontend-client' - Auto-generated client for the Blox Frontend API
\--- Project ':frontend-infrastructure' - AWS Cloudformation templates and deployment scripts for the Blox Frontend API
```
#### Testing
I've added a `.travis.yml` that will run `gradle check` in the build, which should succeed (because there's nothing in here yet):

```
$ gradle check

BUILD SUCCESSFUL in 1s
```

#### Licensing


wjbuys:
Hm, should we also copy over the design docs from the current `dev` branch?

build.gradle
```diff
@@ -0,0 +1,3 @@
+group 'com.amazonaws.blox'
+version '0.0-SNAPSHOT'
```
narehayrapetyan: why 0.0?

data-model/build.gradle
```diff
@@ -0,0 +1,3 @@
+description 'Shared POJOs and interfaces for the Blox DataService'
```
narehayrapetyan: data-service-model? it's not the data model which is confusing

data-service/build.gradle
```diff
@@ -0,0 +1,3 @@
+description "Handler implementations for the Blox DataService"
```
narehayrapetyan: I think handler is confusing. Data service implementation covers it imo

narehayrapetyan (changes requested):
I know we talked about pushing this to a different branch but since I moved everything to v0.3 branch already and there's no way to rename branches I think we should just rebase this on dev and push directly to dev. We also won't need to copy things like docs over.

build.gradle
wjbuys: It's totally non-working at the moment. Should I make it 0.1 instead?

.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: java
+before_cache:
```
narehayrapetyan: don't we need to have a build step here?

data-model/build.gradle
wjbuys: Do you have a suggestion? What do we call the packages that contain the DAO POJOs and Java interfaces that define each Lambda microservice?

data-service/build.gradle
wjbuys: Yeah, handler is vague. I was thinking of the entrypoint interface that all Lambda functions implement, i.e. `RequestHandler`. Will change to just "Implementation of the Blox DataService"

.travis.yml
wjbuys: The docs linked in the commit message explain how this works, I'll add a comment inline too.

Basically, travis configures java/gradle builds automatically if the repo contains a `build.gradle` (see https://docs.travis-ci.com/user/languages/java/#Projects-Using-Gradle)

data-model/build.gradle
narehayrapetyan: I was suggesting dataservice-model

build.gradle
narehayrapetyan: Sounds good

wjbuys: Okay, fixed it now

CONTRIBUTING.md
```diff
@@ -0,0 +1,16 @@
...
+
+### Before you start
+If you would like to make a significant change, it's a good idea to first open an issue to discuss it.
+
+### Making the request
+Development takes place against the dev branch of this repository and pull requests should be opened against that branch.
+
+### Testing
+Any contributions should pass all tests.
+You may run all tests by running `make` inside cluster-state-service and daemon-scheduler (requires go, go-swagger to be installed). If you have golang installed, you may run `make get-deps` to install dependencies.
```
narehayrapetyan, line 12: this needs to be removed.

narehayrapetyan (commented):
Other than removing the go makefile comment in the contributing doc, looks good.

narehayrapetyan (approved):

CONTRIBUTING.md
wjbuys: Gotcha, will do

wjbuys:
This has been force-pushed into `blox:dev`.

Edit: Just to explain the force push here, we were faced with two options:
- continue this as a continuous line of history from the current `dev` branch. This would introduce a commit with a huge diff of "delete everything", and would force us to carry around the history of the unrelated GoLang codebase in every (non-shallow) clone. Users pulling the dev branch would immediately get conflicts for all their changes (since we deleted all files).
- create this as a new orphan branch and start from scratch, then force push into `dev` to switch over the history. New clones get the actually-in-development code, with none of the unrelated baggage. The old code is still available in the `v0.3` branch. If anyone has cloned the repo at the `dev` branch and tries to pull, they'll get slightly more cryptic conflicts for all changes.

Neither of these are great options; we opted to go for keeping the history separate, and dealing with the fallout of users' clones of the dev branch no longer being pullable. The solution in either case is for users to switch to the `v0.3` branch in their working copy, and pull from that.

PR #218, opened 2017-08-04
Add the skeleton of an AWS API Gateway + Lambda frontend service

wjbuys:
#### Summary
This adds a Spring MVC/REST web service, running as a Lambda function, fronted by API Gateway. See `docs/frontend.md` for an overview of how this works.

Since this is the first real commit to the repository it also sets up [`spotless`](https://github.com/diffplug/spotless/tree/master/plugin-gradle) for enforcing and applying Java code style according to the [Google Java Styleguide](https://google.github.io/styleguide/javaguide.html). This choice was driven mainly by the need for a style that is easy to automatically apply through build scripts and IDE plugins. See `CONTRIBUTION.md` for details.

#### Implementation details
The main moving parts are:
- Spring controller from which Swagger model is derived: `com.amazonaws.blox.frontend.controllers.EnvironmentController`
- Serverless Application Model template: `frontend-infrastructure/api/template.yml`
- Generated Swagger model: `frontend-service/api/swagger.yml`

There is some custom build logic in `buildSrc` that takes care of generating a Swagger model from the Spring controller, and to add all the necessary extensions to make it work with API Gateway.

I've separated the auto-generated client into a separate commit to make it clear which code is auto-generated for ease of reviewing.

#### Testing
The change includes some basic unit and end-to-end tests, mostly just to test the happy path for integration.

#### Licensing


docs/design.md
```diff
@@ -0,0 +1,33 @@
...
+
+Lambda also supports wiring up multiple API operations to the same lambda function, and then exposes the original HTTP request to the lambda function. Using this approach unlocks some big benefits:
+
+* Our Cloudformation template and Swagger model are simpler. Since all operations are wired up to a single lambda function, there's only one value to wire through to the Swagger model.
+* It's easier to reason about deployments. Since we only have two AWS resources (the API and the Lambda function) we can easily version them for more deterministic deployment.
+* It improves active container reuse. Basically, due to the way lambda routing works, if you have all requests going to a single lambda function, each request is more likely to hit an active container, resulting in lower latency.
+
+### Why use a Servlet implementation in the lambda function?
+Instead of using `aws-serverless-java-container` and using a Servlet implementation, we could some of our own glue code to correctly call the right code in response to incoming API calls. However, using a Servlet implementation that supports writing RESTful controllers has some compelling benefits:
```
arminc: Sentence wrong? "we could some of our own glue code"  -> "we could use/create some of our own glue code"

arminc (commented):
Small documentation mistake

docs/design.md
wjbuys: Looks like copy paste mistake, should be *"we could **write** some of our own glue code"*.

wjbuys: Fixed in latest revision, also moved the file to `frontend_design.md` to make the distinction clear.

CONTRIBUTING.md
```diff
@@ -7,6 +7,9 @@ If you would like to make a significant change, it's a good idea to first open a
 ### Making the request
 Development takes place against the dev branch of this repository and pull requests should be opened against that branch.
 
+### Code Style
+This project follows the [Google Java Styleguide](https://google.github.io/styleguide/javaguide.html), and this style is enforced as part of the `check` task. We recommend you install [the `google-java-format` plugin for your IDE](https://github.com/google/google-java-format), or use the `gradle spotlessApply` task to format code before checking in.
```
narehayrapetyan, line 11: is this or should it be applied as part of build?

README.md
```diff
@@ -9,6 +9,66 @@ Blox is being delivered as a managed service via the Amazon ECS Console, API and
...
+
+```
+./gradlew projects
+```
+
+### Testing
+To run the full unit test suite, run:
+
+```
+./gradlew check
```
narehayrapetyan, line 23: why check instead of unit-tests?

README.md
```diff
@@ -9,6 +9,66 @@ Blox is being delivered as a managed service via the Amazon ECS Console, API and
...
+                "iam:*"
+            ],
+            "Resource":"*"
+        }]
+    }
+
+    ```
+
+  These permissions are pretty broad, so we recommend you use a separate, test account.
+- configure the `default` profile with the AWS credentials for the user you created above
```
narehayrapetyan, line 52: I'm not sure if we should recommend that they change their default user to this, it really depends on what the customer is doing most often.

build.gradle
```diff
@@ -1,3 +1,103 @@
...
+}
+
+ext {
+    // This can be overridden by specifying -PawsProfile=my-other-profile
+    // on the Gradle command line:
+    awsProfile = project.hasProperty("awsProfile") ? project.awsProfile : "default"
+    println("Profile: ${awsProfile}")
+
+    awsCli = "/usr/local/bin/aws"
+    awsRegion = "us-west-2"
```
narehayrapetyan, line 42: should this be configurable?

build.gradle
```diff
@@ -1,3 +1,103 @@
...
+    awsPrefix = System.getenv("USER")
+
+    stackName = "${awsPrefix}-blox-frontend-${awsRegion}"
+    s3BucketName = "${stackName}-${awsProfile}"
```
narehayrapetyan, line 46: should we be tying the name to the local profile name? what if you move your credentials into a different profile or create a different user?

build.gradle
```diff
@@ -1,3 +1,103 @@
...
+    stageName = "Beta"
```
narehayrapetyan, line 47: configurable?

buildSrc/build.gradle
```diff
@@ -1,3 +1,20 @@
...
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    compile gradleApi()
+    compile localGroovy()
+
+    compile 'com.github.kongchen:swagger-maven-plugin:+'
```
narehayrapetyan: is this dependency going to be included in all artifacts? We don't need swagger everywhere do we?

buildSrc/src/main/java/com/amazonaws/blox/swagger/ApiGatewayExtensionsFilter.java
```diff
@@ -0,0 +1,80 @@
...
+   * The template for the Lambda function name to proxy to, in Cloudformation's Fn::Sub syntax.
+   *
+   * <p>A typical format for this is:
+   *
+   * <p>arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${MyLambdaFunction.Arn}/invocations
+   */
+  @Input private String lambdaFunctionArnTemplate;
+
+  public ApiGatewayExtensionsFilter(String lambdaFunctionArnTemplate) {
+    this.lambdaFunctionArnTemplate = lambdaFunctionArnTemplate;
```
narehayrapetyan: @allargsconstructor?

buildSrc/src/main/java/com/amazonaws/blox/swagger/ApiGatewayExtensionsFilter.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  public Map<String, Object> defaultExtensions() {
+    HashMap<String, Object> extensions = new HashMap<String, Object>();
```
narehayrapetyan: HashMap -> Map?

buildSrc/src/main/java/com/amazonaws/blox/tasks/PostProcessCloudformation.java
```diff
@@ -0,0 +1,116 @@
...
+  /** Mapper used to read the source swagger/SAM YAML templates */
+  private ObjectMapper inputMapper = new ObjectMapper(new YAMLFactory());
+  /** Mapper used to produce combined JSON template */
+  private ObjectMapper outputMapper = new ObjectMapper(new JsonFactory());
+
+  /**
+   * The name of the API GW resource in the SAM template into which the swagger template should be
+   * embedded.
+   */
+  @Input private String apiName;
```
narehayrapetyan, line 70: Would it be useful to have an option of passing these in from command line or another script instead of only accepting input from gradle? (Maybe something that can be done later)

buildSrc/src/main/java/com/amazonaws/blox/swagger/ApiGatewayExtensionsFilter.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.Map;
+import org.gradle.api.tasks.Input;
+
+/**
+ * Add API Gateway extensions to all operations of a swagger spec.
+ *
+ * <p>This will add the necessary x-amazon-apigateway-integration section to all operations in a
+ * Swagger spec to correctly proxy all requests to a single lambda function.
+ */
+public class ApiGatewayExtensionsFilter implements SwaggerFilter {
```
narehayrapetyan: Do we need these classes for all projects or just frontend?

docs/frontend_design.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Why?
+### Why a single lambda function?
+The conventional way to use Lambda with API Gateway is to have a separate lambda function for each operation in the API. Each function has to be wired up to its corresponding API operation by ARN through custom Swagger extensions.
+
+Lambda also supports wiring up multiple API operations to the same lambda function, and then exposes the original HTTP request to the lambda function. Using this approach unlocks some big benefits:
+
+* Our Cloudformation template and Swagger model are simpler. Since all operations are wired up to a single lambda function, there's only one value to wire through to the Swagger model.
+* It's easier to reason about deployments. Since we only have two AWS resources (the API and the Lambda function) we can easily version them for more deterministic deployment.
+* It improves active container reuse. Basically, due to the way lambda routing works, if you have all requests going to a single lambda function, each request is more likely to hit an active container, resulting in lower latency.
```
narehayrapetyan: It's not because of routing is it? It's because containers are shut down if they don't get requests and more code in the same container means it will be hit more frequently.

frontend-infrastructure/build.gradle
```diff
@@ -1 +1,134 @@
...
+        classpath 'io.swagger:swagger-models:+'
+    }
+}
+
+ext {
+    templateFile = file("cloudformation/template.yml")
+    processedTemplateFile = file("${buildDir}/template.json")
+    outputTemplateFile = file("${buildDir}/template.output.json")
+
+    stackOutputsFile = file("${buildDir}/${stackName}-${awsProfile}.outputs.json")
```
narehayrapetyan, line 23: Again, I don't think we should be including the profile name in the file name. What's the advantage of doing that?

frontend-service/src/main/java/com/amazonaws/blox/frontend/LambdaHandler.java
```diff
@@ -0,0 +1,65 @@
...
+import com.amazonaws.serverless.proxy.internal.servlet.AwsHttpServletResponse;
+import com.amazonaws.serverless.proxy.internal.servlet.AwsProxyHttpServletRequest;
+import com.amazonaws.serverless.proxy.spring.SpringLambdaContainerHandler;
+import com.amazonaws.services.lambda.runtime.Context;
+import com.amazonaws.services.lambda.runtime.RequestHandler;
+import org.apache.log4j.Logger;
+
+/** Entrypoint for mapping incoming API GW requests to Lambda into Spring controllers */
+public class LambdaHandler implements RequestHandler<AwsProxyRequest, AwsProxyResponse> {
+  Logger log = Logger.getLogger(LambdaHandler.class);
```
narehayrapetyan: Lombok has logger annotations we can use to skip this line everywhere

frontend-service/src/main/java/com/amazonaws/blox/frontend/models/Environment.java
```diff
@@ -0,0 +1,27 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.blox.frontend.models;
+
+public final class Environment {
+  private final String name;
+
+  public final String getName() {
```
narehayrapetyan: getter and constructor can be replaced with @value

frontend-service/src/main/java/com/amazonaws/blox/frontend/LambdaHandler.java
wjbuys: Oh, neat! I didn't add a lombok dependency to the main project yet, so it's not available in this code at the moment. Will add.

frontend-service/src/main/java/com/amazonaws/blox/frontend/models/Environment.java
wjbuys: Same as above, we didn't have lombok in the project when I started writing this, will update.

buildSrc/src/main/java/com/amazonaws/blox/tasks/PostProcessCloudformation.java
wjbuys: I think later. Since it's tightly tied to the names checked in in the template file itself, I don't think it's particularly useful to override it from the command line rather than source code.

buildSrc/src/main/java/com/amazonaws/blox/swagger/ApiGatewayExtensionsFilter.java
wjbuys: Whoops...

buildSrc/src/main/java/com/amazonaws/blox/swagger/ApiGatewayExtensionsFilter.java
wjbuys: Oh right, missed this, will update.

buildSrc/build.gradle
wjbuys: No, as indicated by its location in `buildSrc`, it's a compile-dependency of our build logic. It will never be included in any artifacts.

buildSrc/src/main/java/com/amazonaws/blox/swagger/ApiGatewayExtensionsFilter.java
wjbuys: We only need them for the frontend, but they do no harm in the other projects. The reason that these are shared, is that Gradle is a bit inflexible in how it supports code that should be available in the build script itself. According to [the docs](https://docs.gradle.org/current/userguide/custom_tasks.html#sec:packaging_a_task_class), the only options are:
1. embed it in the script itself
2. put it in the `buildSrc` subproject, of which there can only be one per root-project
3. put it in a completely separate project, and consume it through a Maven repo

I think once we're comfortable that we are done iterating on this logic, we can move these from (2) to (3). However, since (3) implies making a separate repository and pushing to a Maven repo, I'd like to punt on it for now. I'll add a TODO to the code to that effect?

build.gradle
wjbuys: Can we quickly whiteboard out how we want to make this configurable? As mentioned in another comment, all this AWS settings are just super draft to make it work.

build.gradle
wjbuys: Open to suggestions; the intent behind this was for my own purposes of testing stuff in my personal AWS account and the blox beta account, and this was the easiest way to namespace things at the time. Unless I'm missing something, I don't think we can actually programmatically get at a more fine-grained value to namespace on, so that would have to be yet another configuration setting.

build.gradle
wjbuys: Perhaps. I'm not sure how yet, though. Maybe wait until we see how we may want to reconfigure this?

(Gradle's only configuration mechanism is pretty cumbersome, you have to include `-pKey=Value` on the command line, there doesn't seem to be an easy way to set persistent local configuration).

README.md
wjbuys: Good point. What do you suggest instead? We have to default to some profile name when not provided...

README.md
wjbuys: That's the Gradle idiom, and the default task that is provided by the Java plugin: https://docs.gradle.org/3.3/userguide/java_plugin.html#sec:java_tasks

I figured it was easier to exclude the end-to-end tests from the default task run by CI (see the `build.gradle` in the `end-to-end-tests` project), instead of reconfiguring every single unit test task to have a different name.

CONTRIBUTING.md
wjbuys: The `spotless` plugin will verify that all files meet the configured style, but it will not modify source files without explicitly being requested through the `spotlessApply` task. I think it should not modify source as part of a normal build, to avoid problems with a CI build potentially building code that is different from what was checked in.

wjbuys:
I'll be working in your feedback on the small-level code stuff; do you have any concerns over the larger project structure/design so far? Want to sanity check that this is the shape we're aiming for, before we build too much on top of it.

frontend-infrastructure/build.gradle
wjbuys: See comment above about the purpose of the profile flag. This was merely the quickest fix, I'm open to any better suggestions.

docs/frontend_design.md
wjbuys: I'll update this to reflect that, and refer to the appropriate public docs.

narehayrapetyan:
looks fine overall

wjbuys:
Okay, as discussed in person, I've opened a separate issue for the testing stack naming/configuration stuff related to the AWS properties in the build script: #228

build.gradle
wjbuys: I take that back, I totally missed [the section on `gradle.properties`](https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_properties_and_system_properties) in the docs. It looks like it should be pretty easy to put any necessary config there. As mentioned in comment below, I'll do that in a followup CR.

CONTRIBUTING.md
narehayrapetyan: How do we ensure that the style has been applied? Or are you saying that the check will run as part of the build and fail if the style is not preserved but it won't apply fixes?

README.md
narehayrapetyan: As we talked offline, I don't think we should rely on profile name at all since that is unrelated to any resources and also can change without affecting resources.

CONTRIBUTING.md
wjbuys: Correct.

README.md
wjbuys: Sure, and as mentioned offline, I'll fix in a followup CR for #228

buildSrc/src/main/java/com/amazonaws/blox/swagger/ApiGatewayExtensionsFilter.java
narehayrapetyan: I was mostly concerned about code size because we have to keep the jar as small as possible for Lamdba. What do you think? I think a separate Maven repo is overkill, however.

wjbuys:
Are there any specific changes you'd still like on this CR that aren't covered by #228?

end-to-end-tests/src/test/java/com/amazonaws/blox/EnvironmentTest.java
```diff
@@ -0,0 +1,52 @@
...
+import com.amazonaws.blox.model.DescribeEnvironmentResult;
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.io.File;
+import java.net.URL;
+import org.junit.Test;
+
+public class EnvironmentTest {
+  Blox client = Blox.builder().endpoint(endpoint()).build();
```
narehayrapetyan: private?

frontend-service-client/src/main/java/com/amazonaws/blox/AbstractBlox.java
```diff
@@ -0,0 +1,31 @@
+/**
+
```
narehayrapetyan, line 2: Why is this empty?

narehayrapetyan: Oh is this generated code?

narehayrapetyan (approved):

frontend-service-client/src/main/java/com/amazonaws/blox/AbstractBlox.java
wjbuys: Yup. I'll follow up and see if we can specify a license block or something in the API call that generates this, but I don't remember anything like that...

end-to-end-tests/src/test/java/com/amazonaws/blox/EnvironmentTest.java
wjbuys: Must be public to be a runnable test, I believe.

wjbuys: Ooooh, you mean the `client` var. Sorry, can't see exactly which line you commented on :/ Will fix.

PR #226, opened 2017-08-09
Adding dataservice createEnvironment and startDeployment model and op…

narehayrapetyan:
…eration interfaces

#### Summary

Adds models and operation interfaces for the dataservice client.

#### Licensing


settings.gradle
```diff
@@ -5,5 +5,5 @@ include 'buildSrc'
 include 'frontend-service-client'
 include 'frontend-service'
 include 'frontend-infrastructure'
-include 'data-service-model'
+include 'data-service-client'
```
wjbuys: Oh, this explains the Client class above. Why the change from model to client here? I thought the idea was to keep the model objects/DAOs in a package on their own, so that the server won't have to depend on the code the implements the client, and vice versa.

data-service-client/src/main/java/com/amazonaws/blox/dataservice/client/v1/model/CreateEnvironmentResponse.java
```diff
@@ -0,0 +1,13 @@
...
+import lombok.Builder;
+import lombok.NonNull;
+import lombok.Value;
+
+@Value
+@Builder
+public class CreateEnvironmentResponse {
+
+    @NonNull
+    private String environmentVersion;
```
wjbuys: It looks like most of the create APIs on ECS returns the entire created resource, rather than just an identifier for it (e.g. [createCluster](http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_CreateCluster.html#API_CreateCluster_ResponseSyntax)). Should we be following the same kind of pattern here?

data-service-client/src/main/java/com/amazonaws/blox/dataservice/client/v1/model/InstanceGroup.java
```diff
@@ -0,0 +1,17 @@
...
+import java.util.List;
+
+@Value
+@Builder
+public class InstanceGroup {
+
+    @NonNull
+    private String clusterArn;
+
+    private List<Attribute> attributes;
```
wjbuys: Is attribute ordering or allowing duplicates important? Otherwise a Map or Set might be a better fit here.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Tue Aug 08 15:57:04 PDT 2017
```
wjbuys, line 1: This file will conflict with #218.

data-service-client/src/main/java/com/amazonaws/blox/dataservice/client/v1/client/DataServiceLambdaClient.java
```diff
@@ -0,0 +1,9 @@
+package com.amazonaws.blox.dataservice.client.v1.client;
+
+import lombok.AllArgsConstructor;
+
+@AllArgsConstructor
+public class DataServiceLambdaClient {
```
wjbuys: In this package, I would expect this to be an interface (i.e. `public interface DataService`), and not an implementation. The implementation could then be in a separate client package.

build.gradle
```diff
@@ -1,3 +1,15 @@
 group 'com.amazonaws.blox'
 version '0.1-SNAPSHOT'
 description "Blox: Open Source schedulers for Amazon ECS"
+
+allprojects {
```
wjbuys: This is also going to conflict a bit with #218.

wjbuys (commented):
I think you may want to rebase this on #218, since this adds/modifies some files that have already been added to that PR.

gradle/wrapper/gradle-wrapper.properties
narehayrapetyan: I'll rebase it once it's checked in so I don't have to keep rebasing while you're making changes.

settings.gradle
narehayrapetyan: Yea we have two options: have two packages - one for the model and one for the client or put them both in the same package. Here I was going for the same package. The advantage of separating them would be the dataservice won't have the client file. Either way is fine by me at this point. You have a preference?

wjbuys: Since we're dealing with Lambda functions with JAR size limitations, I think separating these out would be good starting point, especially if we use some third-party libraries for the client. We can always glue them back together if necessary.

data-service-client/src/main/java/com/amazonaws/blox/dataservice/client/v1/model/CreateEnvironmentResponse.java
narehayrapetyan: sounds good

data-service-client/src/main/java/com/amazonaws/blox/dataservice/client/v1/model/InstanceGroup.java
narehayrapetyan: good idea

wjbuys (approved):
LGTM, shipit!

Issue #224, opened 2017-08-07
Add IAM authentication to frontend


wjbuys:
The Frontend Service should use IAM for authentication.

This is configured through a custom Swagger security definition, as outlined in [the docs](http://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-authtype.html).

There's one additional requirement: when hosted in the ECS Control Plane, we'll need to use a slightly different auth configuration in order to make the auth work just like a normal AWS service. In particular, the following should be configurable:

- the `authtype` property of the security definition
- the swagger extensions of the Info section of the swagger model

wjbuys:
Should be fixed by #229.

PR #229, opened 2017-08-09
Add basic support for IAM authentication on all frontend calls (#224)

wjbuys:
#### Summary
This adds another Swagger filter that inserts the necessary API Gateway
swagger extensions to enable IAM Authentication. See #224 for details.

#### Testing
```
gradle check
gradle testEndToEnd
```

#### Licensing

### DIFF
--- a/buildSrc/src/main/java/com/amazonaws/blox/swagger/ApiGatewaySecurityFilter.java
+++ b/buildSrc/src/main/java/com/amazonaws/blox/swagger/ApiGatewaySecurityFilter.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may
+ * not use this file except in compliance with the License. A copy of the
+ * License is located at
+ *
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "LICENSE" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.blox.swagger;
+
+import io.swagger.models.Operation;
+import io.swagger.models.Path;
+import io.swagger.models.Swagger;
+import io.swagger.models.auth.ApiKeyAuthDefinition;
+import io.swagger.models.auth.In;
+import java.util.Collections;
+import lombok.Getter;
+import lombok.Setter;
+import org.gradle.api.tasks.Input;
+
+@Getter
+@Setter
+/**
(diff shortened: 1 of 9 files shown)


buildSrc/src/main/java/com/amazonaws/blox/swagger/ApiGatewaySecurityFilter.java
```diff
@@ -0,0 +1,59 @@
...
+import io.swagger.models.Path;
+import io.swagger.models.Swagger;
+import io.swagger.models.auth.ApiKeyAuthDefinition;
+import io.swagger.models.auth.In;
+import java.util.Collections;
+import lombok.Getter;
+import lombok.Setter;
+import org.gradle.api.tasks.Input;
+
+@Getter
```
narehayrapetyan, line 27: getter and setter can be replaced with data

frontend-service-client/src/main/java/com/amazonaws/blox/BloxClientBuilder.java
```diff
@@ -33,6 +35,53 @@
     }
 
     /**
+     * Specify an implementation of {@link AWSCredentialsProvider} to be used when signing IAM auth'd requests
+     *
+     * @param iamCredentials
+     *        the credential provider
+     */
+    @Override
+    public void setIamCredentials(AWSCredentialsProvider iamCredentials) {
```
narehayrapetyan, line 44: is the FE client not autogenerated?

narehayrapetyan (commented):
It's not clear to me, are some of these files autogenerated?

frontend-service-client/src/main/java/com/amazonaws/blox/BloxClientBuilder.java
wjbuys: Yes, it is. I forgot to pull it into a separate commit this time; I can do that if it's helpful?

buildSrc/src/main/java/com/amazonaws/blox/swagger/ApiGatewaySecurityFilter.java
wjbuys: I'm a bit wary of using that, since this isn't acutally a POJO/Bean. For example, that will also generate a constructor, `equals`, `hashCode`, and `toString`, which might not make sense for this object.

Do you think that condensing these annotations is worth adding those unused features to this object? If it enhances readability/understandability, then I'd be all for it.

narehayrapetyan: that's fine

frontend-service-client/src/main/java/com/amazonaws/blox/BloxClientBuilder.java
narehayrapetyan: That's okay. Just trying to figure out what I need to review and it looks like everything in fe client is auto generated right?

narehayrapetyan (approved):

PR #228, opened 2017-08-15
Figure out how to namespace test stacks

wjbuys:
There was some discussion on #218 about how we namespace our test stacks, and how that integrates with the build logic that acts on AWS resources.

At the moment, it uses the name of the AWS CLI credentials profile to determine the stack name.

What we need from test stacks are:

1. It should be possible to have multiple, independent test stacks within a single account (e.g. I'm working on one feature, and want to test another unrelated feature pending PR, or two developers want to test independent code changes in stacks in the same whitelisted internal account)
2. It should be possible to have separate test stacks in separate accounts, and have developers switch between these by using a Gradle configuration property (e.g. when testing internal vs external stacks)
3. It should be trivial to clean up a test stack completely.

There's 3 main resources for each stack:
- Cloudformation Stack
- S3 Bucket for deployment
- Local files that store stack properties

Of these, S3 bucket names have to be globally unique, and so should be namespaced by stack.

Finally, we need to tie a set of AWS credentials to each stack. The easiest way to do that is through a credential profile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,41 @@ To run the full unit test suite, run:
 This will run the same tests that we run in the Travis CI build.
 
 ### Deploying
-To deploy a personal stack:
+First, take a look at what Blox will put in your personal stack by running the
+`showStackConfig` task:
+
+```
+$ ./gradlew showStackConfig
+
+> Task :showStackConfig
+Blox deployment stack configuration:
+
+  Default resource name         (blox.name): blox-buys-alpha-us-west-2 (default)
+  API Gateway stage            (blox.stage): alpha (default)
+  Stack prefix                (blox.prefix): buys-alpha (default)
+  AWS Region                  (blox.region): us-west-2 (default)
+  AWS Credential Profile     (blox.profile): blox-buys-alpha-us-west-2 (default)
+  Cloudformation stack name (blox.cfnStack): blox-buys-alpha-us-west-2 (default)
+  Deployment S3 bucket name (blox.s3Bucket): blox-buys-alpha-us-west-2 (default)
+
+To customize these values, modify ~/.gradle/gradle.properties to override the property listed.
+
+AWS CLI configuration for profile blox-buys-alpha-us-west-2:
+
+The config profile (blox-buys-alpha-us-west-2) could not be found
+```
+
+If you wish to customize any of these values, you can do so by overriding the
+property in parentheses using [any of the supported ways to override Gradle
+properties](https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_properties_and_system_properties).
+The easiest way is to override it for your user in `~/.gradle/gradle.properties`:
+
+```
+blox.profile=default
+blox.region=us-east-1
+```
+
+Next, in order to deploy your personal stack:
 - install the [official AWS CLI](https://aws.amazon.com/cli/)
 - create an IAM user with the following permissions:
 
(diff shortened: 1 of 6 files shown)


frontend-service-client/src/main/java/com/amazonaws/blox/model/transform/DescribeEnvironmentRequestProtocolMarshaller.java
```diff
@@ -22,7 +22,7 @@
 @SdkInternalApi
 public class DescribeEnvironmentRequestProtocolMarshaller implements Marshaller<Request<DescribeEnvironmentRequest>, DescribeEnvironmentRequest> {
 
-    private static final OperationInfo SDK_OPERATION_BINDING = OperationInfo.builder().protocol(Protocol.API_GATEWAY).requestUri("/Beta/environments/{name}")
+    private static final OperationInfo SDK_OPERATION_BINDING = OperationInfo.builder().protocol(Protocol.API_GATEWAY).requestUri("/alpha/environments/{name}")
```
wjbuys, line 25: This is an autogenerated change; I changed the stage name to `alpha`, to match the profile name. This brings up the concern that the stage name is hardcoded in the client, I'll reach out to the SDK team to see if we can make this configurable instead.

narehayrapetyan:
Before looking at the code, it seems like there is a lot of customization to be able to set this up in your account? Why would we need separate test stacks for different features? The difference in resources should be minimal shouldn't it? At most, you should be adding a new resource in a new feature. 

Also, does this actually run the cloudformation stack to set up resources in your account?

README.md
```diff
@@ -26,7 +26,41 @@ To run the full unit test suite, run:
...
+First, take a look at what Blox will put in your personal stack by running the
+`showStackConfig` task:
+
+```
+$ ./gradlew showStackConfig
+
+> Task :showStackConfig
+Blox deployment stack configuration:
+
+  Default resource name         (blox.name): blox-buys-alpha-us-west-2 (default)
```
narehayrapetyan, line 38: how is this different from cloudformation stack name?

narehayrapetyan, line 38: the defaults are hardcoded to your name?

wjbuys: No, the defaults use `$USER`. I had to use something in the output, and I just copy/pasted these in a hurry. I can replace with `<username>` or such, if this is confusing.

wjbuys: It's the default for all resource names. We can change the default and the stack name independently. Not sure if it's necessary, but considering we're not tightly coupling any resource names (very particularly AWS Profile) anymore, I figured keeping this decoupled too wouldn't hurt.

wjbuys:
> Before looking at the code, it seems like there is a lot of customization to be able to set this up in your account?

The only thing needed is to either create a new AWS Profile named the same as the stack output with the right credentials, or to set up `blox.profile` to point to the one you want to use. Is that what you're referring to?

> Why would we need separate test stacks for different features?

Why wouldn't we? This flexibility isn't costing us anything right now.

> Also, does this actually run the cloudformation stack to set up resources in your account?

No, changes are only made once you run the `createBucket` or `deploy` tasks.

narehayrapetyan:
> Why would we need separate test stacks for different features?
>> Why wouldn't we? This flexibility isn't costing us anything right now.

I'm just trying to figure out what you're proposing our development process should be like. Do we set up a new stack every time we're working on a feature? How do we ensure that the test stack is exactly the same as the beta/gamma/prod stacks? How do we test the new resources end to end before pushing them to beta?

wjbuys:
This PR doesn't try to prescribe anything about our development workflow. It specifically makes no choices, so that we can figure this out without having to come back and change this.

In practice, I suspect that we'll just use the defaults most of the time. However, I've had one or two occasions to override at least `blox.profile` to do some alternate testing on a different branch in a different account.

narehayrapetyan (approved):

PR #230, opened 2017-08-15
Update gradle wrapper

wjbuys:
It looks like gradle will occasionally update the wrapper while running
other tasks. This checks in the latest wrapper changes so that it
doesn't keep showing up in other diffs.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Tue Aug 08 15:57:04 PDT 2017
+#Tue Aug 15 12:10:52 PDT 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.0.2-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.0.2-bin.zip
```
narehayrapetyan: this isn't an update, this is just switching from the all distro to bin. Any reason why we should do that?

wjbuys: Oh, hm. This is what gets generated by running `./gradlew wrapper`. I think I may have manually updated the distro before, and incorrectly updated this URL. Either way, this is the diff that keeps showing up. I'll amend the commit message to reflect the actual change taking place.

narehayrapetyan: I think that's because I have all and you have bin.

wjbuys: Oh hang on, just found this in [the docs](https://docs.gradle.org/current/userguide/gradle_wrapper.html):

> By default, the Wrapper will use a bin distribution. This is the smallest Gradle distribution. Some tools, like Android Studio and Intellij IDEA, provide additional context information when used with the all distribution. 

Sounds like `all` is what we actually want. I'll update this PR to correctly configure the wrapper task for the `all` distribution instead. Good catch!

narehayrapetyan (approved):

PR #231, opened 2017-08-16
POC testing out dynamodb methods and specifically sparse indexes

narehayrapetyan:
#### Summary

Proof of concept data models and Dynamodb operations.

Adds data models and dynamodb models and crud operations + some tests to see how sparse indices work when the index attribute is removed.

#### Testing

gradlew build && gradlew integTests

New tests cover the changes: 
yes

#### Licensing


data-service/src/main/java/com/amazonaws/blox/dataservice/storage/model/DeploymentDDBRecordTranslator.java
```diff
@@ -0,0 +1,62 @@
...
+ * or in the "LICENSE" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.blox.dataservice.storage.model;
+
+import com.amazonaws.blox.dataservice.model.Deployment;
+
+public class DeploymentDDBRecordTranslator {
```
wjbuys, line 19: I like the decoupling between our model and DB, but this kind of repetitive code looks like it's a good place for bugs to hide. It might be a good use case for using [`mapstruct`](http://mapstruct.org/) or [another bean-to-bean mapper](https://github.com/arey/java-object-mapper-benchmark).

data-service/src/main/java/com/amazonaws/blox/dataservice/storage/model/DaemonEnvironmentLatestVersionDDBRecord.java
```diff
@@ -0,0 +1,47 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+
+@DynamoDBTable(tableName = "DaemonEnvironmentLatestVersion")
+@Data
+@Builder
+//Required for builder because an empty constructor exists
+@AllArgsConstructor
+public class DaemonEnvironmentLatestVersionDDBRecord {
```
wjbuys, line 30: What's the intended use case for this model? It doesn't seem to be used in this PR, and I can't figure out just from its structure what it's for.

data-service/src/main/java/com/amazonaws/blox/dataservice/storage/DeploymentStore.java
```diff
@@ -0,0 +1,36 @@
...
+  Deployment updateDeployment(Deployment deployment);
+
+  // POC test method
+  void deleteAllDeployments();
+
+  Deployment getDeploymentById(String deploymentId);
+
+  List<Deployment> getDeploymentsByStatus(DeploymentStatus status);
+
+  List<Deployment> getAllDeploymentsInIndex();
```
wjbuys, line 35: Is this API also just for testing? What would a realistic use case of the index look like? `getAllPendingDeployments`?

data-service/src/integTest/java/com/amazonaws/blox/dataservice/storage/DDBModelIntegTest.java
```diff
@@ -0,0 +1,90 @@
...
+    AmazonDynamoDB dynamoDBClient =
+        AmazonDynamoDBClient.builder()
+            .withCredentials(new ProfileCredentialsProvider(CREDENTIALS_PROFILE_NAME))
+            .build();
+    DynamoDBMapper dynamoDBMapper = new DynamoDBMapper(dynamoDBClient);
+    deploymentDDBStore = new DeploymentDDBStore(dynamoDBMapper);
+  }
+
+  @Test
+  public void testSparseDeploymentStatusIndex() {
```
wjbuys, line 50: This test seems to verify the described behaviour of sparse indices from the docs accurately, and seems like a good harness for testing out our other questions about how this will perform.

Some of the ones I'm worried about from our discussions:
- if our hash key has an extremely small cardinality (i.e. DeploymentStatus alone), does this bottleneck our write performance on the main table? Or does it increase the eventual consistency window?
- what happens when we have a large number of deployments that do meet the criteria for inclusion in the sparse index? In particular, it looks like we'd have to use single-threaded querying to find all deployments in a particular state. What does this mean for the performance of our polling loop for starting deployments? Will it slow down linearly with number of available deployments?

data-service/src/main/java/com/amazonaws/blox/dataservice/storage/DeploymentDDBStore.java
```diff
@@ -0,0 +1,101 @@
...
+    final DeploymentDDBRecord record =
+        dynamoDBMapper.load(DeploymentDDBRecord.withHashKey(deploymentId));
+    return DeploymentDDBRecordTranslator.fromDeploymentDDBRecord(record);
+  }
+
+  @Override
+  public List<Deployment> getDeploymentsByStatus(DeploymentStatus status) {
+    DynamoDBQueryExpression<DeploymentDDBRecord> expression =
+        new DynamoDBQueryExpression<DeploymentDDBRecord>()
+            .withIndexName(DeploymentDDBRecord.DEPLOYMENT_STATUS_GSI_NAME)
```
wjbuys, line 76: [The docs](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBMapper.Methods.html#DynamoDBMapper.Methods.query) suggest that:

> To query an index, you must first model the index as a mapper class. Suppose that the Reply table has a global secondary index named PostedBy-Message-Index. The partition key for this index is PostedBy, and the sort key is Message. The class definition for an item in the index would look like this:
> ```
> @DynamoDBTable(tableName="Reply")
> public class PostedByMessage { 
> ...
> ```
It then goes on to show a separate POJO (than the main table POJO) mapped onto the index, with index annotations on *that* class, and the query executed for *that* record object.

How does it work when the index and the table are defined on the same class? Is this just a shorthand for an index with the exact same schema as the main table?

data-service/src/main/java/com/amazonaws/blox/dataservice/model/Deployment.java
```diff
@@ -0,0 +1,35 @@
...
+
+import lombok.Builder;
+import lombok.Data;
+import lombok.NonNull;
+
+import java.time.Instant;
+
+@Data
+@Builder
+public class Deployment {
```
wjbuys, line 25: Are these intended to be private models to the `data-service` project as they are now, or should they be part of the public model in `data-service-model`? If the former, should probably be in a different package from the models in `data-service-model`

data-service/src/main/java/com/amazonaws/blox/dataservice/storage/DeploymentDDBStore.java
```diff
@@ -0,0 +1,101 @@
...
+    final DeploymentDDBRecord record =
+        DeploymentDDBRecordTranslator.toDeploymentDDBRecord(deployment);
+    dynamoDBMapper.save(record);
+    return DeploymentDDBRecordTranslator.fromDeploymentDDBRecord(record);
+  }
+
+  @Override
+  public Deployment updateDeployment(final Deployment deployment) {
+    DeploymentDDBRecord record =
+        dynamoDBMapper.load(DeploymentDDBRecord.withHashKey(deployment.getDeploymentId()));
```
wjbuys, line 46: Why the extra `load` here? Any reason to not use the same `toDeploymentDDBRecord` we're using for creates? The usage pattern in the test suggests that we'll be querying a valid `Deployment` from the store beforehand anyway...

data-service-client/build.gradle
```diff
@@ -1,6 +1,8 @@
-description 'Blox DataService Client'
+plugins {
+    id "java"
```
wjbuys, line 2: What does this syntax do differently from the `apply plugin: 'java'` syntax? Is it a functional change?

wjbuys (commented):
One or two questions on the basic data model, and then mostly comments on stuff that we can test out with this in place (from our discussion on Friday)

data-service-client/build.gradle
wjbuys: Ah, looks like this is a [new incubating feature](https://docs.gradle.org/current/userguide/plugins.html#sec:plugins_block). Should we be replacing all our existing `apply` blocks?

wjbuys: Answering my own question after [researching this for the spring dependency plugin](https://plugins.gradle.org/plugin/io.spring.dependency-management). The biggest problem for our case is:

> **Cannot be used in conjunction with `subprojects {}`, `allprojects {}` etc.**
> It is not possible to use the familiar pattern of applying a plugin to multiple projects at once using `subprojects {}` etc. at the moment. There is currently no mechanism for applying a plugin to multiple projects at once. At the moment, each project that requires a plugin must declare so in the `plugins {}` block in its buildscript.
>
> Future versions of Gradle will remove this restriction.
>
> If the current limitations of the new plugin mechanism are problematic for your build, the recommended approach is to use the “traditional” mechanism for the time being.

Since we're doing this for the spotless plugin, among others, it's probably not going to work to go this way fully right now, unless we want to duplicate the config for that across all projects.

data-service/src/main/java/com/amazonaws/blox/dataservice/storage/model/DeploymentDDBRecordTranslator.java
narehayrapetyan: oooh that's a great idea, don't know why I didn't think to look for libraries to do this.

narehayrapetyan: I think I'll go through and fix this for everything I have so far and not just this PR cause I'll need to change it for create environment and get environment API PR as well.

data-service/src/main/java/com/amazonaws/blox/dataservice/storage/model/DaemonEnvironmentLatestVersionDDBRecord.java
narehayrapetyan: So this is for keeping track of the latest version of the environment since we need to keep all the versions around to be able to roll back to a specific version. If we don't have this we cannot retrieve the latest version. Do you have a better idea on how to model this?

data-service/src/main/java/com/amazonaws/blox/dataservice/storage/DeploymentStore.java
narehayrapetyan: Correct. I'll clean up the interface once we're done playing with the POC.

data-service/src/integTest/java/com/amazonaws/blox/dataservice/storage/DDBModelIntegTest.java
narehayrapetyan: Yea we should add a task for that

data-service/src/main/java/com/amazonaws/blox/dataservice/storage/model/DeploymentDDBRecordTranslator.java
narehayrapetyan: Created an [issue](https://github.com/blox/blox/issues/242)

data-service/src/integTest/java/com/amazonaws/blox/dataservice/storage/DDBModelIntegTest.java
narehayrapetyan: Created an [issue](https://github.com/blox/blox/issues/246).

data-service/src/main/java/com/amazonaws/blox/dataservice/storage/DeploymentDDBStore.java
narehayrapetyan: It has a higher chance of success if we retrieve the latest version before updating it.

data-service/src/main/java/com/amazonaws/blox/dataservice/storage/DeploymentDDBStore.java
narehayrapetyan: It seems to work just fine. The index schema I think is defined when the table is created (ALL attributes vs just keys).

data-service/src/main/java/com/amazonaws/blox/dataservice/model/Deployment.java
narehayrapetyan: Nope, these are internal models so there is separation between the public request/response objects defined in the data-service-model and the internal representation of data.

data-service/src/main/java/com/amazonaws/blox/dataservice/storage/model/DaemonEnvironmentLatestVersionDDBRecord.java
wjbuys: I find it quite hard to evaluate the fitness of (and consequently suggest improvements to) a data model, unless I can see how it's intended to be used. I'll probably hold off on comments on this until there's some code actually using it.

data-service/src/main/java/com/amazonaws/blox/dataservice/storage/DeploymentDDBStore.java
wjbuys: What does "higher chance of success" mean here? It kind of looks like we're just ignoring conflicts: in particular, it looks like the RecordTranslator [doesn't update the `version` field on the new record to match the one that was given here](https://github.com/blox/blox/pull/231/files#diff-5a38996365ad856775ba66b351b16b0fR36), so if deployment status was updated after this code initially read the record, then the DDB Mapper's optimistic concurrency won't catch it. Is that the intention? If not, this approach should be doing some of its own checking for dirty fields, and failing early if anything might conflict.

Additionally, it's entirely opaque to the consumer of this API that the only fields that will get updated from the given Deployment are `deploymentStatus` and the timestamps.

Is the rest of the record considered immutable? If so, a dedicated `updateStatus(DeploymentId, Status)` API might be less confusing, and provide a better boundary for concurrency control.

data-service/src/main/java/com/amazonaws/blox/dataservice/model/Deployment.java
wjbuys: Okay. So just to sanity check, we'll have at least 5 layers of representation for the `Deployment` object:

- Frontend Client (generated from frontend model)
- Frontend Service Model (defined in code)
- DataService Model (defined in code)
- DataService Internal (defined in code)
- DataService DB (defined in code)

That's 5 places where model changes need to be applied (4 manual changes and one code regen). Is this amount of layering buying us enough to warrant that cost?

The two obvious ones we must isolate is the Frontend Model and the DDB Record. Could we use a single representation internally in the system for the code between those layers? The cost from our point of view would be that all the code that shares that model would be deployment-coupled, but I think we'll need that anyway. Thoughts?

data-service/src/main/java/com/amazonaws/blox/dataservice/storage/DeploymentDDBStore.java
narehayrapetyan: The optimistic concurrency version is updated by dynamodb automatically.

But as far as it's not clear which fields are being updated, I agree, we'll have to update APIs/method names once it's clear which fields we want updated.

data-service/src/main/java/com/amazonaws/blox/dataservice/model/Deployment.java
narehayrapetyan: as we talked offline, I think it depends whether we need FE to be a separate service. Otherwise this can be reduced to FE client model + DS internal model + DS DB model.

wjbuys: Sounds good.

narehayrapetyan:
Moving the proof of concept into its own feature branch for now.

PR #237, opened 2017-08-16
Pin swagger version to 1.5.16

wjbuys:
Swagger just started releasing the 2.0.0 snapshot version, which caused
some dependency problems due to loose version constraints:

```
Could not find io.swagger:swagger-annotations:2.0.0-SNAPSHOT
```

This pins the version to 1.5.16 with a property in the root project.


build.gradle
```diff
@@ -97,6 +97,8 @@ ext {
     awsCommand = "aws"
 
     sdkZip = file("${buildDir}/java-sdk-${version}.zip")
+
+    swaggerVersion = "1.5.16"
```
narehayrapetyan, line 101: Would it make sense to use this dependency management plugin instead? https://github.com/spring-gradle-plugins/dependency-management-plugin

wjbuys: Oh, hm, didn't know about that. Looks like it may take some thinking to integrate that, and I'm pretty unfamiliar with it. Any objections to coming back to retrofit that? This is currently breaking all builds.

wjbuys: Okay, so I've tried that, and it has a bunch of rough edges:

- either I'm missing something, or it doesn't seem to affect `buildScript` dependencies, which is where this problem was introduced
- it doesn't inherit settings in subprojects. You either have to configure it everywhere with `subprojects { }`, or manually include it in each project. So you still need the equivalent of a `${swaggerVersion}` substitution to wire those together.

I've already sunk almost an hour on this; I now no longer think it makes sense to try to use this plugin for this particular issue. We can add it once we depend on the AWS SDK in more places.

narehayrapetyan (approved):

PR #243, opened 2017-08-22
Adding start and get deployment

narehayrapetyan:
#### Summary

Implements start and get deployments (look at the latest commit to exclude previous PR changes)

#### Testing

gradlew build and integTests

#### Licensing


data-service/src/main/java/com/amazonaws/blox/dataservice/storage/model/DeploymentDDBRecord.java
```diff
@@ -0,0 +1,117 @@
...
+
+  public static final String DEPLOYMENT_ID_HASH_KEY = "deploymentId";
+  public static final String DEPLOYMENT_STATUS_GSI_NAME =
+      "deploymentStatus-environmentName-index-copy";
+  public static final String DEPLOYMENT_STATUS_INDEX_HASH_KEY = "deploymentStatus";
+  //TODO: fix name to environmentNameAccountId
+  public static final String DEPLOYMENT_STATUS_INDEX_SORT_KEY = "environmentName";
+
+  @Deprecated
+  public DeploymentDDBRecord() {
```
wjbuys, line 47: Can we use `@NoArgsConstructor` for these instead?

data-service/src/main/java/com/amazonaws/blox/dataservice/mapper/DeploymentMapper.java
```diff
@@ -0,0 +1,34 @@
...
+import com.amazonaws.blox.dataservicemodel.v1.model.GetDeploymentRequest;
+import com.amazonaws.blox.dataservicemodel.v1.model.GetDeploymentResponse;
+import com.amazonaws.blox.dataservicemodel.v1.model.StartDeploymentRequest;
+import com.amazonaws.blox.dataservicemodel.v1.model.StartDeploymentResponse;
+import org.mapstruct.Mapper;
+
+@Mapper
+public interface DeploymentMapper {
+
+  Deployment startDeploymentRequestToDeployment(StartDeploymentRequest request);
```
wjbuys, line 27: How about simplifying the method names here to only include the target type? The source type could be handled only as a method overload:
```
@Mapper
public interface DeploymentMapper {

  Deployment toDeployment(StartDeploymentRequest request);
  Deployment toDeployment(GetDeploymentRequest request);
  StartDeploymentResponse toStartDeploymentResponse(Deployment deployment);
  GetDeploymentResponse toGetDeploymentResponse(Deployment deployment);
}
```
Makes it a little more readable when invoked, since the variable we're passing in already determines the source type.

data-service/src/main/java/com/amazonaws/blox/dataservice/storage/model/DaemonEnvironmentDDBRecordTranslator.java
```diff
@@ -0,0 +1,64 @@
...
+package com.amazonaws.blox.dataservice.storage.model;
+
+import com.amazonaws.blox.dataservice.model.Environment;
+import com.amazonaws.blox.dataservice.model.EnvironmentType;
+import com.amazonaws.blox.dataservice.model.InstanceGroup;
+
+//TODO: use object mapping library
+public class DaemonEnvironmentDDBRecordTranslator {
+
+  public static DaemonEnvironmentDDBRecord toDaemonEnvironmentDDBRecord(
```
wjbuys, line 24: Similarly to the other mapper, since we already know the exact types involved from the signature, perhaps this can just be `toRecord` and the inverse method `fromRecord`?

data-service/build.gradle
```diff
@@ -1,3 +1,103 @@
+plugins {
+    id 'net.ltgt.apt' version '0.11'
```
wjbuys, line 2: For some reason, this plugin's `apt` dependency type breaks the Lombok/IntelliJ integration for me. What does this buy us over declaring the annotation processor dependencies as `compileOnly`?

data-service/src/main/java/com/amazonaws/blox/dataservice/storage/model/DeploymentDDBRecord.java
```diff
@@ -0,0 +1,117 @@
...
+      final DeploymentStatus deploymentStatus,
+      final String environmentName,
+      final String accountId) {
+    return DeploymentDDBRecord.builder()
+        .deploymentStatus(deploymentStatus)
+        .environmentNameAccountId(createEnvironmentNameAccountId(environmentName, accountId))
+        .build();
+  }
+
+  public static String createEnvironmentNameAccountId(
```
wjbuys, line 69: So, this almost feels like this should be an `EnvironmentId` class with a DDB type converter that appropriately converts it to/from a string. I played around with it, and came up with [this](https://github.com/narehayrapetyan/blox/compare/startDeployment...wjbuys:pr/243).

What do you think? I find it a smidge easier to follow.

data-service/src/main/java/com/amazonaws/blox/dataservice/storage/model/DaemonEnvironmentLatestVersionDDBRecord.java
```diff
@@ -0,0 +1,47 @@
...
+
+  public static final String ENVIRONMENT_NAME_HASH_KEY = "environmentName";
+  public static final String ENVIRONMENT_VERSION_SORT_KEY = "environmentVersion";
+
+  @Deprecated
+  public DaemonEnvironmentLatestVersionDDBRecord() {
+    // Needed by dynamodb mapper
+  }
+
+  @DynamoDBHashKey(attributeName = ENVIRONMENT_NAME_HASH_KEY)
```
wjbuys, line 40: Since this attribute name matches the field name, do we gain some benefit from explicitly re-specifying the name here?

data-service/src/main/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandler.java
```diff
@@ -0,0 +1,67 @@
...
+    environment.setStatus(EnvironmentStatus.Inactive);
+    environment.setHealth(EnvironmentHealth.Healthy);
+    environment.setCreatedTime(Instant.now());
+    //TODO: how do we version?
+    environment.setEnvironmentVersion("1");
+
+    //TODO: environment exists exception
+
+    switch (environment.getType()) {
+      case Daemon:
```
wjbuys, line 50: Okay, I have questions on how you see this model evolving. Will we need to keep adding `case` clauses, `environmentStore` methods and DDB tables for every different deployment type? Is the structure of deployment configurations so different that we need to model the data we store about them differently? In particular, if we use separate DDB tables for each type of deployment, will we ever be able to get all deployments of all types, for e.g. UI purposes?

Could we rather encapsulate the things that are different about environments as something like a `DeploymentConfiguration` state object? Something to the effect of [this refactoring](https://refactoring.guru/replace-type-code-with-state-strategy) might work.

data-service/src/main/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandler.java
```diff
@@ -0,0 +1,67 @@
...
+        return environmentStore.createDaemonEnvironment(environment);
+      default:
+        throw new IllegalArgumentException(
+            String.format(
+                "Environment of type %s is not supported", environment.getType().toString()));
+    }
+  }
+
+  public Environment getLatestEnvironmentVersion(final Environment environment)
```
wjbuys, line 59: It's a bit weird that we have a method to get the latest version, but not to get an arbitrary version...

data-service/src/integTest/java/com/amazonaws/blox/dataservice/api/DeploymentIntegTest.java
```diff
@@ -0,0 +1,97 @@
...
+import org.springframework.test.context.junit4.SpringRunner;
+
+@RunWith(SpringRunner.class)
+@ContextConfiguration(classes = {Application.class})
+public class DeploymentIntegTest {
+
+  @Autowired private DataService dataService;
+
+  @Test
+  public void createAndGetDeployment()
```
wjbuys, line 51: This is a good start to test that everything is wired up, but in order to evaluate the data model, we need a more realistic outline of how we'd interact with the Deployment resource. Could we simulate the calls that the scheduling manager would make in a test case here, so that we can ensure the data model serves that purpose appropriately?

In particular, these are ones that stand out from the [design](https://github.com/blox/blox/blob/dev/docs/daemon_design.md#pending-deployments):
- how does the scheduling manager reconcile deployments?
- what queries will the scheduling manager make when new deployments show up?
- how will the scheduling workflow update the deployment state?
- how will the UI display ongoing deployments?

wjbuys (commented):
Some solid foundations in here, looking good.

My main concern right now is the modeling of environments configured for different kinds of deployments, and whether we have a good picture of how interactions with the deployment resource would work.

data-service/src/main/java/com/amazonaws/blox/dataservice/storage/model/DeploymentDDBRecord.java
narehayrapetyan: good idea!

narehayrapetyan:
Closing to review after POC is done

PR #249, opened 2017-08-28
Stub DynamoDB Stream Processor 

narehayrapetyan:
#### Summary

Sets up a SAM template for the stream processor, sets up Lambda logging and reads from the Deployment Dynamodb table stream. (commit #0ca705f)

#### Testing

build && sam local testing && tested by uploading to lambda and invoking the function by running the dataservice integration tests that add/remove items from the deployment table.

#### Licensing


scheduling-manager/src/main/java/com/amazonaws/blox/schedulingmanager/stream/processor/DeploymentRecordProcessor.java
```diff
@@ -0,0 +1,63 @@
...
+@Slf4j
+@Component
+@AllArgsConstructor
+public class DeploymentRecordProcessor implements RecordProcessor {
+
+  //new item
+  private static final String INSERT = "INSERT";
+  private static final String DEPLOYMENTS_HASH_KEY = "deploymentId";
+
+  private ObjectMapper objectMapper;
```
wjbuys: nit: Unused?

scheduling-manager/src/main/java/com/amazonaws/blox/schedulingmanager/stream/processor/StreamProcessorHandler.java
```diff
@@ -0,0 +1,50 @@
...
+  @Override
+  public String handleRequest(final DynamodbEvent dynamodbEvent, final Context context) {
+
+    for (DynamodbStreamRecord record : dynamodbEvent.getRecords()) {
+      log.debug(record.toString());
+
+      final RecordProcessor recordProcessor = applicationContext.getBean(RecordProcessor.class);
+
+      try {
+        recordProcessor.process(record);
```
wjbuys, line 42: Are there other ways in which record processing can fail, maybe non-fatally? If so, what can we do to re-process the record; is there some kind of retry/DLQ support in the DDB Streams world?

Either way, if this is trying out our approach to handling these streams, we should probably build some partial failure handling into the RecordProcessor interface.

scheduling-manager/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Configuration status="INFO" packages="com.amazonaws.services.lambda.runtime.log4j2">
+<Appenders>
+  <Console name="Console" target="SYSTEM_OUT">
```
wjbuys: before shipping: We probably don't want both a Console and a Lambda appender when running inside an actual lambda function, since that will duplicate every log message.

wjbuys (commented):
Is the idea that this will form the base of the record processing pipeline? If so, I think we need to have some idea in here for how we're going to surface the partial failure case.

scheduling-manager/src/main/java/com/amazonaws/blox/schedulingmanager/stream/processor/StreamProcessorHandler.java
narehayrapetyan: If the processing of the record fails, the entire batch that's passed into a Lambda fails and Lambda retries the batch until successful or expired.

scheduling-manager/src/main/java/com/amazonaws/blox/schedulingmanager/stream/processor/DeploymentRecordProcessor.java
narehayrapetyan: removed

scheduling-manager/src/main/java/com/amazonaws/blox/schedulingmanager/stream/processor/StreamProcessorHandler.java
wjbuys: Does that mean we need to ensure that the underlying RecordProcessor is idempotent? Seems like it could be asked to process the same record many times...

wjbuys (approved):

scheduling-manager/src/main/java/com/amazonaws/blox/schedulingmanager/stream/processor/StreamProcessorHandler.java
narehayrapetyan: Yes, it's described in the [design doc.](https://github.com/blox/blox/blob/dev/docs/daemon_infrastructure.md)

PR #252, opened 2017-09-07
Document how users will interact with deployments

wjbuys:
#### Summary
This outlines how users will interact with Blox deployments.

The format is intentionally similar to the [k8s Deployment docs](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/) and the [design doc](https://github.com/blox/blox/blob/dev/docs/daemon_design.md#user-experience), but with a focus on what users will be able to observe during a deployment.

#### Licensing


docs/deployment-ux.md
```diff
@@ -0,0 +1,410 @@
...
+-------------------------
+
+Let's say halfway through a deployment, we realize that we made a
+mistake. Rolling back is one option, but might introduce more churn than
+we can handle. For this case, we can pause all deployments until we can
+fix the environment and deploy the right active revision.
+
+Open questions:
+
+-   Should we support a separate resume operation, or will we only
```
wjbuys: To clarify, we'll be pausing Environments, not Deployments. In particular, pausing an environment means that it will not launch/terminate any tasks that belong to the environment until resumed.

We considered some scenarios where resume would be a useful separate concept to go along with pause:

- Something is wrong with the underlying cluster. We just want to pause the rollout until the underlying problem is fixed, and then resume where we left off. This could be particularly important for very large clusters.
- Some external large scale event is going on, and we don't want to continue making changes. For example: if we're doing a long slow rollout, but the ECS control plane or some startup dependency of the tasks we're launching becomes unavailable, then we want to pause cluster changes until the situation resolves. Neither rolling forward nor rolling back would make any progress, and could adversely affect fleet health. Once the situation resolves, we can resume from where we left off.

However, doing an explicit rollback/roll forward should implicitly resume the environment.

docs/deployment-ux.md
```diff
@@ -0,0 +1,410 @@
...
+to revision 1:
+
+``` {.shell}
+$ aws ecs describe-environment-status --environment-name SomeEnvironment/Prod --table
+REVISION                  STATUS       DESIRED   CURRENT   UP-TO-DATE   AVAILABLE
+SomeEnvironment/Prod:1    Inactive     0         0         0            0        
+SomeEnvironment/Prod:2    Active       5         5         5            5        
+```
+
+**Option 1**: Allow specifying earlier revisions to deploy. That
```
wjbuys: We've decided that we'll go for Option 1, as it provides better insight into which environment revisions are actually running. With Option 2 (create a new environment revision to effect the rollback), we would no longer be able to tell that revision 1 and revision 3 are actually the same.

The tradeoff we'd be making, is that it's harder to see exactly when rollbacks happened. For example, if we repeatedly roll back to the same revision from multiple attempted deployments, all we'd see would be:

```
$ aws ecs describe-environment-status --environment-name SomeEnvironment/Prod --table
REVISION                  STATUS       DESIRED   CURRENT   UP-TO-DATE   AVAILABLE
SomeEnvironment/Prod:1    Active       5         5         5            5        
SomeEnvironment/Prod:2    Reverted     0         0         0            0    
SomeEnvironment/Prod:3    Reverted     0         0         0            0    
SomeEnvironment/Prod:4    Reverted     0         0         0            0    
```
In this case, we probably want to keep some sort of audit log around.

docs/deployment-ux.md
```diff
@@ -0,0 +1,410 @@
...
+{ "EnvironmentRevision": 1 }
+```
+
+At this point, nothing has actually happened to your cluster. You can
+see this by showing your environment's status:
+
+``` {.shell}
+$ aws ecs describe-environment-status --environment-name SomeEnvironment/Prod --table
+REVISION                  STATUS     DESIRED   CURRENT   UP-TO-DATE   AVAILABLE
+SomeEnvironment/Prod:1    Active     5         5         5            5
```
wjbuys: In addition to this view of progress, we may wish to show a clearer breakdown of the progress of the entire environment to its goal state. For example:

```
$ aws ecs describe-environments
ENVIRONMENT           STATUS                  PROGRESS
SomeEnvironment/Gamma Active                  5/5 tasks
SomeEnvironment/Prod  Deploying revision 2    1/5 tasks
```

docs/deployment-ux.md
```diff
@@ -0,0 +1,410 @@
...
+
+``` {.shell}
+$ aws ecs describe-environment-status --environment-name SomeEnvironment/Prod --table
+REVISION                  STATUS       DESIRED   CURRENT   UP-TO-DATE   AVAILABLE
+SomeEnvironment/Prod:1    Undeploying  0         5         5            5        
+SomeEnvironment/Prod:2    Deploying    6         0         0            0        
+```
+
+Note that even though this doesn't change the task definition, it will
+still stop and restart tasks. (TODO: Can we make it smarter, so that it
```
wjbuys: This behaviour (stopping/starting all tasks for an instance group change) is not what we think customers will want (*"I just added an extra instance to my instance group, why is it killing and restarting every single task??"*). It would make much more sense to customers if we seamlessly detected that there was no other change, and just roll out new tasks to new instances that are newly part of the group, or terminate tasks on instances that are no longer in the group.

Our limitation to making this smarter is that we can't attach mutable metadata to instances to identify which environment revision it belongs to. We think we can do something smart to implement it this way anyway, so we'll aim for the better, more efficient solution.

docs/deployment-ux.md
```diff
@@ -0,0 +1,410 @@
...
+
+Success:
+
+``` {.shell}
+$ aws ecs describe-environment-status --environment-name SomeEnvironment/Prod --table
+REVISION                  STATUS          DESIRED   CURRENT   UP-TO-DATE   AVAILABLE
+SomeEnvironment/Prod:1    Active          5         5         5            5         
+```
+
+Updating an environment
```
wjbuys: This section needs to be expanded to show what users will see if the underlying cluster state changes while an update deployment is going on. In particular:

- What happens when new instances join the cluster while an environment is deploying?
- What happens when instances are terminated while an environment is deploying?
- What happens when instances don't come up, or can't be placed (e.g. because the target instance can't accommodate more tasks)?

docs/deployment-ux.md
```diff
@@ -0,0 +1,410 @@
+Blox Deployments User Experience
+================================
+
+Use Cases
+---------
+
+The following are typical use cases for Deployments:
+
+-   Create an Environment to deploy a new environment revision. The
```
wjbuys: We discussed terminology around this a bit. In particular, we have some overlapping concepts with k8s that have different meanings there (particularly `deployment`). We're still evaluating whether we should adopt the terms that are already in widespread use with k8s (and run the risk of confusion because they don't behave exactly the same), or to use entirely new terms (and make it harder for new customers familiar with container orchestration to understand the entities in our system). Having duplicate terms with totally different meanings seems like something we should try and avoid, though.

docs/deployment-ux.md
```diff
@@ -0,0 +1,410 @@
...
+EOF
+{ "EnvironmentRevision": 1 }
+```
+
+At this point, nothing has actually happened to your cluster. You can
+see this by showing your environment's status:
+
+``` {.shell}
+$ aws ecs describe-environment-status --environment-name SomeEnvironment/Prod --table
+REVISION                  STATUS     DESIRED   CURRENT   UP-TO-DATE   AVAILABLE
```
wjbuys: I copied the table headers from `k8s`s format. Up-to-date and Available don't yet have analogous concepts in Blox, so I'll remove them from all examples.

wjbuys (commented):
I've summarized the feedback from an in-person discussion with @narehayrapetyan, @simplycloud and @shubharao as review comments. If there's something I left out, please leave a review comment.

docs/deployment-ux.md
```diff
@@ -0,0 +1,410 @@
...
+every instance in the target Instance Group:
+
+``` {.shell}
+$ aws ecs describe-environment-status --environment-name SomeEnvironment/Prod --table
+REVISION                  STATUS       DESIRED   CURRENT   UP-TO-DATE   AVAILABLE
+SomeEnvironment/Prod:1    Undeploying  0         5         5            5        
+SomeEnvironment/Prod:2    Deploying    5         0         0            0        
+```
+
+Once the deployment reaches a stead state, the old revision is marked
```
dbenhur: nit: stead**y** state

wjbuys: Thanks; there's a bunch of typos I'll fix shortly. Any other feedback would be appreciated as well.

docs/deployment-ux.md
```diff
@@ -0,0 +1,410 @@
...
+
+    ``` {.shell}
+    $ aws ecs describe-environment-status --environment-name SomeEnvironment/Prod --table
+    REVISION                  STATUS       DESIRED   CURRENT   UP-TO-DATE   AVAILABLE
+    SomeEnvironment/Prod:1    Inactive     0         0         0            0        
+    SomeEnvironment/Prod:2    Active       5         5         5            5        
+    ```
+
+2.  Terminate after replace
+
```
dbenhur: It may be premature at present, but we might want to think about advanced features of this mode.  There are apps out there that do a bunch of management for graceful upgrade, coordinating start of new and hand-off from old via signals (TTOU,TTIN,USR1,etc) and sometimes shared memory. (I'm thinking of things like [haproxy](https://www.haproxy.org/download/1.7/doc/management.txt), [nginx](http://nginx.org/en/docs/control.html), [unicorn](https://bogomips.org/unicorn/SIGNALS.html)). I haven't tried any of this with containers, whose isolation boundaries may make this sorta coordination more difficult.  There's some CLOSED WONTIMPLEMENT docker tickets for [`docker replace`](https://github.com/moby/moby/issues/2733) or other [graceful upgrade capabilities](https://github.com/docker/compose/issues/1663), so it looks like Docker isn't interested in solving the problem, but that doesn't mean we couldn't innovate on behalf of our customers and come up with a workable scheme supported by our scheduler and agent managing the container runtime.

docs/deployment-ux.md
```diff
@@ -0,0 +1,410 @@
...
+    If you're using this deployment method, then Blox will launch
+    `MaxOver` tasks on as many instances, and only kill old tasks once
+    the new task is confirmed to be running. This will ensure that
```
dbenhur: We really want a better signal than "running" to trigger terminating the prior version; the lifecycle state one wants is "ready".  Until we support stronger lifecycle hooks/health-checks, we might get by with a user attribute specifying delay-before-terminate-old-secs.

docs/deployment-ux.md
```diff
@@ -0,0 +1,410 @@
...
+$ aws ecs describe-environment-status --environment-name SomeEnvironment/Prod --table
+REVISION                  STATUS       DESIRED   CURRENT   UP-TO-DATE   AVAILABLE
+SomeEnvironment/Prod:1    Inactive     0         0         0            0        
+SomeEnvironment/Prod:2    Undeploying  6         6         6            6        
+SomeEnvironment/Prod:3    Deploying    0         0         0            0        
+$ aws ecs describe-environment-status --environment-name SomeEnvironment/Prod --table
+REVISION                  STATUS       DESIRED   CURRENT   UP-TO-DATE   AVAILABLE
+SomeEnvironment/Prod:1    Inactive     0         0         0            0        
+SomeEnvironment/Prod:2    Inactive     0         0         0            0        
+SomeEnvironment/Prod:3    Active       6         6         6            6
```
dbenhur: I like these status displays, but environments are going to accumulate a lot of deployment revisions over time.  Do we have thoughts on how far back we keep/display history?  May want `--limit N` arg to let user tell us how much they want to see.

wjbuys: Indeed, the idea is to hide `Inactive` by default, and provide a `showInactive=true` option and pagination APIs; will add a bit at the top to that effect.

docs/deployment-ux.md
wjbuys: Good point. And if/when we get support for some kind of health check (e.g. aws/amazon-ecs-agent#534), we can use that.

docs/deployment-ux.md
wjbuys: On second thought, implicitly resuming could be dangerous if some automated system does a deployment while the reason for pausing the environment is still valid, that could cause harmful effects.

wjbuys:
Okay, I think I've addressed most of the comments. @narehayrapetyan please give it another look when you get a chance.

docs/deployment-ux.md
```diff
@@ -300,111 +310,193 @@ impacts
...
+    REVISION                  STATUS       DESIRED   CURRENT
+    SomeEnvironment/Prod:1    Inactive     0         0
+    SomeEnvironment/Prod:2    Active       5         5
+    ```
+
+#### Changes to cluster membership while a deployment is ongoing
+
+For larger clusters, or clusters that are backed by Autoscaling Groups, the available instances in the cluster might change (sometimes dramatically). Since the total number of available instances determines the desired end-state of the Environment, these state changes could affect the ability of an Environment to make progress while honoring its deployment constraints.
+
+The behaviour we're aiming for is to minimize the number of instances that are not running the Daemon task at all at any given point in time. If a configurable number of instances are without the correct version of a Daemon task for a configurable amount of time, the deployment will enter the `Failing` state and emit an alarm. However, it will continue to try and make progress.
```
narehayrapetyan: It's not clear why the instances won't have the correct version and result in a failing state.

docs/deployment-ux.md
```diff
@@ -300,111 +310,193 @@ impacts
...
+
+- If a deployment fails to start a task on a particular host (e.g. because the task definition is faulty, or the Task fails to start up in a nondeterministic way), priority will be given to restarting that task, rather than continuing to replace existing tasks.
```
narehayrapetyan: I think we should have a start/update deployment section and this should go there.

docs/deployment-ux.md
```diff
@@ -300,111 +310,193 @@ impacts
...
-    SomeEnvironment/Prod:1    Inactive     0         0         0            0        
-    SomeEnvironment/Prod:2    Active       5         5         5            5        
     ```
+    $ aws ecs describe-environment-status --all --environment-name SomeEnvironment/Prod --table
+    REVISION                  STATUS       DESIRED   CURRENT
+    SomeEnvironment/Prod:1    Inactive     0         0
+    SomeEnvironment/Prod:2    Active       5         5
+    ```
+
+#### Changes to cluster membership while a deployment is ongoing
```
narehayrapetyan: I think we should break this out into sections: how deployment behaves when there are no cluster changes and how do we handle instances being added/removed and task failures.

docs/deployment-ux.md
```diff
@@ -300,111 +310,193 @@ impacts
...
+
+The behaviour we're aiming for is to minimize the number of instances that are not running the Daemon task at all at any given point in time. If a configurable number of instances are without the correct version of a Daemon task for a configurable amount of time, the deployment will enter the `Failing` state and emit an alarm. However, it will continue to try and make progress.
+
+- If a deployment fails to start a task on a particular host (e.g. because the task definition is faulty, or the Task fails to start up in a nondeterministic way), priority will be given to restarting that task, rather than continuing to replace existing tasks.
+- If a Daemon task terminates abnormally during a deployment on an instance that is not currently having its task replace, priority will be given to replacing that task over starting to terminate tasks on other instances.
```
narehayrapetyan: task replace -> task replaced

docs/deployment-ux.md
```diff
@@ -300,111 +310,193 @@ impacts
...
+- If new instances join the cluster while an environment is deploying, running the daemon task on those instances will be given priority over replacing existing instances of the daemon. The new instances will be visible only as a change in the `DESIRED` column for the `Deploying` environment revision:
```
narehayrapetyan: I think we should discuss alternatives and give pros and cons explaining why this is a better option.

narehayrapetyan (commented):
I didn't review the content ie whether I agree with the user experience since we should do that in the meeting but mostly focused on whether I think you covered the cluster change use cases. We can probably think of some other edge cases, I'll let you know if I think of some.

docs/deployment-ux.md
wjbuys: Ah, I think I meant "EnvironmentRevision" here, not "version". Basically, if the scheduler has to choose whether to

1. first replace an existing, older revision of the environment's task on an instance, or
2. start the current revision of the environment's task on a host that doesn't have any task running on it at all,

 it should prioritize the latter.

I'll see if I can reword this to be clearer.

docs/deployment-ux.md
wjbuys: Yeah, that's the bit that I mentioned will take some more time; I tried to distill the behaviour into more of a generalized tenet (try to minimize the number of instances that don't run any version of the environment at all). I'll do that when I get back on Wednesday.

In the mean time, are there any particular alternatives that you think we should cover here?

docs/deployment-ux.md
wjbuys: > start/update deployment section

Could you clarify? What kind of detail would you like to see in that section? I don't think I understand how this is different from the general description of first launch at the start of the doc.

docs/deployment-ux.md
wjbuys: I think the "deployment strategies" section basically illustrates the "what happens when the cluster doesn't change"? I can rename that to reflect that it's the happy path, and then break this section out into separate scenarios for both deployment strategies. Come to think of it, the behaviour might be different, depending on strategy, so that's a good move either way.

Would that cover it?

docs/deployment-ux.md
wjbuys: As discussed in person, will extract the invariant here into a section, and then discuss the implications for the state changed/no state changed scenarios.

narehayrapetyan (approved):

PR #254, opened 2017-09-08
Adding deployment workflow template with one lambda fun…

narehayrapetyan:
…ction per workflow support

#### Summary

Defines a template to create stepfunctions wf and lambda functions with dummy lambda handlers that only pass input/output. There is only one lambda function that handles the routing to the correct handler based on the lambda function alias.

#### Testing

build and ran the workflow with input

* ./gradlew build
* aws cloudformation  package \
    --template-file scheduling-manager-workflows/templates/scheduling_manager_workflows.yml \
    --output-template-file scheduling-manager-workflows/templates/output.yml \
    --s3-bucket us-east-1-blox-scheduler-manager
* aws cloudformation deploy --template-file scheduling-manager-workflows/templates/output.yml --stack-name deploymentworkflows --capabilities CAPABILITY_IAM
* then ran the workflow from the console with initial input:
{
  "name": "n",
  "account": "a"
}

got output:

{
  "clusterName": "cluster1"
}

#### Licensing


scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/DeploymentWorkflowApplication.java
```diff
@@ -0,0 +1,76 @@
...
+    return new ObjectMapper();
+  }
+
+  /**
+   * This map is used for routing in the main workflow lambda handler. The lambda function aliases
+   * used in the workflow definition need to match the map keys.
+   */
+  @Bean
+  public Map<String, Handler> handlers() {
+    return new HashMap<String, Handler>() {
```
wjbuys: nit: The double brace initialization trick for initializing hashes is generally not recommended, since it messes with equality checking (see http://wiki.c2.com/?DoubleBraceInitialization for details). I don't think it matters for this case, but since this isn't doing field initialization here, it's actually also shorter to not use this trick (4 lines instead of 6):

```java
HashMap<String, Handler> h = new HashMap<>();
h.put("GetDeploymentData", getDeploymentData());
h.put("GetStateData", getStateData());
return h;
```

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/GetStateData.java
```diff
@@ -0,0 +1,48 @@
...
+
+  private ObjectMapper mapper;
+
+  @Override
+  public String handleRequest(String input, Context context) throws IOException {
+    log.debug("getStateData lambda");
+
+    final DeploymentData deploymentData = mapper.readValue(input, DeploymentData.class);
+    log.debug(
+        String.format(
```
wjbuys: nit, throughout: The SLF4J `Logger` interface [already supports string formatting](https://www.slf4j.org/api/org/slf4j/Logger.html), no need to use `String.format`:

```java
log.debug("deployment data deployment id {} and cluster name {}", deploymentData.deploymentId, deploymentData.clusterName);
```

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/Handler.java
```diff
@@ -0,0 +1,26 @@
...
+
+import com.amazonaws.services.lambda.runtime.Context;
+import java.io.IOException;
+
+/**
+ * The lambda functions that will be handling the workflow steps should implement this interface.
+ */
+public interface Handler {
+
+  String handleRequest(String input, Context context) throws IOException;
```
wjbuys: What's the benefit of handling Strings here instead of a `Reader`/`Writer` pair? IIRC, the Jackson API will quite happily work with those as succinctly as with Strings.

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/GetDeploymentData.java
```diff
@@ -0,0 +1,53 @@
...
+        String.format(
+            "deployment input name %s and account %s",
+            deploymentInput.getName(), deploymentInput.getAccount()));
+
+    final DeploymentData deploymentData =
+        DeploymentData.builder()
+            .deploymentId(UUID.randomUUID().toString())
+            .clusterName("cluster1")
+            .build();
+    final Map<String, String> map = mapper.convertValue(deploymentData, Map.class);
```
wjbuys: Why is this necessary? Can't we write `deploymentData` straight to the Mapper? Only noticed this because of the `unchecked assignment` warning :(

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/GetStateData.java
```diff
@@ -0,0 +1,48 @@
...
+import lombok.AllArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Component;
+
+@Slf4j
+@Component
+@AllArgsConstructor
+public class GetStateData implements Handler {
+
+  private ObjectMapper mapper;
```
wjbuys: I wonder if it makes sense to have an abstract handler base class where we can pull all the serialization/deserialization cruft. It would be nice if we could rewrite handlers like:

```java
public class GetStateData extends Handler {
  public void handle(HandlerContext c) {
    DeploymentData deployment = c.input(DeploymentData.class)
    // do stuff
    c.output(outputObject);
  }
}
```

First prize would be doing something like LambdaSandboxJava itself does to power the typed `Handler<I,O>` interfaces, but that gets a bit reflection-y, and might be overkill.

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/MainLambdaHandler.java
```diff
@@ -0,0 +1,70 @@
...
+import java.io.InputStreamReader;
+import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
+import java.util.Map;
+import java.util.stream.Collectors;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class MainLambdaHandler extends DeploymentWorkflowApplication
```
wjbuys: Why does this extend the application configuration class?

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/MainLambdaHandler.java
```diff
@@ -0,0 +1,70 @@
...
+      throws IOException {
+
+    final String inputJson;
+
+    try (BufferedReader buffer = new BufferedReader(new InputStreamReader(inputStream))) {
+      inputJson = buffer.lines().collect(Collectors.joining("\n"));
+    }
+
+    // find handler matching the lambda function alias
+    final Map<String, Handler> handlers = applicationContext.getBean("handlers", Map.class);
```
wjbuys: Can we inject these into the handler with `@Autowired` instead of reaching into the context for it manually? This line generates an `unchecked assignment` warning as is, and it looks like the only way to avoid that is to move it to an injected field/constructor parameter.

scheduling-manager-workflows/templates/output.yml
```diff
@@ -0,0 +1,75 @@
+AWSTemplateFormatVersion: '2010-09-09'
```
wjbuys: Should this file be checked in?

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/types/StateData.java
```diff
@@ -0,0 +1,30 @@
...
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.blox.schedulingmanager.deployment.handler.types;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import lombok.Builder;
+import lombok.Value;
+
+@JsonDeserialize(builder = StateData.StateDataBuilder.class)
```
wjbuys: Cool! Didn't realize Jackson supported this.

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/types/DeploymentInput.java
```diff
@@ -0,0 +1,31 @@
...
+import lombok.Value;
+
+@JsonDeserialize(builder = DeploymentInput.DeploymentInputBuilder.class)
+@Builder
+@Value
+public class DeploymentInput {
+  private final String name;
+  private final String account;
+
+  @JsonPOJOBuilder(withPrefix = "")
```
wjbuys: Does this generate an additional builder class, if it's used alongside the lombok `@Builder` annotation? Or does it reuse that one?

wjbuys (commented):
Concept looks proven, great! Some suggestions for handler structure inline, and a couple of easy nitpicks.

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/DeploymentWorkflowApplication.java
narehayrapetyan: that works

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/GetStateData.java
narehayrapetyan: So to put all the input/output types into handlercontext? how does handlercontext know which object to deserialize to?

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/GetStateData.java
narehayrapetyan: cool

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/Handler.java
narehayrapetyan: can you give an example of what you're saying?

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/GetDeploymentData.java
narehayrapetyan: Oh, good catch, this is a leftover from a parent request class that held all objects in a map, I'll remove it.

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/MainLambdaHandler.java
narehayrapetyan: To instantiate applicationcontext. Previously when I had multiple lambda functions I didn't want that code everywhere, now that we have one main lambda function, I can put 

  public static final ApplicationContext applicationContext =
      new AnnotationConfigApplicationContext(DeploymentWorkflowApplication.class);
here

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/MainLambdaHandler.java
narehayrapetyan: This is the main function that initializes the beans, I don't think autowired can find beans before they're initialized, unless you know how to do that?

scheduling-manager-workflows/templates/output.yml
narehayrapetyan: I wasn't sure if we want to check it in so we know what exactly we're deploying? What do you think?

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/types/DeploymentInput.java
narehayrapetyan: Not sure actually, but it's the same name the lombok builder uses so I'm assuming there's only one builder. I'll check.

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/GetStateData.java
wjbuys: It doesn't, it just holds a handler to a mapper, and both streams:

```java
  public <IN> IN input(Class<IN> clazz) throws IOException {
    return mapper.readValue(input, clazz);
  }

  public <OUT> void output(OUT outputValue) throws IOException {
    mapper.writeValue(output, outputValue);
  }
```

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/Handler.java
wjbuys: See the comment with the `HandlerContext` implementation. Specifically, why explicitly read the streams into `String`s for this handler in `MainLambdaHandler`, when the `mapper.readValue(InputStream stream, Class<T> clazz)` and `mapper.writeValue(OutputStream stream, T object)` methods both operate on `InputStream`/`OutputStream` natively?

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/MainLambdaHandler.java
wjbuys: Makes sense. Might it then also make sense to move the `Map<String, Handler> handlers` to a simple field on this class, so that we don't have to deal with injection for that list?

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/MainLambdaHandler.java
wjbuys: See comment above about potentially not injecting those. Otherwise, 3 tiers of handlers could take care of it:

- `MainHandler`: instantiates spring context, and instantiates `DispatcherHandler`
- `DispatcherHandler`: uses the injected `Map` of `StepHandler`s to look up the appropriate handler for the given alias
- `*StepHandler`s: actually handles the step(s)

scheduling-manager-workflows/templates/output.yml
wjbuys: Since these will refer to user-specific resources that aren't necessary accessible to others, and probably shouldn't be deployed to production, I think we should generate them every time. Otherwise I'll have a noisy diff each time I run a deployment, even if I didn't change anything.

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/types/DeploymentInput.java
wjbuys: Yeah, that sounds right. Unfortunate that the boilerplate for that is a bit opaque :(

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/MainLambdaHandler.java
narehayrapetyan: What I can do is just initialize the context in this class and do the routing in a different class that is ready to be injected.

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/MainLambdaHandler.java
narehayrapetyan: Yep that's what I was suggesting above by having a separate class find the correct handler.

scheduling-manager-workflows/templates/output.yml
narehayrapetyan: What do you mean they shouldn't be deployed to production? Also, the only time the output template will change if the code changes and generates a new jar. This way you know explicitly that you've made changes to the template and are deploying a new version.

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/GetStateData.java
narehayrapetyan: Instead of putting this in an abstract class that each stephandler subclasses, I'll pull it out into and encoder class composition style.

scheduling-manager-workflows/templates/output.yml
narehayrapetyan: However, I'm not dead set on checking this in. Just wondering what you mean by it shouldn't be deployed to production.

wjbuys: Sure, my main concern is [this line](https://github.com/blox/blox/pull/254/files/6e163026060b85ace31bfcd6849fa7b58e3cc837#diff-be2bdc987d76ac6407f30b834de53337R23):

```
CodeUri: s3://us-east-1-blox-scheduler-manager/f0c12d0a9d4ed8b32b50c3a2fe104f40
```

That line refers to an S3 bucket owned by the account of whoever last ran `aws cloudformation package`, containing code that was built on a dev laptop. I don't think that is what we actually want to ship to production. Diff-wise, it will also change any time a different bucket is used, or the jar is rebuilt with slightly different timestamps.

Since that's also the only thing that's different from the input template, I'd say that a CR of the input template is enough change control for this.

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/MainLambdaHandler.java
wjbuys: 👍

scheduling-manager-workflows/templates/output.yml
narehayrapetyan: sure I'll remove it

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/Encoder.java
```diff
@@ -0,0 +1,35 @@
...
+import java.io.InputStream;
+import java.io.OutputStream;
+import lombok.AllArgsConstructor;
+
+@AllArgsConstructor
+public class Encoder {
+
+  private ObjectMapper mapper;
+
+  public <T> T decode(InputStream input, Class<T> clazz) throws IOException {
```
wjbuys, line 28: Hm. This interface has exactly the same signature as `ObjectMapper`. My original comment about this was more around encapsulating the `InputStream`/`OutputStream` binding behind a `HandlerContext` object, so that you wouldn't have to expose the streams in the `StepHandler` interface at all.

In this incarnation, I'm not sure this is adding value over using ObjectMapper directly, unless I'm missing something?

wjbuys (approved):
Changes look good; but if you have time to consider it, I'm not sure Encoder's the right abstraction here.

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/handler/Encoder.java
narehayrapetyan: Yea I don't think it's very useful either but I don't understand what you're suggesting in that case. I think it's fine to leave it as is so we can change the objectmapper later or just use the objectmapper directly.

PR #257, opened 2017-09-20
Listing instances in the cluster and starting startTask workflows for each instance

narehayrapetyan:
#### Summary

Lists instances in the specified cluster and starts a workflow per instance to start a task on each.

#### Testing

*    ./gradlew build

*    aws cloudformation package
    --template-file scheduling-manager-workflows/templates/scheduling_manager_workflows.yml
    --output-template-file scheduling-manager-workflows/templates/output.yml
    --s3-bucket us-east-1-blox-scheduler-manager

*    aws cloudformation deploy --template-file scheduling-manager-workflows/templates/output.yml --stack-name deploymentworkflows --capabilities CAPABILITY_IAM

*    start the deployment workflow from the console with initial input:
    {
    "name": "n",
    "account": "a"
    }

*    this will kick off one startTask workflow per instance

#### Licensing


scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/steps/GetStateData.java
```diff
@@ -14,35 +14,52 @@
...
 
     final DeploymentData deploymentData = encoder.decode(input, DeploymentData.class);
 
-    log.debug(
-        "deployment data deployment id {} and clustername {}",
-        deploymentData.deploymentId,
-        deploymentData.clusterName);
+    final ECSWrapper ecsWrapper =
+        ecsWrapperFactory.getWrapper(
+            ecsWrapperFactory.getCredentialsProvider(
```
wjbuys: Any reason to not have `ecsWrapperFactory.forRole(deploymentData.getEcsRole(), ROLE_SESSION_NAME_PREFIX)`, without the intermediate call?

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/steps/GetStateData.java
```diff
@@ -14,35 +14,52 @@
...
+                deploymentData.getEcsRole(), ROLE_SESSION_NAME_PREFIX));
+
+    final ListContainerInstancesResult listContainerInstancesResult =
+        ecsWrapper.listInstances(deploymentData.getCluster());
+
+    final StateData stateData =
```
wjbuys, line 53: What is a `StateData`? I can't see a cohesive purpose for it, since it mixes attributes from both the deployment and the cluster state.

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/wrapper/ECSWrapperFactory.java
```diff
@@ -0,0 +1,36 @@
...
+package com.amazonaws.blox.schedulingmanager.wrapper;
+
+import com.amazonaws.auth.AWSCredentialsProvider;
+import com.amazonaws.services.ecs.AmazonECSClient;
+import com.amazonaws.services.securitytoken.AWSSecurityTokenService;
+import lombok.AllArgsConstructor;
+import lombok.NonNull;
+
+@AllArgsConstructor
+public class ECSWrapperFactory implements WrapperFactory<ECSWrapper> {
```
wjbuys, line 23: It might be more accurate to name this `CrossAccountAmazonECS` or similar, and to have it also implement the `AmazonECS` interface, perhaps using [Lombok's `@Delegate` annotation](https://projectlombok.org/features/experimental/Delegate).

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/wrapper/ECSWrapperFactory.java
```diff
@@ -0,0 +1,36 @@
...
+
+  @NonNull private AWSSecurityTokenService stsClient;
+
+  public ECSWrapper getWrapper(final AWSCredentialsProvider credentialsProvider) {
+    return new ECSWrapper(AmazonECSClient.builder().withCredentials(credentialsProvider).build());
+  }
+
+  public AWSCredentialsProvider getCredentialsProvider(
+      final String roleArn, final String roleSessionName) {
+    return WrapperFactory.super.getCredentialsProvider(stsClient, roleArn, roleSessionName);
```
wjbuys: This doesn't override anything, and isn't different from the `default` implementation in the `WrapperFactory<T>` interface?

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/steps/GetDeploymentData.java
```diff
@@ -34,19 +35,22 @@
...
 
     log.debug(
         "deployment input name {} and account {}",
-        deploymentInput.getName(),
-        deploymentInput.getAccount());
+        deploymentWorkflowInput.getName(),
+        deploymentWorkflowInput.getAccount());
 
+    //TODO: retrieve from deployment table
     final DeploymentData deploymentData =
```
wjbuys, line 48: What does the `Data` suffix mean? It doesn't seem to consistently mean "coming from a data store" (cf. `TaskData`), nor that it's representing some sort of raw data, so it's a bit confusing. Personally I'd omit it everywhere in favour of a more descriptive base name.

wjbuys (commented):
The basics look good, but we'll have to sync up on #258 for this, I think.

narehayrapetyan:
I think this is a good starting point no matter in which direction we go so it doesn't make sense blocking this until we finalize #258.

wjbuys:
Fair point.

wjbuys (approved):

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/steps/GetStateData.java
narehayrapetyan: good point

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/steps/GetStateData.java
narehayrapetyan: I was just trying to get things working with some initial inputs/outputs, I think we're going to have to change what we're passing through anyway depending on how much data we need to and can pass. If you have better names in mind please suggest.

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/wrapper/ECSWrapperFactory.java
narehayrapetyan: why do we need to reimplement the ecs interface?

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/steps/GetDeploymentData.java
narehayrapetyan: suggestions?

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/steps/GetStateData.java
narehayrapetyan: I was just trying to get things working with some initial inputs/outputs, I think we're going to have to change what we're passing through anyway depending on how much data we need to and can pass. If you have better names in mind please suggest.

wjbuys: Okay, we can revisit when we have more clarity on what these inputs/outputs will mean.

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/deployment/steps/GetDeploymentData.java
wjbuys: As per previous comment, we can revisit once we have more clarity on what inputs/outputs are needed here.

scheduling-manager-workflows/src/main/java/com/amazonaws/blox/schedulingmanager/wrapper/ECSWrapperFactory.java
wjbuys: Single responsibility: If the `CrossAccountAmazonECS` class is only doing auth-decoration, that's a single responsibility. If it's also changing method signatures, it's also acting as an adapter, which is an orthogonal responsibility.

Considering we'd probably want an ECS adapter interface (something like  `ECSCluster(clusterID)` so that we don't have to pass cluster ID in for every call, for example), mixing in the auth concern with the adaptation would make this harder to test than if those were separate classes.

narehayrapetyan: We do have two separate classes don't we? The wrapper factory deals with auth/role, the wrapper with ECS calls.

PR #260, opened 2017-09-29
Adding environment ddb table definition, model and dynamodbmapper def…

narehayrapetyan:
…inition

./gradlew build

#### Licensing


data-service/src/main/java/com/amazonaws/blox/dataservice/repository/model/DaemonEnvironmentLatestVersionDDBRecord.java
```diff
@@ -0,0 +1,47 @@
...
+import lombok.NoArgsConstructor;
+
+@DynamoDBTable(tableName = "DaemonEnvironmentLatestVersion")
+@Data
+@Builder
+//Required for builder because an empty constructor exists
+@AllArgsConstructor
+//Required for dynamodbmapper
+@NoArgsConstructor
+public class DaemonEnvironmentLatestVersionDDBRecord {
```
wjbuys: This doesn't look like it's specific to `Daemon` environments.

data-service/src/main/java/com/amazonaws/blox/dataservice/model/InstanceGroup.java
```diff
@@ -0,0 +1,29 @@
...
+import lombok.NonNull;
+import lombok.Value;
+
+import java.util.Set;
+
+@Value
+@Builder
+public class InstanceGroup {
+
+  @NonNull private String clusterArn;
```
wjbuys, line 27: I think we'll have to  make Environments a subresource of a Cluster, like Services are today (i.e. and Environment is tied to one and only one Cluster for its lifetime). The way [clusters are defined](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ECS_clusters.html), I can't think of a reason why a user would change the cluster an Environment is targeted at. Additionally, I think that cluster-changing deployment would be hard to implement, since we'd then have to consider state from all clusters where an environment has ever been deployed in order to undeploy tasks that may have lingered on those clusters.

Maybe put it on Environment instead?

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/model/DaemonEnvironmentDDBRecord.java
```diff
@@ -0,0 +1,90 @@
...
+import lombok.NoArgsConstructor;
+
+@DynamoDBTable(tableName = "DaemonEnvironments")
+@Data
+@Builder
+//Required for builder because an empty constructor exists
+@AllArgsConstructor
+//Required for dynamodbmapper
+@NoArgsConstructor
+public class DaemonEnvironmentDDBRecord {
```
wjbuys: Nothing in this class looks specific to `Daemon` environments.

data-service/templates/data_service.yml
```diff
@@ -0,0 +1,55 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Transform: AWS::Serverless-2016-10-31
+Resources:
+
+  DynamoDBTable:
+    Type: AWS::DynamoDB::Table
+    Properties:
+     #TODO: change ddb table names to have region prefix passed in as a resource
```
wjbuys, line 8: Why would we need a region prefix for the table name? We'll have separate accounts per-region, right?

data-service/src/main/java/com/amazonaws/blox/dataservice/model/InstanceGroup.java
```diff
@@ -0,0 +1,29 @@
...
+  private Set<Attribute> attributes;
```
wjbuys, line 28: How would this work? Today, the only instance grouping criteria available in the ECS API is through the Cluster Query Language, which is what the [`listContainerInstances`'s `filter` argument accepts](http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ListContainerInstances.html#ECS-ListContainerInstances-request-filter).

data-service/templates/data_service.yml
```diff
@@ -0,0 +1,55 @@
...
+           WriteCapacityUnits: 15
+     ProvisionedThroughput:
+       ReadCapacityUnits: 15
+       WriteCapacityUnits: 15
+
+  DynamoDBTable:
+    Type: AWS::DynamoDB::Table
+    Properties:
+     #TODO: change ddb table names to have region prefix passed in as a resource
+     TableName: DaemonEnvironmentsLatestVersion
```
wjbuys: I wonder if "Latest" is the right term here. We need to keep track of the "target" version, which may be much smaller than the "Latest" version, if there have been no deployments, but there has been many updates.

In fact, we'd probably want to keep multiple version pointers for each environment, such as "Latest" (last update), "Target" (what we're currently enforcing), and "Previous" (for quick rollbacks).

Nit: I find the "Environment"/"EnvironmentLatestVersion" naming confusing. Conceptually the resource is an "Environment" which *has-many* "Versions", so calling the top level resource "Environment" and the individual versions "EnvironmentVersion" seems more clear to me.

wjbuys (changes requested):
Looks reasonable, but I strongly suggest against having one table per environment type. Looking at the models defined here, none of them are particular to `Daemon` environments. I think `Daemon` shouldn't have to be more than just a `Type` field on the Environment record.

(We will, of course, have other strategy-esque classes that implement behaviour specific to daemon environments, but that doesn't need to be set up through different tables named "Daemon")

wjbuys (approved):
Actually, none of these are one way doors. The most important thing is the backend query requirements for this:

- listEnvironmentsByCluster(cluster_id) -> List(environment_id, environment_version)
- describeEnvironmentVersion(environment_id, version) -> (full details)

Or:

- listEnvironmentsByCluster(cluster_id) -> List(environment_id)
- describeEnvironment(version = Target, environment_id)

As long as we can answer those queries, all should be well. Apply the naming changes if they seem like an improvement, otherwise shipit.

data-service/src/main/java/com/amazonaws/blox/dataservice/model/InstanceGroup.java
narehayrapetyan: I don't understand, how does having the cluster in the instance group prevent it from being immutable?

data-service/templates/data_service.yml
narehayrapetyan: just makes our naming consistent since we can't omit the region for S3

narehayrapetyan:
How would we handle getting clusters for all daemon environments for example? One alternative I can think of is now that 5 local secondary indexes are supported we can make environment type an index. What do you think?

data-service/templates/data_service.yml
narehayrapetyan: I'm not sure I follow your naming strategy either.

wjbuys:
We wouldn't have to (and probably don't want to) do that for the backend, as long as the environment definition also contains the Type. At the time a scheduler receives the definition, it can look up the appropriate implementation by type.

data-service/src/main/java/com/amazonaws/blox/dataservice/model/InstanceGroup.java
wjbuys: It would be inconsistent at the API level to be able to change one part of the instance group, and not the other.

That's not a super strong argument though. I think being able to get environment by cluster is a better argument. If we indexed on cluster in this table, there would be no way to get just one record per environment when querying by cluster. And if we put it in both tables, then we can't atomically update it.

wjbuys:
Either way, we'll probably have to revisit this as we prove it out against the kinds of queries we'll have to do. I think it might be more productive to build the query APIs and take those learnings back to the model, than to anticipate them here.

data-service/templates/data_service.yml
wjbuys: 👍

Issue #261, opened 2017-10-02
Build scheduling manager steel thread


wjbuys:
The outcome from #238 was that we'll build the following:

We wire up the two event sources to the scheduling manager function:

![image](https://user-images.githubusercontent.com/31602/31101005-a1d0c746-a780-11e7-8569-6e993c6c23d3.png)

This needs to parallelize over clusters, somehow. We can either do that before the manager (i.e. an additional lambda function consumes only the CWE Cron Event, and then kicks off a manager per cluster), or parallelize inside the manager itself. For now, I'm just going to do this serially in the manager.

Once the manager has a single (cluster, environment) mapping, it invokes the scheduling function for that tuple. Since we don't want to re-query cluster state for every environment, we query it once from the manager, and pass it into the scheduling function along with the environment ID.

![image](https://user-images.githubusercontent.com/31602/31102090-80f91f32-a785-11e7-9483-3e8175251f09.png)

The scheduling function then calls the data service to describe the environment's target by ID. Using this description along with the cluster state, it calculates which tasks to launch/terminate, and calls the ECS control plane.

Finally, we may have to wait for all scheduling functions to complete in the manager function.

This issue is to set up the lambda function definitions and interfaces, with stub implementations.

- [x] set up local test harness
- [x] define lambda structure
- [x] figure out lambda-to-lambda call interface
- [x] configure cloudformation template for lambdas
- [x] SingleTask scheduling implementation
- [x] implement ECSState interface to get cluster state snapshot
- [x] integrate with DataService implementation

#### Followup tasks
These tasks are operational/incremental improvements that we should revisit immediately after this issue, broken out as cards on our planning board:
- [figure out error marshaling between functions](https://github.com/blox/blox/projects/2#card-5333492)
- [migrate from custom `CloudWatchEvent` type to built-in `ScheduledEvent` type](https://github.com/blox/blox/projects/2#card-5333521)
- [add API integration test suite that runs against beta stack (needs a beta stack)](https://github.com/blox/blox/projects/2#card-5333497)
- [add logging/error handling `RequestHandler` wrappers](https://github.com/blox/blox/projects/2#card-5333509)

wjbuys:
We've discussed this, and we probably need to have a separate lambda function for the call from the reconciliation event, so that we can parallelize on all clusters. I'll update the description above.

wjbuys:
[...]

PR #265, opened 2017-10-09
Work in progress: Add draft implementation of core scheduling lambda functions

wjbuys:
This adds the skeleton for #261.

Things I'll clean up before opening a PR on `blox:dev`:
- `@SneakyThrows`, will replace with typed exceptions
- better test coverage (unit + integration)
- clean up dependencies to depend on specific versions, and only what's actually used

Things I'd like feedback on:
- project structure/naming: how should we break these into projects? Are the names of the top-level functions okay? `Stater` seems particularly dumb.
- data service interfaces: I've opted for a more plain Java-ey interface, rather than using Input/Output types, but I'm willing to change that if it'll make things more clear/consistent.
- cluster state interfaces: I'm not sure if the `Id` class abstractions are helpful, open to replacing them with primitive types.

Things that are janky because they need to be:
- The JacksonRequestStreamHandler relies on reflection. I couldn't figure out how to do this without reflection. It's necessary for us to do custom serialization of the lambda inputs, this would be where we can add binary serialization and/or gzipping to fit cluster state into the Lambda request body size limits.


wjbuys:
Meant to open this against my own fork just to have a backup. I'll re-open individual smaller PRs for this future here.

PR #266, opened 2017-10-09
Adding create and get environments in environmentRepository

narehayrapetyan:
./gradlew build


data-service/src/main/java/com/amazonaws/blox/dataservice/arn/EnvironmentArnGenerator.java
```diff
@@ -0,0 +1,31 @@
...
+import lombok.Value;
+
+@Builder
+@Value
+public class EnvironmentArnGenerator {
+
+  private static final CharSequence DELIMITER = "_";
+
+  public static String generateEnvironmentArn(
+      final String environmentName, final String accountId) {
```
wjbuys, line 27: Should this signature also include the region? That's part of the ARN right?

data-service/src/main/java/com/amazonaws/blox/dataservice/model/Environment.java
```diff
@@ -15,19 +15,25 @@
...
 @Data
 @Builder
+// required for builder
+@AllArgsConstructor
+// required for mapstruct
+@NoArgsConstructor
 public class Environment {
 
-  @NonNull private String accountId;
+  @NonNull private String environmentId;
```
wjbuys, line 32: Is this an ID or an ARN?

data-service/src/main/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandler.java
```diff
@@ -0,0 +1,55 @@
...
+import com.amazonaws.blox.dataservicemodel.v1.exception.EnvironmentExistsException;
+import com.amazonaws.blox.dataservicemodel.v1.exception.EnvironmentNotFoundException;
+import java.util.List;
+import lombok.AllArgsConstructor;
+import lombok.NonNull;
+import org.springframework.stereotype.Component;
+
+@Component
+@AllArgsConstructor
+public class EnvironmentHandler {
```
wjbuys, line 29: How is this intended to be used? Is there another class that implements DataService and calls this handler? Or should this implement the DataService interface?

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/model/EnvironmentTargetVersionDDBRecord.java
```diff
@@ -15,33 +15,35 @@
...
-@DynamoDBTable(tableName = "EnvironmentLatestVersion")
+@DynamoDBTable(tableName = "EnvironmentTargetVersion")
 @Data
 @Builder
 //Required for builder because an empty constructor exists
 @AllArgsConstructor
 //Required for dynamodbmapper
 @NoArgsConstructor
-public class EnvironmentLatestVersionDDBRecord {
+public class EnvironmentTargetVersionDDBRecord {
```
wjbuys, line 33: Is this missing the actual version pointer to the target version of the Environment record?

data-service/src/main/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandler.java
```diff
@@ -0,0 +1,55 @@
...
+  public Environment describeEnvironment(
+      final String environmentId, final String environmentVersion) {
+    return null;
+  }
+
+  public List<String> listClustersWithEnvironments() {
+    return null;
+  }
+
+  public Environment getLatestEnvironmentVersion(final String environmentId)
```
wjbuys, line 51: I think you may want to implement this method as part of this review, since I can't see how you'd answer this question with the current data model.

wjbuys (commented):
I'm having trouble seeing how the data model could be used to implement the `handler.getLatestEnvironmentVersion` method; it might be useful to implement at least that method to validate that the model works.

data-service/src/main/java/com/amazonaws/blox/dataservice/arn/EnvironmentArnGenerator.java
narehayrapetyan: I'm not sure we can implement ARNs externally, hence the TODO

narehayrapetyan:
This is an incremental PR, there's no need to validate everything. If the model doesn't work in the next PR I'll modify it. Otherwise we have to implement everything at once.

wjbuys (approved):

PR #272, opened 2017-10-11
Add core Scheduling lambda functions that can place a single task

wjbuys:
This review adds the core Reconciler, Manager, and Scheduler lambda functions that comprise the Scheduling Manager, implementing #261.

- Reconciler: Reacts to periodic Cloudwatch events and kicks off a Manager instance for every cluster concurrently.
- Manager: Invoked per-cluster to retrieve a cluster state snapshot, and invoke a Scheduler instance for every Environment in that cluster.
- Scheduler: Invoked per-environment to make scheduling decisions (i.e. start/stop tasks)

A good place to get an overview of how this works, is in the `SchedulingSystemTest` test case. This wires up all the functions directly (without going through Lambda), and verifies that it starts a task.

In a production context, the functions  invoke each other through a `LambdaFunction<INPUT, OUTPUT>` interface, which has `*Async` methods that return `CompletableFuture` instance for fully concurrent operation. The concrete `AwsSdkV2LambdaFunction` implementation implements this interface with the new non-blocking Netty-based Lambda client, that allows us to reach 100s of concurrent invocations from a single lambda function.

In order to support more flexible serialization of the function inputs and outputs, this also implements a `JacksonRequestStreamHandler`, which allows us to customize the ObjectMapper to use for serializing/deserializing the input/output types. This would allow us to e.g. use Gzipped JSON or binary ION as the actual payload format between lambda functions.

The next steps not included in this review are:
- implement the actual ECSState interface to capture a cluster snapshot
- implement the actual DataService interface to get environment details

### DIFF
--- a/data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/client/DataService.java
+++ b/data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/client/DataService.java
@@ -20,10 +20,16 @@ import com.amazonaws.blox.dataservicemodel.v1.exception.EnvironmentVersionNotFou
 import com.amazonaws.blox.dataservicemodel.v1.exception.EnvironmentVersionOutdatedException;
 import com.amazonaws.blox.dataservicemodel.v1.exception.InvalidParameterException;
 import com.amazonaws.blox.dataservicemodel.v1.exception.ServiceException;
-import com.amazonaws.blox.dataservicemodel.v1.model.CreateEnvironmentRequest;
-import com.amazonaws.blox.dataservicemodel.v1.model.CreateEnvironmentResponse;
-import com.amazonaws.blox.dataservicemodel.v1.model.StartDeploymentRequest;
-import com.amazonaws.blox.dataservicemodel.v1.model.StartDeploymentResponse;
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.CreateEnvironmentRequest;
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.CreateEnvironmentResponse;
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.DescribeTargetEnvironmentRevisionResponse;
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.DescribeTargetEnvironmentRevisionRequest;
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.ListClustersRequest;
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.ListClustersResponse;
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.ListEnvironmentsRequest;
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.ListEnvironmentsResponse;
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.StartDeploymentRequest;
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.StartDeploymentResponse;
 
 public interface DataService {
 
@@ -35,4 +41,16 @@ public interface DataService {
   StartDeploymentResponse startDeployment(StartDeploymentRequest request)
       throws EnvironmentNotFoundException, EnvironmentVersionNotFoundException,
           EnvironmentVersionOutdatedException, InvalidParameterException, ServiceException;
+
+  /** Lists all environments that target a given cluster */
+  ListEnvironmentsResponse listEnvironments(ListEnvironmentsRequest request);
+
+  /** Lists all clusters managed by blox */
+  ListClustersResponse listClusters(ListClustersRequest request);
+
+  /**
+   * Describes the revision of the given EnvironmentVersion that is the current deployment Target.
+   */
+  DescribeTargetEnvironmentRevisionResponse describeTargetEnvironmentRevision(
+      DescribeTargetEnvironmentRevisionRequest request);
 }
--- a/data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/EnvironmentVersion.java
+++ b/data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/EnvironmentVersion.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may
+ * not use this file except in compliance with the License. A copy of the
+ * License is located at
+ *
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "LICENSE" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.blox.dataservicemodel.v1.model;
+
+import lombok.Builder;
+import lombok.Value;
+
+@Value
+@Builder
+public class EnvironmentVersion {
+  private final String environmentId;
+  private final String environmentVersion;
+  private final String taskDefinition;
+  private final DeploymentType deploymentType;
+  private final DeploymentConfiguration deploymentConfiguration;
+}
--- a/scheduling-manager/build.gradle
+++ b/scheduling-manager/build.gradle
@@ -12,6 +12,7 @@ repositories {
 dependencyManagement {
     imports {
         mavenBom 'com.amazonaws:aws-java-sdk-bom:1.11.180'
+        mavenBom 'software.amazon.awssdk:bom:2.0.0-preview-4'
     }
     dependencies {
         dependencySet(group: 'org.springframework', version: '4.3.10.RELEASE') {
@@ -24,11 +25,12 @@ dependencyManagement {
 }
 
 dependencies {
-    compile (
+    compile(
             'com.amazonaws:aws-lambda-java-core:1.1.0',
             'com.amazonaws:aws-lambda-java-events:1.3.0',
             'com.amazonaws:aws-lambda-java-log4j2:1.0.0',
 
+            'software.amazon.awssdk:ecs',
             'org.projectlombok:lombok:1.16.18',
 
             'org.apache.logging.log4j:log4j-slf4j-impl:2.8.+',
@@ -38,8 +40,12 @@ dependencies {
             'org.springframework:spring-beans',
             'org.springframework:spring-context',
 
+            'com.spotify:completable-futures:+',
+            project(":data-service-model"),
     )
 
+    testCompile "org.mockito:mockito-core:2.+"
+    testCompile "org.hamcrest:hamcrest-junit:2.+"
     testCompile group: 'junit', name: 'junit', version: '4.12'
 }
 
--- a/scheduling-manager/build.gradle
+++ b/scheduling-manager/build.gradle
@@ -48,10 +48,12 @@ dependencies {
             project(":data-service-model"),
     )
 
-    testCompile 'org.springframework:spring-test'
-    testCompile "org.mockito:mockito-core:2.+"
-    testCompile "org.hamcrest:hamcrest-junit:2.+"
-    testCompile group: 'junit', name: 'junit', version: '4.12'
+    testCompile(
+            'org.springframework:spring-test',
+            'org.mockito:mockito-core:2.+',
+            'org.hamcrest:hamcrest-junit:2.+',
+            'junit:junit:4.12'
+    )
 }
 
 task buildZip(type: Zip) {
--- a/scheduling-manager/build.gradle
+++ b/scheduling-manager/build.gradle
@@ -31,6 +31,10 @@ dependencies {
             'com.amazonaws:aws-lambda-java-log4j2:1.0.0',
 
             'software.amazon.awssdk:ecs',
+            'software.amazon.awssdk:lambda',
+
+            'org.apache.commons:commons-lang3:3.6+',
+
             'org.projectlombok:lombok:1.16.18',
 
             'org.apache.logging.log4j:log4j-slf4j-impl:2.8.+',
@@ -44,6 +48,7 @@ dependencies {
             project(":data-service-model"),
     )
 
+    testCompile 'org.springframework:spring-test'
     testCompile "org.mockito:mockito-core:2.+"
     testCompile "org.hamcrest:hamcrest-junit:2.+"
     testCompile group: 'junit', name: 'junit', version: '4.12'
--- a/scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/manager/ManagerApplication.java
+++ b/scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/manager/ManagerApplication.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may
+ * not use this file except in compliance with the License. A copy of the
+ * License is located at
+ *
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "LICENSE" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.blox.scheduling.manager;
+
+import com.amazonaws.blox.lambda.LambdaFunction;
+import com.amazonaws.blox.lambda.AwsSdkV2LambdaFunction;
+import com.amazonaws.blox.scheduling.SchedulingApplication;
+import com.amazonaws.blox.scheduling.scheduler.SchedulerInput;
+import com.amazonaws.blox.scheduling.scheduler.SchedulerOutput;
+import com.amazonaws.blox.scheduling.state.ECSStateClient;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ComponentScan("com.amazonaws.blox.scheduling.manager")
+public class ManagerApplication extends SchedulingApplication {
+
+  // Wired in through environment variable in CloudFormation template
+  @Value("${scheduler_function_name}")
+  String schedulerFunctionName;
+
+  @Bean
+  public LambdaFunction<SchedulerInput, SchedulerOutput> scheduler() {
+    return new AwsSdkV2LambdaFunction<>(
+        lambdaClient(), mapper(), SchedulerOutput.class, schedulerFunctionName);
+  }
+
+  @Bean
+  public ECSStateClient ecsState() {
+    return new ECSStateClient();
+  }
+}
--- a/scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/manager/ManagerHandler.java
+++ b/scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/manager/ManagerHandler.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may
+ * not use this file except in compliance with the License. A copy of the
+ * License is located at
+ *
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "LICENSE" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.blox.scheduling.manager;
+
+import com.amazonaws.blox.dataservicemodel.v1.client.DataService;
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.ListEnvironmentsRequest;
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.ListEnvironmentsResponse;
+import com.amazonaws.blox.lambda.LambdaFunction;
+import com.amazonaws.blox.scheduling.scheduler.SchedulerInput;
+import com.amazonaws.blox.scheduling.scheduler.SchedulerOutput;
+import com.amazonaws.blox.scheduling.state.ClusterSnapshot;
+import com.amazonaws.blox.scheduling.state.ECSState;
+import com.amazonaws.services.lambda.runtime.Context;
+import com.amazonaws.services.lambda.runtime.RequestHandler;
+import com.spotify.futures.CompletableFutures;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.stream.Stream;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.log4j.Log4j2;
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor
+@Log4j2
+public class ManagerHandler implements RequestHandler<ManagerInput, ManagerOutput> {
+  private final DataService data;
+  private final ECSState ecs;
+  private final LambdaFunction<SchedulerInput, SchedulerOutput> scheduler;
+
+  @Override
+  public ManagerOutput handleRequest(ManagerInput input, Context context) {
+    ListEnvironmentsResponse r =
+        data.listEnvironments(new ListEnvironmentsRequest(input.getClusterArn()));
+    List<String> environments = r.getEnvironmentIds();
+
+    ClusterSnapshot state = ecs.snapshotState(input.getClusterArn());
+
+    Stream<CompletableFuture<SchedulerOutput>> pendingRequests =
+        environments
+            .stream()
+            .map(environmentId -> scheduler.callAsync(new SchedulerInput(state, environmentId)));
+
+    List<SchedulerOutput> outputs = pendingRequests.collect(CompletableFutures.joinList()).join();
+
+    return new ManagerOutput(input.getClusterArn(), outputs);
+  }
+}
--- a/scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/reconciler/CloudWatchEvent.java
+++ b/scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/reconciler/CloudWatchEvent.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may
+ * not use this file except in compliance with the License. A copy of the
+ * License is located at
+ *
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "LICENSE" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.blox.scheduling.reconciler;
+
+import com.amazonaws.regions.Regions;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.time.Instant;
+import java.util.List;
+import lombok.Data;
+
+/**
+ * Representation of an Event from Cloudwatch.
+ *
+ * <p>See http://docs.aws.amazon.com/AmazonCloudWatch/latest/events/EventTypes.html for details.
+ *
+ * @param <T> The type of the "detail" field.
+ */
+@Data
+public class CloudWatchEvent<T> {
+
+  private String version;
+  private String id;
+  private String account;
+  private String source;
+  private Instant time;
+  private Regions region;
+  private List<String> resources;
+  private T detail;
+
+  @JsonProperty("detail-type")
+  private String detailType;
+
+  /**
+   * Setter for Jackson to deserialize Regions enum from string
+   *
+   * @param region
+   */
+  public void setRegion(String region) {
+    this.region = Regions.fromName(region);
+  }
+}
--- a/scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/reconciler/ReconcilerHandler.java
+++ b/scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/reconciler/ReconcilerHandler.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may
+ * not use this file except in compliance with the License. A copy of the
+ * License is located at
+ *
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "LICENSE" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.blox.scheduling.reconciler;
+
+import com.amazonaws.blox.dataservicemodel.v1.client.DataService;
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.ListClustersRequest;
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.ListClustersResponse;
+import com.amazonaws.blox.lambda.LambdaFunction;
+import com.amazonaws.blox.scheduling.manager.ManagerInput;
+import com.amazonaws.blox.scheduling.manager.ManagerOutput;
+import com.amazonaws.services.lambda.runtime.Context;
+import com.amazonaws.services.lambda.runtime.RequestHandler;
+import com.spotify.futures.CompletableFutures;
+import java.util.List;
+import java.util.Map;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.log4j.Log4j2;
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor
+@Log4j2
+public class ReconcilerHandler implements RequestHandler<CloudWatchEvent<Map>, Void> {
+  final DataService dataService;
+  final LambdaFunction<ManagerInput, ManagerOutput> stateFunction;
+
+  @Override
+  public Void handleRequest(CloudWatchEvent<Map> input, Context context) {
+    ListClustersResponse r = dataService.listClusters(new ListClustersRequest());
+    List<String> clusters = r.getClusterArns();
+
+    clusters
+        .stream()
+        .map(c -> stateFunction.triggerAsync(new ManagerInput(c)))
+        .collect(CompletableFutures.joinList())
+        .join();
+
+    return null;
+  }
+}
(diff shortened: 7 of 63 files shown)


wjbuys:
The failing test is because of two eagerly instantiated beans for AWS clients. I'm working on disabling those for tests now.

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/client/DataService.java
```diff
@@ -35,4 +41,16 @@ CreateEnvironmentResponse createEnvironment(CreateEnvironmentRequest request)
...
+  /** Lists all environments that target a given cluster */
+  ListEnvironmentsResponse listEnvironments(ListEnvironmentsRequest request);
+
+  /** Lists all clusters managed by blox */
+  ListClustersResponse listClusters(ListClustersRequest request);
+
+  /**
+   * Describes the revision of the given EnvironmentRevision that is the current deployment Target.
+   */
+  DescribeEnvironmentRevisionResponse describeTargetEnvironmentRevision(
```
narehayrapetyan: should the request and response objects have the same name? so describeTargetEnvironmentResponse instead of describeEnvironmentRevisionResponse?

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/EnvironmentRevision.java
```diff
@@ -0,0 +1,28 @@
...
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.blox.dataservicemodel.v1.model;
+
+import lombok.Builder;
+import lombok.Value;
+
+@Value
+@Builder
+public class EnvironmentRevision {
```
narehayrapetyan: Is this a public type? Where is this used?

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/EnvironmentRevision.java
```diff
@@ -0,0 +1,28 @@
...
+  private final String environmentArn;
```
narehayrapetyan: we should probably not name it arns until we know whether we can use arns or not

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/EnvironmentRevision.java
```diff
@@ -0,0 +1,28 @@
...
+  // TODO: Maybe this should just be an environmentRevisionArn, instead of a separate arn and version?
+  private final String environmentVersion;
+  private final DeploymentMethod deploymentMethod;
```
narehayrapetyan: What's this for?

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/operations/CreateEnvironmentRequest.java
```diff
@@ -12,8 +12,9 @@
  * express or implied. See the License for the specific language governing
  * permissions and limitations under the License.
  */
-package com.amazonaws.blox.dataservicemodel.v1.model;
+package com.amazonaws.blox.dataservicemodel.v1.model.operations;
```
narehayrapetyan: This isn't an operation, it's a model type

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/operations/DescribeEnvironmentRevisionResponse.java
```diff
@@ -0,0 +1,23 @@
...
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.blox.dataservicemodel.v1.model.operations;
+
+import com.amazonaws.blox.dataservicemodel.v1.model.EnvironmentRevision;
+import lombok.Value;
+
+@Value
+public class DescribeEnvironmentRevisionResponse {
+  private final EnvironmentRevision environmentRevision;
```
narehayrapetyan: I don't think we should return an object here so we can change the fields of different response objects independently. So if we need to change describEnvResponse we don't have to change createEnvResponse (ran into this problem before)

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/operations/DescribeTargetEnvironmentRevisionRequest.java
```diff
@@ -0,0 +1,23 @@
...
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.blox.dataservicemodel.v1.model.operations;
+
+import lombok.Value;
+
+@Value
+public class DescribeTargetEnvironmentRevisionRequest {
+
+  private final String environmentArn;
```
narehayrapetyan: We should just take the environmentName and construct the id/arn internally

narehayrapetyan (changes requested):

scheduling-manager/build.gradle
```diff
@@ -31,6 +31,10 @@ dependencies {
             'com.amazonaws:aws-lambda-java-log4j2:1.0.0',
 
             'software.amazon.awssdk:ecs',
+            'software.amazon.awssdk:lambda',
```
narehayrapetyan: what's this dependency? shouldn't it be com.amazonaws:aws-java-sdk-lambda? http://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/setup-project-gradle.html

scheduling-manager/build.gradle
```diff
@@ -44,6 +48,7 @@ dependencies {
             project(":data-service-model"),
     )
 
+    testCompile 'org.springframework:spring-test'
```
narehayrapetyan: nit: these can also be grouped into testCompile {..}

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/EnvironmentRevision.java
wjbuys: Oh, there's a possibility we can't use Arns? I missed that.

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/EnvironmentRevision.java
wjbuys: See the next commit with the core schedulers, it's used by the inner scheduler loop.

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/operations/CreateEnvironmentRequest.java
wjbuys: Hm. Any better categorizations for the types that only serve as input/output wrappers? I don't really think of those as `model` types, and it makes it hard to see the real domain types.

scheduling-manager/build.gradle
narehayrapetyan: Ohh is this the v2?

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/operations/DescribeTargetEnvironmentRevisionRequest.java
wjbuys: Why? If ARNs are the canonical identifier for our resources, shouldn't we use them? In particular, we do need at least `(accountId, environmentName)` in order to make the correct determination here.

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/operations/DescribeEnvironmentRevisionResponse.java
wjbuys: Could you give an example where that would apply to EnvironmentRevisions specifically? I can sort of see the use case for entities that are mutable, and can have some properties unset at creation time, but since EnviornmentRevisions are immutable, it seems like we couldn't have that situation.

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/client/DataService.java
wjbuys: Yeah, I was wondering about that. It looks like both `describeTargetEnvironmentRevision` and `describeEnvironmentRevision` would return exactly the same response, but I never ended up needing the `describeEnvironmentRevision` API call, so I removed it from the interface. I'll rename this, and we can rename/reuse the type later if needed.

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/operations/DescribeTargetEnvironmentRevisionRequest.java
wjbuys: Hm, thinking about this a bit more, we may want to put more/less in the EnvironmenID than is in the Arn (e.g. cluster ID). Do you think it's worth making an EnvironmentId value type? Then all the logic for converting to/from ARNs and such can live in there.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/manager/ManagerApplication.java
```diff
@@ -0,0 +1,46 @@
...
+import com.amazonaws.blox.scheduling.scheduler.SchedulerOutput;
+import com.amazonaws.blox.scheduling.state.ECSStateClient;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ComponentScan("com.amazonaws.blox.scheduling.manager")
+public class ManagerApplication extends SchedulingApplication {
```
narehayrapetyan, line 30: you don't need to extend another configuration class, you can just import it: https://docs.spring.io/spring-javaconfig/docs/1.0.0.M4/reference/html/ch04s03.html

narehayrapetyan (changes requested):

narehayrapetyan:
All right, got through 3 commits. Have not looked at "lambda scheduling core"  yet

scheduling-manager/build.gradle
wjbuys: Yep, that's exactly it: http://docs.aws.amazon.com/sdk-for-java/v2/developer-guide/welcome.html

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/manager/ManagerApplication.java
wjbuys: Ah. How do you reference beans from the base config in that case? E.g. on line 37 the `scheduler()` bean references `lambdaClient()` from the base class.

wjbuys: Looks like you can @Autowire the base Config instance; but then it's not clear how I'd override the beans from the base class in tests...

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/manager/ManagerHandler.java
```diff
@@ -0,0 +1,60 @@
...
+import com.amazonaws.services.lambda.runtime.RequestHandler;
+import com.spotify.futures.CompletableFutures;
+import java.util.List;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.log4j.Log4j2;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor(onConstructor = @__({@Autowired}))
```
narehayrapetyan: you can just have @allargsconstructor and it autowires it without the onConstructor annotation. not sure if you can skip it with @requiredargs

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/manager/ManagerHandler.java
```diff
@@ -0,0 +1,60 @@
...
+@Log4j2
+public class ManagerHandler implements RequestHandler<ManagerInput, ManagerOutput> {
+
+  final DataService data;
```
narehayrapetyan: private? (everywhere)

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/reconciler/CloudWatchEvent.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Representation of an Event from Cloudwatch.
+ *
+ * <p>See http://docs.aws.amazon.com/AmazonCloudWatch/latest/events/EventTypes.html for details.
+ *
+ * @param <T> The type of the "detail" field.
+ */
+@Data
+public class CloudWatchEvent<T> {
```
narehayrapetyan, line 31: why not use the provided type so it's kept up to date? (https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/ScheduledEvent.java)

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/reconciler/ReconcilerHandler.java
```diff
@@ -0,0 +1,54 @@
...
+@RequiredArgsConstructor(onConstructor = @__({@Autowired}))
+@Log4j2
+public class ReconcilerHandler implements RequestHandler<CloudWatchEvent<Map>, Void> {
+
+  final DataService dataService;
+  final LambdaFunction<ManagerInput, ManagerOutput> stateFunction;
+
+  @Override
+  public Void handleRequest(CloudWatchEvent<Map> input, Context context) {
+    ListClustersResponse r = dataService.listClusters(new ListClustersRequest());
```
narehayrapetyan, line 41: how are we thinking about handling exceptions? logging? (everywhere there are service calls)

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/SchedulerHandler.java
```diff
@@ -0,0 +1,72 @@
...
+@RequiredArgsConstructor(onConstructor = @__({@Autowired}))
+@Log4j2
+public class SchedulerHandler implements RequestHandler<SchedulerInput, SchedulerOutput> {
+  private final DataService data;
+  private final ECSAsyncClient ecs;
+  private final SchedulerFactory schedulerFactory;
+
+  @Override
+  public SchedulerOutput handleRequest(SchedulerInput input, Context context) {
+    log.debug("Request: {}", input);
```
narehayrapetyan, line 46: delete?

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/SchedulerHandler.java
```diff
@@ -0,0 +1,72 @@
...
+
+    DescribeEnvironmentRevisionResponse r =
```
narehayrapetyan: I usually use finals for all variables

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/StartTask.java
```diff
@@ -0,0 +1,43 @@
...
+import lombok.Value;
+import software.amazon.awssdk.services.ecs.ECSAsyncClient;
+import software.amazon.awssdk.services.ecs.model.StartTaskRequest;
+import software.amazon.awssdk.services.ecs.model.StartTaskResponse;
+
+@Value
+public class StartTask implements SchedulingAction {
+  private final String clusterArn;
+  private final String containerInstanceArn;
+  private final String taskDefinition;
```
narehayrapetyan: if everything's named arn why not this one?

scheduling-manager/src/test/java/com/amazonaws/blox/scheduling/SchedulingSystemTest.java
```diff
@@ -0,0 +1,119 @@
...
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+import software.amazon.awssdk.services.ecs.ECSAsyncClient;
+import software.amazon.awssdk.services.ecs.model.StartTaskRequest;
+import software.amazon.awssdk.services.ecs.model.StartTaskResponse;
+
+@Log4j2
+public class SchedulingSystemTest {
+
+  private static final String CLUSTER_ARN = "arn:::::cluster1";
```
narehayrapetyan, line 60: we're going to have to add validation for inputs so these will fail

scheduling-manager/src/test/java/com/amazonaws/blox/scheduling/SchedulingSystemTest.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.concurrent.CompletableFuture;
+import lombok.extern.log4j.Log4j2;
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+import software.amazon.awssdk.services.ecs.ECSAsyncClient;
+import software.amazon.awssdk.services.ecs.model.StartTaskRequest;
+import software.amazon.awssdk.services.ecs.model.StartTaskResponse;
+
+@Log4j2
+public class SchedulingSystemTest {
```
narehayrapetyan, line 58: which file is this unit test for? we also need to unit test failure cases and in general all paths (for all unit tests files in this PR)

scheduling-manager/src/test/java/com/amazonaws/blox/scheduling/manager/ManagerHandlerTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+  ArgumentCaptor<SchedulerInput> schedulerArgument = ArgumentCaptor.forClass(SchedulerInput.class);
+  @Mock private LambdaFunction<SchedulerInput, SchedulerOutput> scheduler;
+  @Mock private ECSState ecs;
+  @Mock private DataService dataService;
+
+  @Test
+  @SuppressWarnings("unchecked")
+  public void invokesSchedulerForAllEnvironments() {
+    when(dataService.listEnvironments(new ListEnvironmentsRequest("arn:::::cluster1")))
```
narehayrapetyan: let's move the strings into constants

narehayrapetyan (changes requested):

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/manager/ManagerHandler.java
```diff
@@ -0,0 +1,60 @@
...
+        data.listEnvironments(new ListEnvironmentsRequest(input.getClusterArn()));
+    List<String> environments = r.getEnvironmentArns();
+
+    ClusterSnapshot state = ecs.snapshotState(input.getClusterArn());
+
+    List<SchedulerOutput> schedulerResults =
+        environments
+            .stream()
+            .map(environmentId -> scheduler.callAsync(new SchedulerInput(state, environmentId)))
+            .collect(CompletableFutures.joinList())
```
narehayrapetyan: so how does this work? does this block until all the results are available?

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/operations/CreateEnvironmentRequest.java
narehayrapetyan: You can name it something else but something that signifies a type and not a method/operation

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/StartTask.java
wjbuys: Good catch, will fix

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/operations/DescribeEnvironmentRevisionResponse.java
narehayrapetyan: It was more of a general comment for all response types but for this one, I'm not sure, what if we add some timestamp or metadata later?

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/operations/DescribeTargetEnvironmentRevisionRequest.java
narehayrapetyan: Yes, this is already checked in the environmentId/arnGenerator

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/manager/ManagerApplication.java
narehayrapetyan: why do you need to override anything? unit tests shouldn't rely on config and integ tests should use the exact same config so we're not testing something different that's going to go into prod

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/SchedulerHandler.java
wjbuys: I was meaning to ask you about that. If I recall correctly, after Java 5 or 6 the compiler became smart enough to correctly infer `final` from context, and the only technical reason to use `final` for locals is to allow the variables to be captured in an anonymous function. Since then I stopped using `final` for locals, since it makes things noisy and harder to read, unless it's a *really* long method and there's some risk of the variable being modified in the body somewhere.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/reconciler/ReconcilerHandler.java
wjbuys: I'd like to put the error handling/request logging in middleware, probably between this handler and the top-level JacksonRequestStreamHandler. That's probably next after the state client. I've added a task to the issue for that.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/reconciler/CloudWatchEvent.java
wjbuys: Oh! I couldn't find that when I went looking, will update.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/manager/ManagerHandler.java
wjbuys: Will do.

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/operations/CreateEnvironmentRequest.java
wjbuys: Okay, how about `wrappers`?

scheduling-manager/build.gradle
wjbuys: Will do

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/SchedulerHandler.java
wjbuys: Will do once we have that logging handler (added a task for it to #261 ).

scheduling-manager/src/test/java/com/amazonaws/blox/scheduling/SchedulingSystemTest.java
wjbuys: This is, I guess, an integration test of the entire scheduling subsystem, to make sure that everything works when they're plugged together. But since it doesn't actually run in Lambda, and does still mock out services at the boundary of our code, it doesn't sound quite right to call it "integration". If it makes more sense, I can call it "SchedulingManagerFastIntegrationTest"?

scheduling-manager/src/test/java/com/amazonaws/blox/scheduling/manager/ManagerHandlerTest.java
wjbuys: Ah, did that for some other cases, didn't get around to this one.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/manager/ManagerHandler.java
wjbuys: It will rapidly call `callAsync` for all environments, which immediately returns a `CompletableFuture`. The `.collect(CompletableFutures.joinList())` line will then efficiently wait for all the futures to complete, and return a single `CompletableFuture` that yields the list of all the values of the futures from `callAsync`.

Thinking about it, I think it's worth breaking this up to make it clearer what's happening like so:

```java
Stream<CompletableFuture<SchedulerOutput>> pendingRequests = environments.stream().map(...);
CompletableFuture<List<SchedulerOutput>> pendingResponses = pendingRequests.collect(CompletableFutures.joinList());
List<SchedulerOutput> results = pendingResponses.join();
```

Would that be clearer?

scheduling-manager/src/test/java/com/amazonaws/blox/scheduling/SchedulingSystemTest.java
wjbuys: Those are actually valid ARNs according to the spec. If we want to really validate inputs, we'll need a robust ARN parser.

I have some ideas for that I'd like to talk through, mind if we come back to that?

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/manager/ManagerHandler.java
wjbuys: Well that's weird: that totally failed for me initially, but I just tested it, and it now works without the `onConstructor` incantation (no need for `@AllArgsConstructor`). Will remove.

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/operations/DescribeEnvironmentRevisionResponse.java
wjbuys: Then those would be sibling members of the `environmentRevision` property on the response object? The only thing I can think of is something that shows e.g. environment status, but that would be an `EnvironmentRevisionStatus` object then, not a revision.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/manager/ManagerApplication.java
wjbuys: Again, these are in that fuzzy area between unit/integration tests, in that they don't need access to production resources, but they do check that most of our spring config is sane.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/reconciler/CloudWatchEvent.java
wjbuys: Ah, it looks like that is in the 2.0.0 version of the events package, which was released after I started working on this. Will have to upgrade to that version first.

wjbuys: Aaaand that one doesn't play nicely with the Jackson serializer, as caught by ReconcilerEntryPointTest:

```
com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "version" (class com.amazonaws.services.lambda.runtime.events.ScheduledEvent), not marked as ignorable (8 known properties: "detail", "region", "resources", "detailType", "account", "source", "time", "id"])
 at [Source: java.io.BufferedInputStream@1cb96; line: 2, column: 15] (through reference chain: com.amazonaws.services.lambda.runtime.events.ScheduledEvent["version"])
```

I'll dig into it a bit, I wonder if that works in Lambda.

wjbuys: Looks like they have some internal Jackson magic to make it work. I think I'm just going to rip out the JacksonRequestStreamHandler for this function, and use the POJO request handler instead.

scheduling-manager/src/test/java/com/amazonaws/blox/scheduling/SchedulingSystemTest.java
narehayrapetyan: Why have these not quite unit not quite integration tests if we're going to have both integration and unit tests? It seems like double coverage

scheduling-manager/src/test/java/com/amazonaws/blox/scheduling/SchedulingSystemTest.java
narehayrapetyan: sure

scheduling-manager/src/test/java/com/amazonaws/blox/scheduling/SchedulingSystemTest.java
wjbuys: As mentioned, this will either evolve into the scheduling system test that works against mocked dependencies, or get folded into integration tests once we have those (task added to #261).

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/manager/ManagerApplication.java
wjbuys: As mentioned, these are taking the place of the nonexistent integ tests right now, they will fold into those once we have them.

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/EnvironmentRevision.java
wjbuys: It's unclear what resolution you would like for this comment. Would you prefer that this be named `environmentId`? Or something else?

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/reconciler/CloudWatchEvent.java
wjbuys: This is proving more difficult than I thought. I've broken out a task for it in the issue to come back after this PR.

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/EnvironmentRevision.java
narehayrapetyan: Yes environmentId is fine and that's what it's called in the dataservice. Also it's called environmentVersion everywhere not revision

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/operations/CreateEnvironmentRequest.java
narehayrapetyan: sure

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/operations/DescribeEnvironmentRevisionResponse.java
narehayrapetyan: yes siblings

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/reconciler/ReconcilerHandler.java
narehayrapetyan: Did you figure out how the checked exceptions from dependencies are being thrown here? Can we wrap it in an application specific exception here before logging in middleware?

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/SchedulerHandler.java
narehayrapetyan: Can you point me to the compiler context referral? Quick googling didn't surface it

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/manager/ManagerHandler.java
narehayrapetyan: So it does wait for futures to complete and return a result? I don't think that's unclear what I'm asking is then this is essentially a blocking method, correct?

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/SchedulerHandler.java
wjbuys: Yeah, it's kinda buried in the JLS:

From [JLS 4.12.4](http://docs.oracle.com/javase/specs/jls/se8/html/jls-4.html#jls-4.12.4):
> A local variable whose declarator has an initializer (§14.4.2) is effectively final if all of the following are true:
> * It is not declared final.
> * It never occurs as the left hand side in an assignment expression (§15.26). (Note that the local variable declarator containing the initializer is not an assignment expression.)
> * It never occurs as the operand of a prefix or postfix increment or decrement operator (§15.14, §15.15).

It does look like I misremembered when this was added, though; looks like it only made it into the spec for Java 8.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/manager/ManagerHandler.java
wjbuys: Yes, correct, this method will block on the last `.join()` of the composite future until all requests are complete. Since all methods are using nonblocking IO, this more-or-less boils down to blocking for as long as it takes for the slowest of the calls to complete.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/reconciler/ReconcilerHandler.java
wjbuys: The answer is that they aren't. The `listClusters` API call does not declare any checked exceptions in this PR. Once that is added to the method signature, then we can indeed to some exception wrapping either in the middleware, or at this callsite.

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/EnvironmentRevision.java
wjbuys: Hm, it looks like I mixed up my terminology here. I was this should be `DeploymentType`, as per [the UX doc](https://github.com/blox/blox/blob/dev/docs/deployment-ux.md#deploymenttypes).

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/EnvironmentRevision.java
wjbuys: Okay, I'll update these to id and version.

wjbuys:
I'll squash all these commits together before merging, I've just been pushing one commit per review comment addressed.

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/EnvironmentVersion.java
```diff
@@ -19,11 +19,10 @@
...
 @Builder
-public class EnvironmentRevision {
-  private final String environmentArn;
-  // TODO: Maybe this should just be an environmentRevisionArn, instead of a separate arn and version?
+public class EnvironmentVersion {
+  private final String environmentId;
   private final String environmentVersion;
   private final String taskDefinition;
-  private final DeploymentMethod deploymentMethod;
+  private final DeploymentType deploymentType;
```
narehayrapetyan: why do we need this here?

narehayrapetyan (approved):

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/EnvironmentVersion.java
wjbuys: So that we can determine which scheduling algorithm to use once we reach the `Scheduler` function: https://github.com/blox/blox/pull/272/files#diff-0e6557774ed72aae34818e81eafbb6ebR19

PR #273, opened 2017-10-12
Adding environment and environment target version handler and reposit…

narehayrapetyan:
…ory logic

./gradlew build which also runs the integration tests

@wjbuys To make it easier for you to review, I implemented these changes top down: integration tests, all the business logic and repo interactions. Hopefully this gives you a better idea of what's going on.  The things I skipped that need to be implemented in the next pass are:

* api layer with request/response objects
* unit tests
* integration tests interact with api layer and not the handler directly
* non-happy path integration tests
* integration tests need to be moved to cucumber

Let me know if this works better for you.


data-service/src/main/java/com/amazonaws/blox/dataservice/repository/model/EnvironmentTargetVersionDDBRecord.java
```diff
@@ -46,4 +61,24 @@
   private String cluster;
 
   @DynamoDBVersionAttribute private Long recordVersion;
+
+  private String environmentVersion;
+
+  @DynamoDBTypeConverted(converter = InstantDDBConverter.class)
```
wjbuys: What's the benefit we're getting from duplicating the other fields of EnvironmentDDBRecord in EnvironmentTargetVersionDDBRecord? Since the individual versioned environments are immutable, we could do a two-step lookup for environmentVersion from this table, and the actual values from the other table.

That would save us from having to keep the two data models in sync, and potentially allow us to keep pointers to other useful versions in this table for other uses (e.g. "previousVersion" for rollbacks without an explicit version, or "lastCreatedVersion" for keeping track of the newest update to the environment in the absence of ordered version numbers.

data-service/src/main/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandler.java
```diff
@@ -30,26 +31,71 @@
...
+          String.format(
+              "Exception occurred when getting environment with id %s and version %s",
+              environmentId, environmentVersion),
+          e);
+    }
   }
 
-  public List<String> listClustersWithEnvironments() {
-    return null;
+  public List<String> listEnvironmentIdsWithCluster(final String cluster) throws ServiceException {
```
wjbuys: Since most AWS API's that start with `list` return `Id`s of some sort, should this maybe just be `listEnvironmentsWithCluster`?

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
```diff
@@ -76,4 +83,81 @@ public Environment getEnvironment(
 
     return environmentMapper.toEnvironment(environmentDDBRecord);
   }
+
+  @Override
+  public Environment createEnvironmentTargetVersion(final Environment environment)
+      throws StorageException, EnvironmentExistsException {
```
wjbuys: It doesn't look like this code actually ever throws an EnvironmentExistsException?

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
```diff
@@ -76,4 +83,81 @@ public Environment getEnvironment(
...
+    }
+
+    return environmentMapper.toEnvironment(environmentTargetVersionDDBRecord);
+  }
+
+  @Override
+  public List<String> listClusters() throws StorageException {
+    try {
+      List<EnvironmentTargetVersionDDBRecord> scanResult =
+          dynamoDBMapper.scan(
```
wjbuys: Can't we get this from scanning the cluster GSI instead of scanning through the entire EnvironmentTargetVersion table?

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
```diff
@@ -76,4 +83,81 @@ public Environment getEnvironment(
...
+              EnvironmentTargetVersionDDBRecord.class, new DynamoDBScanExpression());
+      //TODO: handle pagination. paginatedlist returns more elements on demand does stream accomplish that?
```
wjbuys: `stream` does it on-demand, but `collect` is a terminal stream operation, so it will load everything, eventually.

If we really want to eager-load all results before mapping over them, there's `PaginatedList.loadAllResults()`, but I can't see much benefit to that.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
```diff
@@ -76,4 +83,81 @@ public Environment getEnvironment(
...
+            .withIndexName(EnvironmentTargetVersionDDBRecord.ENVIRONMENT_CLUSTER_GSI_NAME)
+            .withConsistentRead(false)
+            .withHashKeyValues(EnvironmentTargetVersionDDBRecord.withGSIHashKey(cluster));
+
+    try {
+      //TODO: handle pagination. paginatedlist returns more elements on demand does stream accomplish that?
+      List<EnvironmentTargetVersionDDBRecord> queryResult =
+          dynamoDBMapper.query(EnvironmentTargetVersionDDBRecord.class, queryExpression);
+
+      if (queryResult.isEmpty()) {
```
wjbuys: Why the special edge case for `queryResult.isEmpty()`? AFAICT, the mapping on line 157 will have the same effect if the source list is empty.

data-service/src/main/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandler.java
```diff
@@ -30,26 +31,71 @@
...
+          String.format(
+              "Exception occurred when creating environment with id %s and version %s",
+              environment.getEnvironmentId(), environment.getEnvironmentVersion()),
+          e);
+    }
   }
 
-  public Environment describeEnvironment(final String environmentId) {
-    return null;
+  public Environment createEnvironmentTargetVersion(final Environment environment)
```
wjbuys: nit: Is this doing a `create`, or an upsert? It looks like the latter; in which case the name might be better off as something that more closely corresponds to the semantics, like `setEnvironmentVersionAsTarget`, or something similar.

data-service/src/main/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandler.java
```diff
@@ -30,26 +31,71 @@
...
+      throws EnvironmentExistsException, ServiceException {
+    try {
+      return environmentRepository.createEnvironmentTargetVersion(environment);
```
wjbuys: Tracing this through, it looks like this makes it possible for callers to set totally different values on the `environment` being passed in than what is actually in the `Environment` table. Should this method perhaps only take `(environmentId, environmentVersion)` as input?

wjbuys (commented):
That was nice and easy to review, thanks for making the effort! I have one or two naming nits, which we can revisit, and one potential issue with consistency between values in `Environment` and `EnvironmentTargetVersion`.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/model/EnvironmentTargetVersionDDBRecord.java
narehayrapetyan: The advantage is that we don't do two reads. We can always keep extra information here too but I'm open to doing it either way.

data-service/src/main/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandler.java
narehayrapetyan: what's an upsert?

data-service/src/main/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandler.java
narehayrapetyan: yeah you're right. and now that I think about it we don't want them to set it on createEnvironment either: things like type and health etc. I'll change it

data-service/src/main/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandler.java
narehayrapetyan: sure

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
narehayrapetyan: oh yea good catch, forgot to add a use case. That's why integration tests that test every exception are so valuable

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
narehayrapetyan: oh did that in the query and not here? silly

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
narehayrapetyan: so collect will load everything? good to know, we'll have to add an integration test for that

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
narehayrapetyan: oh interesting, the test failed previously, I must've done something else wrong

data-service/src/main/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandler.java
narehayrapetyan: oh this will be a conditional fail for existing environments, will fix

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/model/EnvironmentTargetVersionDDBRecord.java
wjbuys: As mentioned in the other comment, I think if you want to avoid using the (potentially modified) values provided by the user to createEnvironmentTargetVersion, you'd have to do the read to get the values anyway...

data-service/src/main/java/com/amazonaws/blox/dataservice/mapper/EnvironmentMapper.java
```diff
@@ -29,9 +30,9 @@
   @Mapping(source = "cluster", target = "instanceGroup.cluster")
   Environment toEnvironment(EnvironmentDDBRecord environmentDDBRecord);
 
-  @Mapping(source = "instanceGroup.cluster", target = "cluster")
-  EnvironmentTargetVersionDDBRecord toEnvironmentTargetVersionDDBRecord(Environment environment);
+  EnvironmentTargetVersionDDBRecord toEnvironmentTargetVersionDDBRecord(
```
wjbuys: This mapper is throwing a bunch of warnings because of umapped properties:

```
:data-service:compileJava
/home/travis/build/blox/blox/data-service/src/main/java/com/amazonaws/blox/dataservice/mapper/EnvironmentMapper.java:28: warning: Unmapped target properties: "recordVersion, attributes".
  EnvironmentDDBRecord toEnvironmentDDBRecord(Environment environment);
                       ^
/home/travis/build/blox/blox/data-service/src/main/java/com/amazonaws/blox/dataservice/mapper/EnvironmentMapper.java:31: warning: Unmapped target property: "deploymentConfiguration".
  Environment toEnvironment(EnvironmentDDBRecord environmentDDBRecord);
              ^
/home/travis/build/blox/blox/data-service/src/main/java/com/amazonaws/blox/dataservice/mapper/EnvironmentMapper.java:33: warning: Unmapped target property: "recordVersion".
  EnvironmentTargetVersionDDBRecord toEnvironmentTargetVersionDDBRecord(
                                    ^
3 warnings
```

Could we add the annotations to ignore those to get rid of it? If they will soon be mapped, that's also a good spot to add a TODO:

```java
@Mappings({
   @Mapping(source = "recordVersion", ignore = true),
   // TODO: Map attributes
   @Mapping(target = "attributes", ignore = true)
})
```
... etc.

wjbuys (approved):
LGTM! Would be nice to fix the warnings noted below before shipping, though.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/model/EnvironmentTargetVersionDDBRecord.java
narehayrapetyan: you would in the case of create but not describe if we have them in both places

PR #274, opened 2017-10-17
Environment handler and repository unit tests

narehayrapetyan:
#### Summary
unit tests for the previous PR

#### Licensing


data-service/src/main/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandler.java
```diff
@@ -38,10 +39,17 @@
   @NonNull private EnvironmentRepository environmentRepository;
 
   //TODO: createEnvRequest only contains some fields that are set on the env object
-  //The object is precreated in the API layer to avoid passing in and adding/removing parameters in the method definition
-  public Environment createEnvironment(final Environment environment)
+  public Environment createEnvironment(@NonNull final Environment environment)
       throws EnvironmentExistsException, ServiceException {
 
+    Validate.notNull(environment.getEnvironmentId());
```
wjbuys: nit: Can (should?) we validate these with `@NonNull` on the Environment object? That way the client of this code gets an immediate validation error when they create the invalid object. Since we're using `@Builder` anyway, it seems like this should work.

data-service/src/test/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandlerTest.java
```diff
@@ -0,0 +1,338 @@
...
+  @Test(expected = EnvironmentNotFoundException.class)
+  public void createEnvironmentTargetVersionGetReturnsNull()
+      throws EnvironmentNotFoundException, EnvironmentExistsException, ServiceException,
+          StorageException {
+    when(environmentRepository.getEnvironment(ENVIRONMENT_ID, ENVIRONMENT_VERSION))
+        .thenReturn(null);
+    environmentHandler.createEnvironmentTargetVersion(ENVIRONMENT_ID, ENVIRONMENT_VERSION);
+  }
+
+  @Test(expected = ServiceException.class)
```
wjbuys, line 156: nit: For exception handling, have you considered using the [`ExpectedException`](http://junit.org/junit4/javadoc/4.12/org/junit/rules/ExpectedException.html) JUnit Rule instead? I find it makes tests a bit easier to read, because the exception assertion is closer to the code causing the exception, especially in longer test cases like this one.

data-service/src/test/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandlerTest.java
```diff
@@ -0,0 +1,338 @@
...
+  @Test(expected = EnvironmentVersionNotFoundException.class)
+  public void describeEnvironmentTargetVersionGetEnvironmentTargetVersionNull()
+      throws EnvironmentVersionNotFoundException, ServiceException, StorageException {
+    when(environmentRepository.getEnvironmentTargetVersion(ENVIRONMENT_ID)).thenReturn(null);
+    environmentHandler.describeEnvironmentTargetVersion(ENVIRONMENT_ID);
+  }
+
+  @Test
+  public void describeEnvironmentTargetVersion()
+      throws EnvironmentVersionNotFoundException, ServiceException, StorageException {
```
wjbuys: nit: Is there any particular value in including the full checked exception signature on the test method? Usually these are only valuable to callers, and test methods don't have any callers in our code. I've often found it more readable to use `throws Exception` instead (and this is probably the only acceptable use of `throws Exception`).

data-service/src/test/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandlerTest.java
```diff
@@ -0,0 +1,338 @@
...
+        Environment.builder()
+            .environmentId(ENVIRONMENT_ID)
+            .environmentName(ENVIRONMENT_NAME)
+            .taskDefinition(TASK_DEFINITION_ARN)
+            .role(ROLE_ARN)
+            .instanceGroup(InstanceGroup.builder().cluster(CLUSTER_ARN).build())
+            .type(EnvironmentType.Daemon)
+            .build();
+
+    getEnvironment = createEnvironment;
```
wjbuys: Since this assigns a reference, there's only one Environment object here, and modifying `createEnvironment` is also modifying `getEnvironment`. Reusing the `Builder` instance and calling `build()` twice could work.

data-service/src/test/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandlerTest.java
```diff
@@ -0,0 +1,338 @@
...
+        .thenThrow(new StorageException(""));
+    environmentHandler.createEnvironmentTargetVersion(ENVIRONMENT_ID, ENVIRONMENT_VERSION);
+  }
+
+  @Test(expected = EnvironmentNotFoundException.class)
+  public void createEnvironmentTargetVersionGetReturnsNull()
+      throws EnvironmentNotFoundException, EnvironmentExistsException, ServiceException,
+          StorageException {
+    when(environmentRepository.getEnvironment(ENVIRONMENT_ID, ENVIRONMENT_VERSION))
+        .thenReturn(null);
```
wjbuys: Is this how the `getEnvironment` method is defined to behave? I couldn't find a test for `repo.getEnvironment()` that tests the case where the environment is not found.

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
```diff
@@ -198,4 +255,197 @@ public void geEnvironment() throws StorageException, EnvironmentNotFoundExceptio
...
+
+    final Set<String> expectedClusters = new HashSet<>();
+    for (EnvironmentTargetVersionDDBRecord record : environmentTargetVersionDDBRecords) {
+      expectedClusters.add(record.getCluster());
+    }
+
+    assertEquals(expectedClusters.size(), clusters.size());
+    for (String c : expectedClusters) {
+      if (!clusters.contains(c)) {
+        fail(String.format("Does not contain %s", c));
```
wjbuys: We've got Hamcrest, so `assertThat(clusters, containsInAnyOrder(expectedClusters.toArray()))` should work.

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
```diff
@@ -198,4 +255,197 @@ public void geEnvironment() throws StorageException, EnvironmentNotFoundExceptio
...
+
+    when(dynamoDBMapper.scan(
+            eq(EnvironmentTargetVersionDDBRecord.class), isA(DynamoDBScanExpression.class)))
+        .thenReturn(environmentTargetVersionDDBRecordPaginatedScanList);
+
+    final List<String> clusters = environmentRepositoryDDB.listClusters();
+
+    final Set<String> expectedClusters = new HashSet<>();
+    for (EnvironmentTargetVersionDDBRecord record : environmentTargetVersionDDBRecords) {
+      expectedClusters.add(record.getCluster());
```
wjbuys: nit: Is this really necessary? AFAICT `expectedClusters` is a two-element list with `FIRST_CLUSTER, SECOND_CLUSTER` in it...

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
```diff
@@ -198,4 +255,197 @@ public void geEnvironment() throws StorageException, EnvironmentNotFoundExceptio
...
+            .environmentId(SECOND_ENVIRONMENT_ID)
+            .environmentVersion(UUID.randomUUID().toString())
+            .cluster(environmentTargetVersionDDBRecord.getCluster())
+            .build();
+    environmentTargetVersionDDBRecords.add(secondRecord);
+    when(environmentTargetVersionDDBRecordPaginatedQueryList.stream())
+        .thenReturn(environmentTargetVersionDDBRecords.stream());
+
+    when(dynamoDBMapper.query(
+            eq(EnvironmentTargetVersionDDBRecord.class), isA(DynamoDBQueryExpression.class)))
```
wjbuys: The `isA(DynamoDBQueryExpression.class)` causes an `unchecked assignment` warning here because of the raw type. Since we're already testing the actual value captured in a test case below, perhaps just using `any()` here is sufficient?

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
```diff
@@ -133,34 +192,32 @@ public void createEnvironment() throws StorageException, EnvironmentExistsExcept
 
     verify(dynamoDBMapper).save(environmentDDBRecordCaptor.capture());
     assertEquals(
-        environmentDDBRecord.getAttributes(),
+        environment.getInstanceGroup().getAttributes(),
```
wjbuys: Does this work for non-empty attributes? AFAICT, `InstanceGroup::getAttributes` returns a `Set<Attribute>`, but `EnvironmentDDBRecord::getAttributes` returns a `Set<String>`.

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
```diff
@@ -198,4 +255,197 @@ public void geEnvironment() throws StorageException, EnvironmentNotFoundExceptio
...
+            .environmentVersion(UUID.randomUUID().toString())
+            .cluster(SECOND_CLUSTER)
+            .build();
+    environmentTargetVersionDDBRecords.add(thirdRecordSameCluster);
+
+    when(environmentTargetVersionDDBRecordPaginatedScanList.stream())
+        .thenReturn(environmentTargetVersionDDBRecords.stream());
+
+    when(dynamoDBMapper.scan(
+            eq(EnvironmentTargetVersionDDBRecord.class), isA(DynamoDBScanExpression.class)))
```
wjbuys, line 351: The `isA(DynamoDBScanExpression.class)` is causing an `unchecked assignment` warning because it's referring to the raw `DynamoDBScanExpression<>` type.

Since we're testing the actual value of this argument with a captor later, would it be sufficient to use `any()` or `notNull()` instead? In general though, since Java's doing the type checking here, I'm not sure using the `isA(class)` matcher is adding value unless we care about the value being a specific subtype.

(This happens a couple of times in this file)

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
```diff
@@ -198,4 +255,197 @@ public void geEnvironment() throws StorageException, EnvironmentNotFoundExceptio
...
+            new DynamoDBQueryExpression<EnvironmentTargetVersionDDBRecord>()
+                .withIndexName(EnvironmentTargetVersionDDBRecord.ENVIRONMENT_CLUSTER_GSI_NAME)
+                .withConsistentRead(false)
+                .withHashKeyValues(
+                    EnvironmentTargetVersionDDBRecord.withGSIHashKey(
+                        environmentTargetVersionDDBRecord.getCluster()));
+
+    assertEquals(
+        expectedDynamodbQueryExpression.getIndexName(),
+        dynamoDBQueryExpressionCaptor.getValue().getIndexName());
```
wjbuys, line 431: If we're not comparing the two expression objects directly, why not use the values directly instead of constructing the intermediate expression?

wjbuys (changes requested):
Implementation looks good, but there's a couple of issues with the unit tests.

data-service/src/main/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandler.java
narehayrapetyan: Yea, good point. Previously I had one object for environment and environmentVersion so couldn't set these fields to nonnull on the object and forgot to change it after the separation.

data-service/src/test/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandlerTest.java
narehayrapetyan: works for me

data-service/src/test/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandlerTest.java
narehayrapetyan: whoops, good catch

data-service/src/test/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandlerTest.java
narehayrapetyan: DynamodbMapper.load returns null if the record is not found

data-service/src/test/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandlerTest.java
narehayrapetyan: sure, I can start using it in the future unit tests

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
narehayrapetyan: how does java surface the error?

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
narehayrapetyan: Yes there's a TODO in the code to add and map attributes

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
narehayrapetyan: This is to generalize it and not depend on previous knowledge of arguments. I can use the hamcrest method you recommend below

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
narehayrapetyan: how do you use the values directly? if i just use the expected in verify it compares the objects and throws something to the effect of DynamoDBScanExpression@2c88b9fc != DynamoDBScanExpression@3918c187

data-service/src/test/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandlerTest.java
wjbuys: 👍

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
wjbuys: I just ask, because this is also throwing an `unchecked assignment` warning in the build output.

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
wjbuys: Yeah, what I mean is, why even construct `expectedDynamoDbQueryExpression` in the previous line if you're not going to use it? Rather use the value directly.

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
wjbuys: In the [build output](https://travis-ci.org/blox/blox/builds/289718504#L581); before we had no warnings, now we have:

```
Note: /home/travis/build/blox/blox/data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
```

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
narehayrapetyan: Sure that would work too but this works as well and I don't think it warrants another change.

wjbuys (approved):
Good to go once the compiler warnings are fixed.

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
wjbuys: Thinking about this, we shouldn't have to go dig in build output for issues like these in PRs. Any objection to adding `-Wlint:all -Werror` to our `javac` compiler flags so that these issues actually become visible as build errors?

PR #275, opened 2017-10-23
Implement efficient cluster state snapshots in ECSStateClient

wjbuys:
This uses an `ECSAsyncClient` to implement the `ECSState` interface, and updates the `SingleTaskScheduler` to actually only run a single task.

This also changes the types in the snapshot to be lists of instances/tasks, since the maps weren't very useful in practice.

The integration test is very basic, but we may be able to build on it for further integration tests that require a test cluster.

Tracking issue: #261

#### Licensing


wjbuys:
Looks like there's another ECS client getting started in the Spring tests, modified the build logic to show more failure details in Travis to see what's going on.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/state/FakeECSStateClient.java
```diff
@@ -0,0 +1,32 @@
...
+package com.amazonaws.blox.scheduling.state;
+
+import java.util.Collections;
+import lombok.SneakyThrows;
+
+/**
+ * Temporary fake implementation of ECS State client that just takes time and returns an empty
+ * cluster.
+ */
+public class FakeECSStateClient implements ECSState {
```
wjbuys: That's weird, I thought I deleted this file. I'll remove it.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/manager/ManagerApplication.java
```diff
@@ -14,33 +14,34 @@
...
 import org.springframework.context.annotation.Configuration;
+import software.amazon.awssdk.services.lambda.LambdaAsyncClient;
 
 @Configuration
-@ComponentScan("com.amazonaws.blox.scheduling.manager")
+@ComponentScan({
+  "com.amazonaws.blox.scheduling.manager",
+  "com.amazonaws.blox.scheduling.state",
+})
 public class ManagerApplication extends SchedulingApplication {
```
narehayrapetyan, line 34: would it be possible to put all the configuration in one package so it's easy to find?

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/SingleTaskScheduler.java
```diff
@@ -19,22 +19,47 @@
...
 /**
  * Temporary scheduler implementation that just starts a single task on the first instance in the
  * snapshot.
+ *
+ * <p>TODO Remove this, it's only a demo scheduler
  */
 public class SingleTaskScheduler implements Scheduler {
 
   @Override
   public List<SchedulingAction> schedule(ClusterSnapshot snapshot, EnvironmentVersion environment) {
```
narehayrapetyan: you'll have to create a new internal object for environmentversion or something similar because the returned type is now describeEnvironment which contains separate fields. can be done in the refactoring cr

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/SingleTaskScheduler.java
```diff
@@ -19,22 +19,47 @@
...
-    // Only run a Task on the first ContainerInstance in the snapshot:
-    for (Entry<String, ContainerInstance> entry : snapshot.getInstances().entrySet()) {
+    // Naively only run a Task on the first ContainerInstance in the snapshot:
+    ContainerInstance instance =
+        snapshot
+            .getInstances()
+            .stream()
+            .sorted(Comparator.comparing(ContainerInstance::getArn))
```
narehayrapetyan, line 40: do we need to sort? I suppose this is gonna be removed anyway

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/StartTask.java
```diff
@@ -15,29 +15,37 @@
...
   public CompletableFuture<Boolean> execute(ECSAsyncClient ecs) {
-    // TODO: This will probably require setting the task group to the environment ID too
     CompletableFuture<StartTaskResponse> pendingRequest =
         ecs.startTask(
             StartTaskRequest.builder()
                 .cluster(clusterArn)
                 .containerInstances(containerInstanceArn)
                 .taskDefinition(taskDefinitionArn)
+                .group(group)
+                .startedBy("blox")
```
narehayrapetyan, line 43: constant

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/state/PaginatedResponseIterator.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.stream.StreamSupport;
+import lombok.RequiredArgsConstructor;
+
+/**
+ * Abstract iterator that exposes a paginated list API call as an Iterator/Stream.
+ *
+ * @param <T> The response type returned by the list API call
+ */
+@RequiredArgsConstructor
+class PaginatedResponseIterator<T> implements Iterator<T> {
```
narehayrapetyan, line 32: why are these classes not public?

scheduling-manager/src/test/java/com/amazonaws/blox/scheduling/state/ContainerInstanceListerTest.java
```diff
@@ -0,0 +1,112 @@
...
+    ContainerInstanceLister containerInstances = new ContainerInstanceLister(ecs, CLUSTER_1);
+
+    List<ClusterSnapshot.ContainerInstance> describe = containerInstances.describe().join();
+
+    assertThat(describe, empty());
+  }
+
+  @Test
+  public void returnsContainerInstancesFromAllPages() {
+    ECSAsyncClient ecs = mock(FakeECSAsyncClient.class, Mockito.CALLS_REAL_METHODS);
```
narehayrapetyan, line 59: is this one of the temporary unit/integ mix tests? can we move them to a separate package so they're easy to refactor into integ tests later?

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/manager/ManagerApplication.java
wjbuys: In this case, it's not the configuration that's in a different package, it's the `@Component` annotation on the state stuff. I'm open to any suggestions on how to deal with that.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/SingleTaskScheduler.java
wjbuys: That's what I did for #278

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/SingleTaskScheduler.java
wjbuys: yes. Just so we don't start the task multiple times on different instances just because the ordering is different. (I had that before, and ended up with a cluster full of random tasks)

scheduling-manager/src/test/java/com/amazonaws/blox/scheduling/state/ContainerInstanceListerTest.java
wjbuys: No, this unit test has value, and will need to be expanded to cover failure cases that are hard/impossible to reproduce against a real ECS client (e.g. partial asynchronous failure)

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/state/PaginatedResponseIterator.java
wjbuys: Think it was left over from extracting this as a private inner class on one of the listers. Should it be public? I do not have strong feelings about java class visibility beyond what's needed to make the compiler happy, so I'll defer to whatever your guideline is here.

wjbuys:
Just to follow up, I extracted some of this PR into #278 to fix the build, so I'll have to rebase this PR after that one is merged to `dev`.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/StartTask.java
wjbuys: It's a single usage of an unimportant string in a temporary class. I'll extract it, but I'm not convinced it's helpful.

wjbuys:
I've rebased this PR onto dev, should be ready to review/ship now.

scheduling-manager/src/test/java/com/amazonaws/blox/scheduling/state/ContainerInstanceListerTest.java
wjbuys: Oh, and to answer the second part of your question: They are pure unit tests, since they mock out all dependencies, do I don't think they belong in a separate package.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/manager/ManagerApplication.java
narehayrapetyan: What do you mean? I think the schedulingapplication configuration is not in com...manager

narehayrapetyan (approved):

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/manager/ManagerApplication.java
wjbuys: Oh, wait, dammit github highlighting. Do you mean "why is SchedulingApplication in a different package"? If so, the reason is that it's the shared config between all components of the scheduling-manager package. If I were to put it in any of the subpackages, then it would no longer be clear that it's common to all of them.

An alternative could be to add a separate top-level `com.amazonaws.blox.scheduling.config` package, and add all the config in there. Is that what you were alluding to?

PR #276, opened 2017-10-23
Adding data service api layer

narehayrapetyan:
Adding data service api layer


wjbuys:
This one should be rebased against current `dev` branch before review, right?

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/wrappers/DescribeEnvironmentResponse.java
```diff
@@ -41,8 +39,8 @@
   @NonNull private InstanceGroup instanceGroup;
   @NonNull private EnvironmentType type;
   @NonNull private String environmentVersion;
-  @NonNull private EnvironmentStatus status;
-  @NonNull private EnvironmentHealth health;
+  @NonNull private String status;
```
wjbuys, line 42: Why the change from enum to string?

wjbuys (approved):

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/wrappers/DescribeEnvironmentResponse.java
narehayrapetyan: Because when you have enums on the server and client side making changes becomes very difficult as it's hard to guarantee that everyone has the latest changes at the same time. Constants allow us to have access to allowed values while avoiding the deserialization problem with enums.

PR #277, opened 2017-10-24
Adding data service interface and wrapper objects

narehayrapetyan:
As agreed, pulling out the dataservice interface and contract from the previous PR to make the refactoring of the scheduler easier.


wjbuys (approved):
I'll do the followup CR to fix the scheduler side of things ASAP.

Issue #227, opened 2017-08-09
Wire up the Scheduling Manager to the DataService


wjbuys:
We want to host the DataService as a single Lambda function, but we don't want to front it with AWS API Gateway (since we're not exposing the service as a public API).

This means that we'll have to figure out some way to route calls for different codepaths in the same Lambda function based on the payload of the invocation.

For example, from the FrontendService lambda function, we want to be able to call:

```
DataService dataService = createClient();
dataService.createEnvironment(new Environment().withName("foo"));
Environment env = dataService.describeEnvironment("foo");
```

Which should invoke methods on the DataService lambda function:

```
public class DataServiceActivity implements DataService {
  public void createEnvironment(Environment env) { ... }
  public Environment describeEnvironment(String name) { ... }
}
```

Somewhere in between, we have to convert the calls to something that the Lambda wire format can handle.

On the client side, this means that the request has to be converted into a [ByteBuffer/String on `InvokeRequest`](http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/lambda/model/InvokeRequest.html#setPayload-java.nio.ByteBuffer-), and the response converted from a [ByteBuffer/String on `InvokeResult`](http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/lambda/model/InvokeResult.html#getPayload--)

On the server side, we can use [`RequestStreamHandler`](http://docs.aws.amazon.com/lambda/latest/dg/java-handler-io-type-stream.html) which gives us an `InputStream`/`OutputStream` to read/write the payload.

wjbuys:
One simple option that comes to mind is to use the [JSON-RPC](https://en.wikipedia.org/wiki/JSON-RPC) protocol. It's pretty simple, and there's an open source implementation that supports tying the service to an `InputStream`/`OutputStream` already: [`jsonrpc4j`](https://github.com/briandilley/jsonrpc4j)

wjbuys:
Alternatively, we can bite the bullet and model this API as a REST API in API Gateway too, and use the same complicated Spring MVC -> Swagger Model -> Generated Client toolchain.

wjbuys:
I think as long as we have the clean interface-based implementation model as it's outlined in #226, it doesn't really matter what we choose for the transport, since it should be pretty easy to change. I'm going to go ahead an prototype something with jsonrpc4j, and if it works I'll integrate it into the project.

wjbuys:
[...]

PR #281, opened 2017-10-26
Wire up Data Service to Scheduling Manager using JSON-RPC

wjbuys:
This deploys the DataService as a lambda function that accepts JSON-RPC payloads, and adds a JSON-RPC client to the Scheduling Manager in order to call the DataService.

Right now, this wires up the DataService's Lambda function into the scheduler using Cloudformation's ImportValue function. While this is nice for development, we'll probably want to figure out a better long term way to do this.

Implements #227

#### Testing
So far, testing has been manual. After running `gradle deploy` to deploy the full stack, you have to manually create an Environment that targets a cluster that your account owns, and then wait for a minutely reconciliation run.

- [ ] TODO: Add end-to-end test that does this

#### Licensing


wjbuys:
I've been trying to set up the end-to-end test for this, and I'm running into issues re-running the tests. In particular, if we don't clear out the data in DDB, then we'll keep scheduling more and more tasks for each new test environment created. I think it might be generally useful to have an easy way to clear out all the state from the test stack, so I'm going to add a utility for that to our end-to-end-tests package.

narehayrapetyan:
Yes, I agree, the tests should clean up after themselves.

json-rpc-lambda-client/build.gradle
```diff
@@ -0,0 +1,18 @@
...
+}
+
+dependencies {
+    compile "com.github.briandilley.jsonrpc4j:jsonrpc4j:+"
+
+    // TODO: Move dependencyManagement block to the root project so that
+    // we don't have to depend on a specific SDK version here:
+    compile "software.amazon.awssdk:lambda:+"
+
+    // For @Slf4j annotation
```
narehayrapetyan: why only for slf4j? won't this comment become out of date when we use lombok for other stuff?

json-rpc-lambda-client/src/main/java/com/amazonaws/blox/clients/JsonRpcLambdaClient.java
```diff
@@ -0,0 +1,193 @@
...
+
+  private final LambdaAsyncClient lambda;
+  private final String functionName;
+  private final JsonRpcClient client;
+
+  private JsonRpcLambdaClient(LambdaAsyncClient lambda, String functionName) {
+    this(defaultObjectMapper(), lambda, functionName);
+  }
+
+  private JsonRpcLambdaClient(ObjectMapper mapper, LambdaAsyncClient lambda, String functionName) {
```
narehayrapetyan: what's the advantage of having two constructors? won't it be more concise to autowire the dependencies?

json-rpc-lambda-client/src/main/java/com/amazonaws/blox/clients/JsonRpcLambdaClient.java
```diff
@@ -0,0 +1,193 @@
...
+   *
+   * @param mapper the Jackson mapper to use for serializing/deserializing requests/responses
+   * @param serviceInterface the interface of the service to call
+   * @param lambdaClient a lambda client that can call the function that's running the service
+   *     implementation
+   * @param functionName the name of the labmda function that's running the service implementation
+   * @param <T>
+   * @return
+   */
+  public static <T> T newProxy(
```
narehayrapetyan: why not move this outside of this class so we don't have a mix of instantiated and static methods? it looks like some methods are calling the same methods on the client as well, it would help with readability imho to separate these out

json-rpc-lambda-client/src/main/java/com/amazonaws/blox/clients/JsonRpcLambdaClient.java
narehayrapetyan: why have 2 different constructors? and why not autowire?

json-rpc-lambda-client/src/main/java/com/amazonaws/blox/clients/JsonRpcLambdaClient.java
```diff
@@ -0,0 +1,193 @@
...
+      ObjectMapper mapper,
+      Class<T> serviceInterface,
+      LambdaAsyncClient lambdaClient,
+      String functionName) {
+    IJsonRpcClient client = new JsonRpcLambdaClient(mapper, lambdaClient, functionName);
+    return ProxyUtil.createClientProxy(
```
narehayrapetyan: from the documentation it looks like this can be autowired?

json-rpc-lambda-client/src/main/java/com/amazonaws/blox/clients/JsonRpcLambdaClient.java
```diff
@@ -0,0 +1,193 @@
...
+   * @param <T>
+   * @return
+   */
+  public static <T> T newProxy(
+      Class<T> serviceInterface, LambdaAsyncClient lambdaClient, String functionName) {
+    return newProxy(defaultObjectMapper(), serviceInterface, lambdaClient, functionName);
+  }
+
+  /** {@inheritDoc} */
+  @Override
```
narehayrapetyan: what's this overriding? (and the rest) we're just implementing an interface no?

narehayrapetyan (commented):
So are the scheduling functions using this already?

json-rpc-lambda-client/src/main/java/com/amazonaws/blox/clients/JsonRpcLambdaClient.java
wjbuys: This was a refactoring step from when the constructor didn't have an ObjectMapper parameter. Will remove the redundant constructor.

json-rpc-lambda-client/src/main/java/com/amazonaws/blox/clients/JsonRpcLambdaClient.java
narehayrapetyan: looks like renaming the client to jsonRPCclient or similar would make it clearer

json-rpc-lambda-client/build.gradle
wjbuys: Sure.

json-rpc-lambda-client/src/main/java/com/amazonaws/blox/clients/JsonRpcLambdaClient.java
wjbuys: I was trying to contain the this implementation detail to as small a surface as possible. After our discussion, I think splitting this off into a separate class, and making the JsonRpcLambdaClient package-local will achieve that more cleanly.

json-rpc-lambda-client/src/main/java/com/amazonaws/blox/clients/JsonRpcLambdaClient.java
wjbuys: As we discussed, I'm trying to localize all of the unpleasant glue for this in one place, so that we don't have multiple classes that refer to the JSON-RPC implementation details spread through our application. I'll add the explanatory javadoc for that.

json-rpc-lambda-client/src/main/java/com/amazonaws/blox/clients/JsonRpcLambdaClient.java
wjbuys: As discussed, these are unfortunately required generic wrappers from the `IJsonRpcClient`interface for the corresponding methods that return `Object`. I'll replace the inheritdoc comment with an explanation.

wjbuys:
> So are the scheduling functions using this already?

Yes, [this](https://github.com/blox/blox/pull/281/files#diff-1d1f6dc21c6be2d5d91e3a504dbf93a0R79) is what wires it up.

narehayrapetyan (approved):

PR #282, opened 2017-11-01
Add an end-to-end test for a SingleTask Environment

wjbuys:
#### Summary
This adds a single JUnit end-to-end test that calls the DataService to create an environment and set it active, and verifies that it actually launches a task.

The TestStack and ServiceWrappers should be replaced with dependency-injected versions once we get Cucumber in place.

Implements #227

#### Testing
Deployed a new personal stack and tested it with:
```
./gradlew testEndToEnd
```

#### Licensing


end-to-end-tests/src/main/java/com/amazonaws/blox/integ/DataServiceSupport.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.amazonaws.blox.dataservicemodel.v1.client.DataService;
+import com.amazonaws.blox.dataservicemodel.v1.serialization.DataServiceMapperFactory;
+import com.amazonaws.blox.integ.CloudFormationStacks.CfnStack;
+import com.amazonaws.blox.jsonrpc.JsonRpcLambdaClient;
+import lombok.RequiredArgsConstructor;
+import software.amazon.awssdk.services.lambda.LambdaAsyncClient;
+
+@RequiredArgsConstructor
+public class DataServiceSupport {
```
narehayrapetyan: instead of having a class for each service should we have something that just sets up the entire service stack?

end-to-end-tests/src/main/java/com/amazonaws/blox/integ/DynamoDBSupport.java
```diff
@@ -0,0 +1,63 @@
...
+import software.amazon.awssdk.services.dynamodb.model.ScanRequest;
+import software.amazon.awssdk.services.dynamodb.model.ScanResponse;
+
+@RequiredArgsConstructor
+public class DynamoDBSupport {
+
+  private final DynamoDBClient ddb;
+
+  void truncateTable(String tableName) {
+    // This will only delete one page of records. Fine for now, but we should probably just use
```
narehayrapetyan: why are you not using the mapper now?

end-to-end-tests/src/main/java/com/amazonaws/blox/integ/PersonalDevelopmentStack.java
```diff
@@ -0,0 +1,86 @@
...
+import software.amazon.awssdk.services.ecs.model.DescribeTasksRequest;
+import software.amazon.awssdk.services.ecs.model.DesiredStatus;
+import software.amazon.awssdk.services.ecs.model.ListTasksRequest;
+import software.amazon.awssdk.services.ecs.model.StopTaskRequest;
+import software.amazon.awssdk.services.ecs.model.Task;
+import software.amazon.awssdk.services.lambda.LambdaAsyncClient;
+
+/** Facade over everything in an end-to-end Blox stack that's needed in tests. */
+public class PersonalDevelopmentStack implements TestStack {
+  private CloudFormationClient cloudformation = CloudFormationClient.create();
```
narehayrapetyan: why not inject all the clients using a spring config file?

end-to-end-tests/src/main/java/com/amazonaws/blox/integ/PersonalDevelopmentStack.java
```diff
@@ -0,0 +1,86 @@
...
+  public List<Task> describeBloxTasks() {
+    List<String> taskArns = listBloxTasks();
+    if (taskArns.isEmpty()) return Collections.emptyList();
+
+    return ecs.describeTasks(
+            DescribeTasksRequest.builder().cluster(cluster).tasks(taskArns).build())
+        .tasks();
+  }
+
+  private List<String> listBloxTasks() {
```
narehayrapetyan: the blox tasks shouldn't be in the dev stack right? this should just set everything up and the tests implement the verificaiton? so there's separation between just stack setup and tests.

narehayrapetyan (commented):
Not super clear on how this is organized, what do you think of this?

We have a section of stack setup outside of tests since this will be useful even without running tests. This can have a configuration file with all the AWS clients (for dependency injection) and a CF section that takes a list of blox service templates to set up. You have most of this just organized in a different way I think and it all lives in the test package.

end-to-end-tests/src/main/java/com/amazonaws/blox/integ/DynamoDBSupport.java
wjbuys: Dependency problems. In order to access the data service repository models, I'd have to add a test dependency on the `data-service` top-level project. This pulls in a bunch of dependencies that weren't playing nice when combined with the other dependencies here, because of the spring-dependency-management plugin. I think we can revisit it once we've done the build logic cleanup to consolidate all of those; will update the comment to reflect that.

end-to-end-tests/src/main/java/com/amazonaws/blox/integ/PersonalDevelopmentStack.java
wjbuys: Yeah, this really should be pulled out into an ECS wrapper class instead of being stuck in here. Will do.

end-to-end-tests/src/main/java/com/amazonaws/blox/integ/DataServiceSupport.java
wjbuys: As mentioned, this doesn't do the stack setup, it's just pulling values out of cloudformation.

end-to-end-tests/src/main/java/com/amazonaws/blox/integ/PersonalDevelopmentStack.java
wjbuys: Let's come back to Spring later, there's not many dependencies right now, so it should be easy to wire things up with Spring once this gets any larger.

end-to-end-tests/src/main/java/com/amazonaws/blox/integ/ECSClusterWrapper.java
```diff
@@ -0,0 +1,76 @@
...
+
+  public String getTaskDefinition() {
+    return stack.output("taskdef");
+  }
+
+  public String getCluster() {
+    return stack.output("cluster");
+  }
+
+  public List<Task> describeTasks() {
```
narehayrapetyan, line 50: should we pass in cluster and task here so the wrapper is agnostic to where who is calling and what for? (ie not have any knowledge of stacks or inputs)

narehayrapetyan (approved):

end-to-end-tests/src/main/java/com/amazonaws/blox/integ/ECSClusterWrapper.java
wjbuys: I'm not sure at this point; let's see where it goes after we do the spring/cucumber refactoring.

PR #283, opened 2017-11-01
Add cucumber build logic and sample features

narehayrapetyan:
Cucumber gradle build logic with spring integration

Should we exclude cucumber steps from spotlessapply?


data-service/src/cucumberTest/java/cucumber/steps/CreateEnvironmentSteps.java
```diff
@@ -0,0 +1,37 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package cucumber.steps;
+
+import cucumber.api.java8.En;
+import org.apache.commons.lang3.NotImplementedException;
+import org.springframework.test.context.ContextConfiguration;
+
+@ContextConfiguration(locations = "classpath:cucumber.xml")
```
wjbuys: Is it necessary to fall back on XML config here? It looks like this should work just fine with `@ContextConfiguration(classes=com.amazonaws.blox.dataservice.Application.class)` instead.

data-service/build.gradle
```diff
@@ -38,6 +53,37 @@ compileJava {
     ]
 }
 
+sourceSets {
+    integTest {
+        java.srcDir file('src/integTest/java')
```
wjbuys, line 58: Would we have integration tests that are not Cucumber features? If not, I don't think another `integTest` SourceSet is actually necessary, since the Cucumber plugin [already sets up the `cucumberTest` SourceSet for us.](https://github.com/samueltbrown/gradle-cucumber-plugin/blob/master/src/main/groovy/com/excella/gradle/cucumber/CucumberPlugin.groovy#L31).

If we do need a separate `integTest` SourceSet (i.e. if we want to run integration tests that need a separate dependency closure from both unit tests and cucumber tests), then it looks like the [`gradle-testsets` plugin](https://github.com/unbroken-dome/gradle-testsets-plugin) would turn all this boilerplate into a single config stanza.

wjbuys (commented):
Neat. It looks like we could avoid most of the common setup code here with the comments below.

data-service/build.gradle
narehayrapetyan: This was just moved up from below, it's not new and will be removed once the integ tests are moved into cucumber.

data-service/src/cucumberTest/java/cucumber/steps/CreateEnvironmentSteps.java
narehayrapetyan: we need separate configuration for integ tests but I can move it into java config

wjbuys (approved):
LGTM!

PR #286, opened 2017-11-06
Moving cucumber tests to integ-tests and moving dataservice lambda cl…

narehayrapetyan:
Moving the data service client config into the client package so it's easy to import
Moving the tests into a common package for wrapper and step reuse


wjbuys:
Could you elaborate a bit on how you think we should organize integration tests? Will all the integration tests for all subsystems live in the `integ-tests` project? Is having one project for all the integ tests going to work in terms of dependencies, and being able to run the integ tests for just one subsystem, for example?

narehayrapetyan:
We can tag tests for different services with their own tags such as @dataservice and run only dataservice tests using the tag. Having all the tests in one place allows us to reuse cucumber setup and wrappers but we could also have them in separate packages with minimal repetition.

data-service-client/src/main/java/com/amazonaws/blox/dataserviceclient/v1/client/DataServiceLambdaClientConfiguration.java
```diff
@@ -0,0 +1,45 @@
...
+import com.amazonaws.blox.dataservicemodel.v1.client.DataService;
+import com.amazonaws.blox.dataservicemodel.v1.serialization.DataServiceMapperFactory;
+import com.amazonaws.blox.jsonrpc.JsonRpcLambdaClient;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import software.amazon.awssdk.services.lambda.LambdaAsyncClient;
+
+/** AWS Lambda client for the DataService using JSONRPC for routing. */
+@Configuration
+public class DataServiceLambdaClientConfiguration {
```
wjbuys: Is this intended only for integration tests? If we model it as a Spring configuration class, I don't think the `@Import` for this will work correctly in e.g. the scheduling manager, since it also defines a lambda client bean, and it gets its function name from an environment variable rather than from CloudFormation. Perhaps this can be only a factory method instead?

integ-tests/src/cucumber/java/configuration/CucumberConfiguration.java
```diff
@@ -0,0 +1,36 @@
...
+@Configuration
+@Import(DataServiceLambdaClientConfiguration.class)
+@ComponentScan("steps")
+public class CucumberConfiguration {
+
+  @Autowired private DataServiceLambdaClientConfiguration dataServiceLambdaClient;
+
+  @Bean
+  public DataServiceWrapper dataServiceWrapper() {
+    return new DataServiceWrapper(dataServiceLambdaClient.dataService());
```
wjbuys: You could avoid the explicit autowire by declaring a `DataService` parameter for the `dataServiceWrapper` bean definition method.

wjbuys (approved):

integ-tests/src/cucumber/java/configuration/CucumberConfiguration.java
narehayrapetyan: you mean dataservicewrapper(Dataservice dataservice) will autowire the bean?

wjbuys (approved):

PR #287, opened 2017-11-06
Clean up Gradle build scripts, move config to root project

wjbuys:
This moves as much of our repeated subproject configuration as makes sense to the root project, and versions more of the most common dependencies using `dependencyManagement` in the root project.

This also consolidates all plugin versions in the root project using the `plugins {}` block, and only applies them in subprojects that need them. In order to make use of this new syntax in more places, this also upgrades the Gradle version to 4.3 ([release notes](https://docs.gradle.org/4.3/release-notes.html#more-use-cases-supported-using-the-plugins-%7B%7D-block))

#### Licensing


build.gradle
```diff
@@ -1,29 +1,65 @@
...
+}
+
+subprojects {
+    apply plugin: 'io.spring.dependency-management'
 
-    dependencies {
-        classpath 'com.diffplug.gradle.spotless:spotless:2.4.1'
+    dependencyManagement {
+        imports {
+            mavenBom 'com.amazonaws:aws-java-sdk-bom:1.11.215'
```
narehayrapetyan, line 29: do we need both sdk version or is this so we don't have to refactor code that doesn't use v2 now?

buildSrc/build.gradle
```diff
@@ -20,6 +22,7 @@ dependencies {
         exclude group: 'com.github.jknack', module: 'handlebars'
     }
 
+    // We can't rely on the dependency-managed version of lombok here, so we have to hardcode it:
```
narehayrapetyan: why not?

data-service/build.gradle
```diff
@@ -71,7 +44,7 @@ dependencies {
             'org.apache.logging.log4j:log4j-slf4j-impl:2.8.+',
```
narehayrapetyan, line 44: can we move logging to dep management too? since there's no reason to use different versions across projects

narehayrapetyan (approved):

build.gradle
wjbuys: Yes, this is so that we don't have to refactor the parts of the dataservice that still depend on V1 now.

buildSrc/build.gradle
wjbuys: The `buildSrc` special project is built before the main project, and so it can't get any configuration from the main project. I'll update the comment to make that clearer.

data-service/build.gradle
wjbuys: Agreed; I'll do that as part of the bigger logging cleanup to move to slf4j everywhere.

PR #288, opened 2017-11-08
Adding dataservice integ test feature definitions

narehayrapetyan:
Refactoring createEnvironmentTargetRevision to throw EnvironmentTargetRevisionExists instead of EnvironmentExists

./gradlew build and cucumber


integ-tests/src/cucumber/resources/features/dataservice/create_environment.feature
```diff
@@ -1,7 +1,15 @@
...
 Feature: Create environment
 
   Scenario: Create an environment
     When I create an environment
     Then the created environment response is valid
+
+  Scenario: Create an environment that already exists
+    Given I create an environment
+    When I try to create another environment with the same name
+    Then there should be an EnvironmentExistsException thrown
```
wjbuys: This specification looks correct to me, and is what I would expect from the createEnvironment API call. However, as far as I can tell from the code, the `Environment` resource in the data service created by `createEnvironment` is the one that *can* have duplicate records for the same environment name, as long as they have different `environmentVersion` fields.

integ-tests/src/cucumber/resources/features/dataservice/create_environment_target_revision.feature
```diff
@@ -0,0 +1,21 @@
...
+
+  Scenario: Create an environment target revision for a non-existent environment
+    When I create an environment target revision
+    Then there should be an EnvironmentNotFoundException thrown
+
+  Scenario: Create an environment target revision that already exists
+    Given I create an environment
+    And I create an environment target revision
+    When I try to create another environment target revision with the same id and version
+    Then there should be an EnvironmentTargetRevisionExistsException thrown
```
wjbuys: This specification looks correct to me, and is what I would expect from an API call that creates a new revision of an environment. However, as far as I can tell from the code, the `EnvironmentTargetRevision` resource in the data service created by `createEnvironmentTargetRevision` is the one that *can not* have duplicate records for the same environment name, even if the `environmentVersion` is different.

wjbuys (approved):
The feature descriptions match what I'd expect from the APIs, but they don't seem to match the implementations of the current `createEnvironment` and `createEnvironmentTargetRevision` APIs. In particular, from the features, it looks like the `Create Environment Target Revision` spec is describing the current behaviour of the `createEnvironment` API call, and vice versa.

I think, if we scrap the "Target" from "Environment Target Revision", and then make the code match these features, they're pretty much spot on, and will match the frontend APIs nicely.

integ-tests/src/cucumber/resources/features/dataservice/create_environment.feature
narehayrapetyan: The tests shouldn't match the code, the code should match what we think should be happening which is defined here :) What we need to add here is that the name cannot be duplicated within a cluster but can be duplicated across different clusters, I'll add a use case. And after the refactor of the environment and environmentRevision models what's described should happen in the APIs.

wjbuys: > The tests shouldn't match the code, the code should match what we think should be happening which is defined here :)

100x :+1:

PR #291, opened 2017-11-14
Environment and EnvironmentRevision data model changes

narehayrapetyan:
Breaking changes so I won't push this by itself but want to double check we're on the same page with regards to the data model.


data-service/src/main/java/com/amazonaws/blox/dataservice/handler/EnvironmentHandler.java
```diff
@@ -87,10 +87,10 @@ public EnvironmentVersion createEnvironmentTargetVersion(
     }
   }
 
-  public EnvironmentVersion describeEnvironmentTargetVersion(@NonNull final String environmentId)
+  public EnvironmentRevision describeEnvironmentTargetVersion(@NonNull final String environmentId)
```
wjbuys, line 90: This signature will become `describeEnvironmentRevision(environmentId, environmentRevisionId)`, then?

data-service/src/main/java/com/amazonaws/blox/dataservice/model/Environment.java
```diff
@@ -31,16 +31,14 @@
 
   @NonNull private String environmentId;
```
wjbuys, line 32: Since this is part of the API, and not the data model, could we expose this as `(accountId, environmentName)` instead? Otherwise we have to spread ID construction/deconstruction logic around the frontend too.

wjbuys (commented):
API model changes look good, barring the EnvironmentId issue

Issue #280, opened 2017-10-25
Wire up data service lambda function to frontend


wjbuys:
After we have some way of calling the DataService (see #227), we should wire this up to the Frontend.

Each data service client needs the following:
- the name of the data service's Lambda function
- permissions to invoke that Lambda function
- a correctly configured JsonRpcLambdaClient that can serialize/deserialize all dataservice types.
 
### Tasks
- [x] add createEnvironment, describeEnvironment, startEnvironmentDeployment APIs
- [x] wire up DataService through spring/cloudformation
- [x] modify end-to-end tests to call frontend
- [ ] add error mapping/logging to frontend

wjbuys:
I'm having some trouble expressing our API in swagger. If I try to structure things for the AWS IAM auth to just work, then the generated client is confusing.

The two constraints that are in tension are:
- Encoding resource names in the URL. This is needed for API Gateway's automatic authorization support. In particular, any attributes we wish to authorize on have to be part of the URL, so that it can be converted to an ARN internally.
- Providing resource properties as a JSON POST payload.

So, for example for `updateEnvironment`, the URL must contain the attributes we wish to support resource-based authorization on, in this case `/environment/{environmentName}`. However, if the swagger definition contains both a PathVariable (`environmentName`) and a request body (`UpdateEnvironmentRequest`):

```
paths:
  /v1/{cluster}/environments/{environmentName}:
    put:
      summary: "Update environment"
      description: ""
      operationId: "updateEnvironment"
      consumes:
      - "application/json"
      produces:
      - "application/json"
      parameters:
      - name: "cluster"
        in: "path"
        required: true
        type: "string"
      - name: "environmentName"
        in: "path"
        required: true
        type: "string"
      - in: "body"
        name: "body"
        required: false
        schema:
          $ref: "#/definitions/UpdateEnvironmentRequest"
      responses:
        200:
          description: "successful operation"
          schema:
            $ref: "#/definitions/UpdateEnvironmentResponse"
definitions:
  UpdateEnvironmentResponse:
    type: "object"
  UpdateEnvironmentRequest:
    type: "object"
    properties:
      taskDefinition:
        type: "string"
```

then the generated client will take a request that looks like:

```
class UpdateEnvironmentRequest {
  String cluster
  String environmentName
  UpdateEnvironmentRequest updateEnvironmentRequest
}
```

[...]

PR #292, opened 2017-11-14
Frontend API skeleton

wjbuys:
This PR contains an outline for the Blox frontend API, and the associated generated Swagger and Client code.

The easiest way to explore the swagger is probably by pasting [the URL for the swagger.yml in this pull request](https://raw.githubusercontent.com/wjbuys/blox/b063e1cd4f730f3f2a28c4766c072f6a9d29fa35/frontend-service/api/swagger.yml) into the [online Swagger editor](https://editor.swagger.io/).

Everything under `frontend-service-client` is generated code from the `Re-generate the frontend-service-client` commit, and is only included to make the tests compile; no need to review it closely.

Implements #280


frontend-service/src/main/java/com/amazonaws/blox/frontend/models/wrappers/DescribeEnvironmentDeploymentResponse.java
```diff
@@ -0,0 +1,26 @@
...
+package com.amazonaws.blox.frontend.models.wrappers;
+
+import com.amazonaws.blox.frontend.models.Deployment;
+import java.util.List;
+import lombok.Builder;
+import lombok.Value;
+
+@Value
+@Builder
+public class DescribeEnvironmentDeploymentResponse {
```
wjbuys: I think it might make sense to move these to static inner classes of the controller class that uses them; they're really hard to navigate around when they're over here. Any objection?

wjbuys:
The failing test passes on my desktop, looking into it.

wjbuys:
I'm going to squash all the generated code commits together, otherwise it's really hard to see the actual code changes.

wjbuys:
Okay, all the generated code is now in the single `Re-generate the frontend-service-client` commit, so it should be easy to review separately.

frontend-service/build.gradle
```diff
@@ -45,7 +47,17 @@ task swagger(type: GenerateSwaggerModel, dependsOn: classes) {
 
     scanClasspath = project.sourceSets.main.runtimeClasspath
 
-    apiClasses.add 'com.amazonaws.blox.frontend.controllers.EnvironmentController'
+    // TODO: Specify API classes by package rather than name
+    apiClasses += [
+            'com.amazonaws.blox.frontend.operations.CreateEnvironment',
+            'com.amazonaws.blox.frontend.operations.DescribeEnvironment',
+            'com.amazonaws.blox.frontend.operations.DescribeEnvironmentDeployment',
```
narehayrapetyan, line 54: does this have a corresponding dataservice API?

frontend-service/src/main/java/com/amazonaws/blox/frontend/models/Environment.java
```diff
@@ -14,9 +14,13 @@
  */
 package com.amazonaws.blox.frontend.models;
 
+import lombok.Builder;
 import lombok.Value;
 
 @Value
+@Builder
 public final class Environment {
```
narehayrapetyan, line 22: accountId?

frontend-service/src/main/java/com/amazonaws/blox/frontend/models/EnvironmentRevision.java
```diff
@@ -0,0 +1,33 @@
...
+
+@Value
+@Builder
+public class EnvironmentRevision {
+  private final String cluster;
+  private final String environmentName;
+  private final String environmentRevisionId;
+  private final String taskDefinition;
+  private final InstanceGroup instanceGroup;
+  private final String environmentType;
```
narehayrapetyan: since the type doesn't change the dataservice models it as part of the environment object.

frontend-service/src/main/java/com/amazonaws/blox/frontend/operations/CreateEnvironment.java
```diff
@@ -0,0 +1,70 @@
...
+    // TODO: Stub implementation, replace with dataservice call
+    String id =
+        getAwsAccountId() + "_" + request.getEnvironmentName() + "_" + UUID.randomUUID().toString();
+
+    return CreateEnvironmentResponse.builder().environmentRevisionId(id).build();
+  }
+
+  @Value
+  @Builder
+  public static class CreateEnvironmentResponse {
```
narehayrapetyan, line 53: why are you embedding the models in the same file?

frontend-service/src/main/java/com/amazonaws/blox/frontend/operations/CreateEnvironment.java
```diff
@@ -0,0 +1,70 @@
...
+
+  @RequestMapping(method = RequestMethod.POST, consumes = "application/json")
+  @ApiOperation(value = "Create a new Environment", notes = CREATE_ENVIRONMENT_NOTES)
+  public CreateEnvironmentResponse createEnvironment(
+      @ApiParam(name = "cluster", value = CLUSTER_PARAM_DESCRIPTION) @PathVariable String cluster,
+      @ApiParam(required = true) @RequestBody CreateEnvironmentRequest request) {
+
+    // TODO: Stub implementation, replace with dataservice call
+    String id =
+        getAwsAccountId() + "_" + request.getEnvironmentName() + "_" + UUID.randomUUID().toString();
```
narehayrapetyan: you can use a string joiner here

narehayrapetyan: shouldn't cluster be part of this too?

narehayrapetyan (approved):
Minor comments, fix as you see fit.

frontend-service/build.gradle
wjbuys: Not yet, I think. I added it for the API review to get feedback, but we'll probably only implement it later.

frontend-service/src/main/java/com/amazonaws/blox/frontend/models/Environment.java
wjbuys: We don't expose the accountId to the customer, do we? IIRC it's only used for internal multitenancy purposes.

frontend-service/src/main/java/com/amazonaws/blox/frontend/models/EnvironmentRevision.java
wjbuys: Good catch, will fix.

frontend-service/src/main/java/com/amazonaws/blox/frontend/operations/CreateEnvironment.java
wjbuys: I'm going to get rid of all of this immediately after this PR, but sure.

wjbuys: It should indeed; though now that we have an EnvironmentId object in the DataService model, I think we can use that.

frontend-service/src/main/java/com/amazonaws/blox/frontend/operations/CreateEnvironment.java
wjbuys: The request/response objects are only used in the controllers that define them, and I found it handy to be able to see them in the file itself (like you would with method parameters, if this were just a POJO). Any particular reason to not embed them?

PR #295, opened 2017-12-06
Updating cucumber definitions with new environment behavior

narehayrapetyan:
environment revisions are not exposed through APIs anymore and created and modified through create/updateEnvironment APIs


integ-tests/src/cucumber/resources/features/dataservice/describe_environment.feature
```diff
@@ -3,13 +3,27 @@
 @describe-environment
 Feature: Describe environment
 
-  Scenario: Describe an existing environment
+  Scenario: Describe a newly created environment
     Given I create an environment
     When I describe the environment
     Then the created and described environments match
```
wjbuys: As we discussed in person yesterday, these specs might be a little easier to verify/understand with respect to the UX if we include more details about what's included in the environment/revision in at least one of the test scenarios. Something that seems like it could work well is to allow specifying concrete values for the create/describe calls, but with Feature-wide defaults if omitted. For example:

```gherkin
  Background:
    Given default environment values:
      | environmentType  | Daemon                |
      | deploymentMethod | ReplaceAfterTerminate |
      | taskDefinition   | daemon-v1             |
      | instanceGroup    | query=foo             |

  Scenario: Creating an environment for the first time
    Given I create an environment named "SomeEnvironment":
      | environmentType  | Daemon                |
      | deploymentMethod | ReplaceAfterTerminate |
      | taskDefinition   | daemon-v1             |
      | instanceGroup    | query=foo             |
    When I describe the environment named "SomeEnvironment"
    Then the environment is:
      | environmentName  | "SomeEnvironment" |
      | environmentType  | Daemon            |
      | targetRevisionId | null              |
    And the environment's revisions are:
      | environmentName | deploymentMethod      | taskDefinition | instanceGroup |
      | SomeEnvironment | ReplaceAfterTerminate | daemon-v1      | query=foo     |

  Scenario: Updating an existing environment
    Given I create an environment named "SomeEnvironment":
      | taskDefinition | daemon-v1 |
    When I update the environment named "SomeEnvironment":
      | taskDefinition | daemon-v2 |
    And I describe the environment named "SomeEnvironment"
    Then the environment's revisions are:
      | environmentName | taskDefinition |
      | SomeEnvironment | daemon-v1      |
      | SomeEnvironment | daemon-v2      |
```

wjbuys (commented):
Good start, but I think the specifications are a little too vague to verify against our requirements as it stands.

While it's possible to encode more concrete things in the actual code that implement the scenarios, I think it makes sense to be a bit more concrete about the shape of our data in at least one Scenario of every Feature.

wjbuys (approved):

PR #296, opened 2017-12-07
Moving dataservice interface and model to old namespace to aid with r…

narehayrapetyan:
…efactoring

Moving to a new namespace since there is quite a bit to refactor to move to a new data model and not expose the env revision APIs and this allows us to do in chunks and not break the build.


wjbuys (approved):
I wonder, might it be easier to just rename the entire `com.amazonaws.blox.dataservicemodel.v1` package? That way the later changes to the code that consumes it can just change the package imports and fix the resulting inconsistencies.

narehayrapetyan:
yea good idea

PR #297, opened 2017-12-08
Adding a new dataservice model

narehayrapetyan:
Instead of refactoring the old model and client, adding a new interface with the APIs changed to satisfy the data model where the environment has a 1:n relationship to environmentRevisions.


data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/exception/EnvironmentActiveException.java
```diff
@@ -0,0 +1,22 @@
...
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "LICENSE" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.blox.dataservicemodel.v1.exception;
+
+public class EnvironmentActiveException extends Exception {
```
wjbuys: Can we extract a common superclass for exceptions thrown by our code here? That will make it significantly easier to do exception mapping in the frontend.

Usually it's helpful to have both ServiceException and ClientException superclasses to make it easy for consumers to know whether or not they should retry an exception.

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/client/DataService.java
```diff
@@ -0,0 +1,82 @@
...
+
+  /** Deletes the provided environment if inactive or if forceDelete is true */
+  DeleteEnvironmentResponse deleteEnvironment(DeleteEnvironmentRequest request)
+      throws EnvironmentNotFoundException, EnvironmentActiveException, InvalidParameterException,
+          ServiceException;
+
+  /** Returns the requested environment revision. */
+  DescribeEnvironmentRevisionResponse describeEnvironmentRevision(
+      DescribeEnvironmentRequest request)
+      throws EnvironmentNotFoundException, EnvironmentRevisionNotFoundException,
```
wjbuys: I think it might be worth avoiding resource-specific subclasses of NotFoundException. Perhaps model these as `ResourceNotFoundException` with `resourceType` and `resourceId` fields?

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/exception/EnvironmentRevisionNotFoundException.java
```diff
@@ -0,0 +1,22 @@
...
+ * or in the "LICENSE" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.blox.dataservicemodel.v1.exception;
+
+public class EnvironmentRevisionNotFoundException extends Exception {
+
+  public EnvironmentRevisionNotFoundException(String message) {
```
wjbuys: Instead of taking a message here, can we rather add an `environmentId` field and take that as constructor argument instead? That makes it easier for clients to find out exactly which revision wasn't found, and removes the need to have the full error message formatting code in the code that throws this exception.

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/exception/InvalidParameterException.java
```diff
@@ -0,0 +1,22 @@
...
+ * or in the "LICENSE" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.blox.dataservicemodel.v1.exception;
+
+public class InvalidParameterException extends Exception {
+
+  public InvalidParameterException(String message) {
```
wjbuys: Instead of taking a message here, can we rather take a separate `parameterNames` field?

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/wrappers/CreateEnvironmentRequest.java
```diff
@@ -0,0 +1,39 @@
...
+// required for builder
+@AllArgsConstructor
+// required for mapstruct
+@NoArgsConstructor
+public class CreateEnvironmentRequest {
+
+  @NonNull private EnvironmentId environmentId;
+  @NonNull private String role;
+  @NonNull private EnvironmentType environmentType;
+  @NonNull private String taskDefinition;
```
wjbuys, line 37: At the data service level, should we perhaps require that this has been resolved to a full ARN? In general, I think it might make things simpler if all IDs passed into the data service were ARNs already, so that we don't have to deal with maybe-arn-maybe-name semantics beyond this interface.

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/exception/ServiceException.java
```diff
@@ -0,0 +1,26 @@
...
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "LICENSE" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.blox.dataservicemodel.v1.exception;
+
+public class ServiceException extends Exception {
```
wjbuys, line 17: If we have ClientException/ServiceException superclasses, it might be worth renaming this to `InternalServiceException` to clarify that it's a fatal, internal exception.

wjbuys (changes requested):
Structure looks good, I've got some quick recommendations for the exception modeling from my experiences using the previous iteration of the dataservice API. In general, these are:
- for domain-specific exceptions, prefer providing IDs/field names to the exception constructor, and construct the message in the exception constructor.
- have ClientException/ServiceException superclasses that allows clients to know whether an exception is safe to retry, and to interception/mapping of exceptions in the frontend

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/client/DataService.java
narehayrapetyan: I think it's useful to have the specific exceptions on the dataservice to aid with debugging. They can be mapped to resourcenotfound on the FE

narehayrapetyan: I guess that would make it hard to populate which resource on the frontend :/

wjbuys: Ayup :/

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/wrappers/CreateEnvironmentRequest.java
narehayrapetyan: what's an id here? environmentid? the taskdef and role are arns.

wjbuys: Specifically the identifiers for entities that aren't owned by blox, in this case. So `role`, `taskDefinition` and `cluster`. Maybe? Maybe it's more a documentation/implementation detail. Not a blocker for this CR, but worth keeping in mind for implementation.

wjbuys (approved):
Looks good!

PR #298, opened 2017-12-08
Add first-pass of Daemon ReplaceAfterTerminate scheduler

wjbuys:
This adds the start of a scheduler that runs a Task on every instance in a cluster. The scheduler is named ReplaceAfterTerminate, but it doesn't quite do that yet.

### Out of scope
- It will only start a Task if the same version isn't already running, but it won't kill old Tasks yet.
- It runs the Task on all instance, not just the ones that match the InstanceGroup
- It will schedule as many Tasks as possible at a time, it doesn't honour DeploymentConfiguration.


build.gradle
```diff
@@ -51,6 +51,16 @@ subprojects {
                 entry 'mapstruct-processor'
             }
 
+            // The AWS SDK v2 BOM is incorrectly pinning Cucumber to an
+            // old version, so we have to manually override the version
+            // of all Cucumber dependencies here.
+            dependencySet(group: 'info.cukes', version: '1.2.5') {
+                entry 'cucumber-core'
+                entry 'cucumber-java'
```
narehayrapetyan: do we need both java and java8?

scheduling-manager/src/cucumber/resources/features/scheduling/replace_after_terminate.feature
```diff
@@ -0,0 +1,38 @@
+Feature: Deploying a Daemon environment to a cluster with Replace After Terminate
```
narehayrapetyan, line 1: should we move this to the integ-test package?

scheduling-manager/src/cucumber/resources/features/scheduling/replace_after_terminate.feature
```diff
@@ -0,0 +1,38 @@
...
+
+  Background:
+    Given a Daemon environment named "DaemonEnvironment":
+      | clusterArn        | TestCluster           |
```
narehayrapetyan: I think the values go below the title not next to

scheduling-manager/src/cucumber/resources/features/scheduling/replace_after_terminate.feature
```diff
@@ -0,0 +1,38 @@
...
+      | deploymentMethod  | ReplaceAfterTerminate |
+      | taskDefinitionArn | v1                    |
+
+  Scenario: Deploying a new environment for the first time to an empty cluster
+    Given a cluster named "TestCluster" with the following instances and tasks:
```
narehayrapetyan: is the name important here?

scheduling-manager/src/cucumber/resources/features/scheduling/replace_after_terminate.feature
```diff
@@ -0,0 +1,38 @@
...
+      | TestCluster | i-1                  | v1                | DaemonEnvironment |
+      | TestCluster | i-2                  | v1                | DaemonEnvironment |
+    And it should not take any further actions
+
+  Scenario: Maintaining running tasks after they're already runnning
+    Given a cluster named "default" with the following instances and tasks:
+      | instance | tasks          |
+      | i-1      | t-1:v1:PENDING |
+      | i-2      | t-2:v1:RUNNING |
+    When the scheduler runs
```
narehayrapetyan: should we specify what's running? are we going to test recon separately from startDeployment loop for example?

scheduling-manager/src/cucumber/resources/features/scheduling/replace_after_terminate.feature
```diff
@@ -0,0 +1,38 @@
...
+  Scenario: Restarting failed tasks
+    Given a cluster named "default" with the following instances and tasks:
+      | instance | tasks          |
+      | i-1      | t-1:v1:STOPPED |
+      | i-2      | t-2:v1:RUNNING |
+    When the scheduler runs
+    Then it should start the following tasks:
+      | containerInstanceArn | taskDefinitionArn | group             |
+      | i-1                  | v1                | DaemonEnvironment |
+    And it should not take any further actions
```
narehayrapetyan, line 39: should we also test that all the expected tasks are still running?

scheduling-manager/src/cucumber/resources/features/scheduling/replace_after_terminate.feature
narehayrapetyan, line 1: we should add a TODO: add update env test cases

scheduling-manager/src/cucumber/java/cucumber/steps/scheduling/daemon/ReplaceAfterTerminateSteps.java
```diff
@@ -0,0 +1,123 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class ReplaceAfterTerminateSteps implements En {
+
+  private EnvironmentDescription environment;
+  private ClusterSnapshot snapshot;
+  private Scheduler scheduler = new ReplaceAfterTerminateScheduler();
```
narehayrapetyan, line 41: autowire?

scheduling-manager/src/cucumber/java/cucumber/steps/scheduling/daemon/ReplaceAfterTerminateSteps.java
```diff
@@ -0,0 +1,123 @@
...
+
+  private Set<SchedulingAction> actions = new HashSet<>();
+
+  public ReplaceAfterTerminateSteps() {
+    Given(
+        "^a Daemon environment named \"([^\"]*)\":$",
+        (String name, DataTable properties) -> {
+          EnvironmentDescriptionBuilder builder = environmentDescriptionFromTable(properties);
+
+          environment = builder.environmentName(name).build();
```
narehayrapetyan, line 51: shouldn't this be pulled from dataservice?

scheduling-manager/src/cucumber/java/cucumber/steps/scheduling/daemon/ReplaceAfterTerminateSteps.java
```diff
@@ -0,0 +1,123 @@
...
+        "^the scheduler runs$",
+        () -> {
+          actions.clear();
+          actions.addAll(scheduler.schedule(snapshot, environment));
+        });
+
+    Then(
+        "^it should start the following tasks:$",
+        (DataTable startTasksTable) -> {
+          List<StartTask> startTasks = startTaskActionsFromTable(startTasksTable);
```
narehayrapetyan, line 76: so we're not actually checking that these tasks were started? shouldn't we?

narehayrapetyan (changes requested):
What is the external interface to the scheduler? What does the recon event call? It doesn't seem like the integration tests are testing the end to end loop of the scheduler getting triggered and the tasks getting started on ecs.

build.gradle
wjbuys: Yes, [`cucumber-java8` depends on `cucumber-java`](https://mvnrepository.com/artifact/info.cukes/cucumber-java8/1.2.5), and without pinning the version of `cucumber-java` we get errors because of version mismatches between the two.

wjbuys:
I think I introduced some confusion here: The Cucumber features in this package aren't integration tests, they're unit tests. The textual format of the cucumber feature definitions just make it much easier to clearly express the behaviour of the scheduler. I'll add some actual integration tests that do test the scheduler end-to-end, but they won't be doing the fine-grained "when the cluster is in this state, this should happen" verifications that this cucumber suite does.

I'll try and document this peculiarity somewhere as we extend this scheduler for more realistic use cases.

scheduling-manager/src/cucumber/resources/features/scheduling/replace_after_terminate.feature
wjbuys: That's implicit in the "should not take any further actions" bit. The scheduling loop is purely functional, it can't modify the state of the cluster apart from the actions it emits.

scheduling-manager/src/cucumber/resources/features/scheduling/replace_after_terminate.feature
wjbuys: Both the reconciliation and startDeployment events eventually trigger this exact same code with the full cluster state; they don't have different code paths apart from the outer bit.

scheduling-manager/src/cucumber/resources/features/scheduling/replace_after_terminate.feature
wjbuys: Cucumber places no restriction on how you parse the DataTable; the underlying test code can parse these values however it makes sense.

When we're specifying a large number of attributes on a small number (i.e. 1) of objects, it's more readable to have the table transposed like this as a large number of rows, rather than having a large number of columns.

wjbuys: (But, if it's confusing to see both styles in the same spec, it's easy to change, just let me know)

scheduling-manager/src/cucumber/resources/features/scheduling/replace_after_terminate.feature
wjbuys: As mentioned in the comment below, these aren't actually integration tests, they're unit tests written in cucumber so that we can use tables for these very data-driven test cases.

scheduling-manager/src/cucumber/java/cucumber/steps/scheduling/daemon/ReplaceAfterTerminateSteps.java
wjbuys: See comment below; these are unit tests, and as such won't verify that anything gets started.

scheduling-manager/src/cucumber/java/cucumber/steps/scheduling/daemon/ReplaceAfterTerminateSteps.java
wjbuys: It will, in the integration tests; these are unit tests.

scheduling-manager/src/cucumber/resources/features/scheduling/replace_after_terminate.feature
wjbuys: Hm, I think this is a vestige of when it was important. Will remove.

(Would ideally like to make it optional, but looks like that's not particularly easy)

scheduling-manager/src/cucumber/resources/features/scheduling/replace_after_terminate.feature
wjbuys: Sure.

narehayrapetyan:
I think it's confusing that part of our unit tests are in cucumber and some are not, don't you? IMO we should keep it consistent e2e and integ tests in cucumber and unit tests just regular junit.

scheduling-manager/src/cucumber/resources/features/scheduling/replace_after_terminate.feature
narehayrapetyan: you can use junit parameterized tests for the data driven part

scheduling-manager/src/cucumber/resources/features/scheduling/replace_after_terminate.feature
narehayrapetyan: oh I see. Yea it's a little confusing to have it both ways

wjbuys:
Originally I had these in JUnit, and they were about twice as long, and very hard to read/understand easily. I'm a big fan of using whatever tool gives useful, concise, readable tests for the thing I'm testing. In this case, I think that that's Cucumber. Cucumber is just a testing tool for expressing behaviour in a human-readable way, it doesn't enforce the level of abstraction/interaction of the tests.

I think our integration and end-to-end tests are well separated in the clearly named top-level `end-to-end-tests` and `integ-tests` projects. Would adding a note in `scheduling-manager/README.md` help clarify things? I want to add one anyway as a place to put the docs on how this all works.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.List;
+import java.util.Set;
+import lombok.RequiredArgsConstructor;
+
+/**
+ * Wrapper around {@link EnvironmentDescription} that makes it easy to match {@link
+ * com.amazonaws.blox.scheduling.state.ClusterSnapshot.Task} instances to a Daemon environment.
+ */
+@RequiredArgsConstructor
+class DaemonEnvironment {
```
narehayrapetyan: public

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/ClusterSummary.java
```diff
@@ -0,0 +1,58 @@
...
+import com.amazonaws.blox.scheduling.state.ClusterSnapshot;
+import com.amazonaws.blox.scheduling.state.ClusterSnapshot.ContainerInstance;
+import com.amazonaws.blox.scheduling.state.ClusterSnapshot.Task;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+class ClusterSummary {
```
narehayrapetyan: public

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/ClusterSummary.java
```diff
@@ -0,0 +1,58 @@
...
+  private final ClusterSnapshot snapshot;
+  private final Map<String, List<Task>> tasksByInstanceArn;
+  private final Map<String, ContainerInstance> instancesByInstanceArn;
+
+  public ClusterSummary(ClusterSnapshot snapshot) {
```
narehayrapetyan, line 32: when is this created?

wjbuys: This wrapper is created in the abstract `DaemonScheduler` class's `schedule` template method.

The idea is that ReplaceAfterTerminate and TerminateAfterReplace will both extend that abstract class with the specific workflow logic, but reuse these same wrappers to work with the flat ClusterSnapshot data.

wjbuys:
The latest commit adds a JUnit test that does what the Cucumber feature spec did. If that looks better to you, let me know and I'll remove the cucumber stuff before merging.

wjbuys:
As discussed, we're keeping the cucumber specs for this particular case. I've removed the duplicate JUnit tests, and cleaned up the cucumber specs.

scheduling-manager/src/cucumber/java/cucumber/steps/scheduling/daemon/ReplaceAfterTerminateSteps.java
narehayrapetyan: so why are we not autowiring all the dependencies?

narehayrapetyan (approved):

scheduling-manager/src/cucumber/java/cucumber/steps/scheduling/daemon/ReplaceAfterTerminateSteps.java
wjbuys: Apologies, though I commented on this one. This code has no dependencies, apart from the scheduler it's testing, so there's no benefit to autowiring.

PR #300, opened 2017-12-12
Update environment and environmentRevision table schemas

narehayrapetyan:
template successfully runs and creates resources


data-service/templates/data_service.yml
```diff
@@ -2,60 +2,68 @@ AWSTemplateFormatVersion: '2010-09-09'
...
          KeyType: HASH
-       - AttributeName: "environmentVersion"
+       - AttributeName: "environmentName"
          KeyType: RANGE
      ProvisionedThroughput:
        ReadCapacityUnits: 15
        WriteCapacityUnits: 15
+     GlobalSecondaryIndexes:
+       -
+         IndexName: "environment-cluster-index"
```
wjbuys: Since we're renaming things, perhaps use the same CamelCase naming as the tables themselves?

data-service/templates/data_service.yml
```diff
@@ -2,60 +2,68 @@ AWSTemplateFormatVersion: '2010-09-09'
...
          AttributeType: S
-       - AttributeName: environmentVersion
+       - AttributeName: environmentName
+         AttributeType: S
+       - AttributeName: accountId
+         AttributeType: S
+       - AttributeName: cluster
          AttributeType: S
      KeySchema:
        - AttributeName: "environmentId"
```
wjbuys, line 20: This won't be an `environmentId` anymore right? The hash key should be `accountId + clusterName` (no environment-specific value there) which unambiguously identifies a cluster, so perhaps `clusterId`?

data-service/templates/data_service.yml
```diff
@@ -2,60 +2,68 @@ AWSTemplateFormatVersion: '2010-09-09'
...
+           ReadCapacityUnits: 15
+           WriteCapacityUnits: 15
 
-  EnvironmentTargetVersionTable:
+  EnvironmentRevisionTable:
     Type: AWS::DynamoDB::Table
     Properties:
      #TODO: change ddb table names to have region prefix passed in as a resource
-      TableName: EnvironmentTargetVersion
+      TableName: EnvironmentRevision
```
wjbuys: Should this be `EnvironmentRevisions` to match the plural `Environments` table?

wjbuys (approved):
Just some naming consistency issues, otherwise LGTM.

data-service/templates/data_service.yml
narehayrapetyan: yea i was going to name the field environmentId in the ddb table but its going to be constructed out of accountId_clusterName

data-service/templates/data_service.yml
narehayrapetyan: sure, I think this is the default ddb naming scheme but I can switch it to be consistent

data-service/templates/data_service.yml
narehayrapetyan: sure

PR #301, opened 2017-12-12
Adding ddbmapper models for environment and environmentrevision

narehayrapetyan:
ddb mapper model corresponding to the schema defined in the [data service sam template](https://github.com/blox/blox/blob/dev/data-service/templates/data_service.yml)


data-service/src/main/java/com/amazonaws/blox/dataservice/repository/model/EnvironmentDDBRecord.java
```diff
@@ -0,0 +1,96 @@
...
+        .environmentId(environmentId)
+        .environmentName(environmentName)
+        .build();
+  }
+
+  public static EnvironmentDDBRecord withGSIKeys(final String accountId, final String cluster) {
+    return EnvironmentDDBRecord.builder().accountId(accountId).clusterName(cluster).build();
+  }
+
+  @DynamoDBHashKey(attributeName = ENVIRONMENT_ID_HASH_KEY)
```
wjbuys: Won't DDBMapper infer the column name from the field name here? Why repeat the name in a string constant?

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/model/EnvironmentDDBRecord.java
```diff
@@ -0,0 +1,96 @@
...
+  private String environmentId;
+
+  @DynamoDBRangeKey(attributeName = ENVIRONMENT_NAME_RANGE_KEY)
+  private String environmentName;
+
+  @DynamoDBIndexHashKey(
+    attributeName = ENVIRONMENT_CLUSTER_INDEX_HASH_KEY,
+    globalSecondaryIndexName = ENVIRONMENT_CLUSTER_GSI_NAME
```
wjbuys: If I understand correctly, since we're re-using this POJO for the index, we'll get `EnvironmentDDBRecord` objects back from a query against the index, but with `null` values for all fields not projected into the index.

Perhaps it would make more sense to create a separate Record class (`EnvironmentClusterIndexDDBRecord`?) for the type of objects stored in the index? This is what [the docs suggest](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBMapper.Methods.html).

narehayrapetyan: Where does the doc suggest creating a separate object? IMO it creates additional complexity with having to map to a different object when creating/reading this record.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/model/EnvironmentDDBRecord.java
narehayrapetyan: Yes but the field names can be different from the table values like cluster/clusterName so in general I've used attritbutenames everywhere and not just where they differ to be consistent.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/model/EnvironmentDDBRecord.java
wjbuys: Wouldn't it be confusing that a query for `listAllClusters` against the DB would return a set of `EnvironmentDDBRecord` objects?

For the docs, sorry, I missed the direct link to the section, [here](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBMapper.Methods.html#DynamoDBMapper.Methods.query) it is:

> To query an index, you must first model the index as a mapper class. Suppose that the Reply table has a global secondary index named `PostedBy-Message-Index`. The partition key for this index is `PostedBy`, and the sort key is `Message`. The class definition for an item in the index would look like this:

```java
@DynamoDBTable(tableName="Reply")
public class PostedByMessage { 
    private String postedBy;
    private String message;

    @DynamoDBIndexHashKey(globalSecondaryIndexName = "PostedBy-Message-Index", attributeName = "PostedBy")
    public String getPostedBy() { return postedBy; }
    public void setPostedBy(String postedBy) { this.postedBy = postedBy; } 

    @DynamoDBIndexRangeKey(globalSecondaryIndexName = "PostedBy-Message-Index", attributeName = "Message")
    public String getMessage() { return message; }
    public void setMessage(String message) { this.message = message; } 

   // Additional properties go here. 
}
```
> The `@DynamoDBTable` annotation indicates that this index is associated with the Reply table. The `@DynamoDBIndexHashKey` annotation denotes the partition key (PostedBy) of the index, and `@DynamoDBIndexRangeKey` denotes the sort key (Message) of the index.

>Now you can use DynamoDBMapper to query the index, retrieving a subset of messages that were posted by a particular user. You must specify withIndexName so that DynamoDB knows which index to query. In the following code snippet, we are querying a global secondary index. Because global secondary indexes support eventually consistent reads, but not strongly consistent reads, we must specify withConsistentRead(false).
```java
HashMap<String, AttributeValue> eav = new HashMap<String, AttributeValue>();
eav.put(":v1",  new AttributeValue().withS("User A"));
eav.put(":v2",  new AttributeValue().withS("DynamoDB"));

DynamoDBQueryExpression<PostedByMessage> queryExpression = new DynamoDBQueryExpression<PostedByMessage>()
    .withIndexName("PostedBy-Message-Index")
    .withConsistentRead(false)
    .withKeyConditionExpression("PostedBy = :v1 and begins_with(Message, :v2)")
    .withExpressionAttributeValues(eav);

List<PostedByMessage> iList =  mapper.query(PostedByMessage.class, queryExpression);
```
> The query returns a collection of PostedByMessage objects.

You would not need to manually map this class, since you'd use it as the query entrypoint.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/model/EnvironmentDDBRecord.java
wjbuys: Wait, why would we want those names to be inconsistent in the first place? I mean, these `*DDBRecord` classes are *already* the at the level of abstraction where they should exactly correspond with the DB Schema. IIRC the main reason to use the explicit attribute names was to interact with legacy DBs that had e.g. `snake_case` column names.

wjbuys (approved):
I think the changes we discussed can be incorporated later. Right now I'd really like the updated DataService interface for the frontend integ.

wjbuys (approved):

PR #305, opened 2017-12-15
Create environment in repository

narehayrapetyan:
./gradlew build


data-service/src/main/java/com/amazonaws/blox/dataservice/model/EnvironmentId.java
```diff
@@ -0,0 +1,49 @@
...
+  public String generateAccountIdCluster() {
+    return new StringJoiner(UNDERSCORE).add(accountId).add(cluster).toString();
+  }
+
+  public static String getAccountIdFromAccountIdCluster(String accountIdCluster) {
+    return accountIdCluster.split(UNDERSCORE)[0];
+  }
+
+  public static String getClusterFromAccountIdCluster(String accountIdCluster) {
+    return accountIdCluster.split(UNDERSCORE)[1];
```
wjbuys: This should probably pass `2` as the second argument to `split()`, to ensure it doesn't split on underscores in the cluster name.

wjbuys (approved):

data-service/src/main/java/com/amazonaws/blox/dataservice/model/EnvironmentId.java
narehayrapetyan: 2 as the limit you mean?

PR #306, opened 2017-12-15
Adding cucumber test helper methods to store inputs from previous steps

narehayrapetyan:
example steps and helper classes

./gradlew build & cucumber


integ-tests/src/cucumber/resources/features/dataservice/create_environment.feature
```diff
@@ -4,7 +4,7 @@
 Feature: Create environment
 
   Scenario: Create an environment
-    When I create an environment named "test"
+    When I create an environment
```
xjhe, line 7: Should we also change the describe_environment.feature to not include the environment name "test"?

integ-tests/src/cucumber/java/steps/dataservice/CreateEnvironmentSteps.java
```diff
@@ -14,29 +14,72 @@
...
-          throw new NotImplementedException("");
+          dataServiceWrapper.createEnvironment(createEnvironmentRequest());
         });
 
-    Then(
-        "^the created environment response is valid$",
-        () -> {
-          throw new NotImplementedException("");
+    When(
+        "^I create an environment named \"([^\"]*)\"$",
```
xjhe, line 48: In the create_environment.feature, we already remove the "named 'test'". So here in steps, should we still keep the name, or should we also remove it?

xjhe (commented):
Some minor comments and questions.

integ-tests/src/cucumber/java/steps/dataservice/CreateEnvironmentSteps.java
narehayrapetyan: we have steps both with the name and without the name being specified that's why we have both definitions

integ-tests/src/cucumber/resources/features/dataservice/create_environment.feature
narehayrapetyan: I tried only keeping a specific name in the definitions if it makes the test easier to understand, so, for example, we create an env named test and describe the env named test. One thing that makes it confusing is we need to create and env prefixed with test and not just named test so the tests can be run in parallel and not fail because of artifacts other tests have created. I'm wondering if we should say 'prefixed with test' instead of 'named test' in the definitions but then it defeats the purpose of trying to signify it's the same environment.

wjbuys: Can we set up cucumber to generate us a unique prefix per test run? That way, each individual scenario execution can use the same random identifier as prefix to avoid collisions, but the test itself can just say `named: "blah"`.

xjhe: I see. I think the prefix is already different each run, since we use UUID for the prefix in the step definition? So I'll take the same pattern for describe environment as well.

integ-tests/src/cucumber/java/steps/dataservice/CreateEnvironmentSteps.java
xjhe: I see. Makes sense!

integ-tests/src/cucumber/resources/features/dataservice/create_environment.feature
wjbuys: Makes sense.

integ-tests/build.gradle
```diff
@@ -17,12 +17,14 @@ dependencies {
     compile project(":data-service-client")
 
     cucumberCompile(
+        'com.google.guava:guava:23.0',
```
wjbuys, line 20: Doesn't look like anything is using this?

integ-tests/src/cucumber/resources/features/dataservice/create_environment.feature
narehayrapetyan: yes, @xjhe, correct. My concern is that future tests being implemented will skip the prefix since it's not specified in the definition.

wjbuys (approved):

integ-tests/build.gradle
narehayrapetyan: multimap

PR #308, opened 2017-12-15
Wire up create/describe Environment/Revision to DataService

wjbuys:
This wires up the createEnvironment, describeEnvironment and
describeEnvironmentRevision API calls from the Frontend to the
DataService. It also adds the necessary mappers for request/response
types.

This implements #280.


frontend-service/src/main/java/com/amazonaws/blox/frontend/mappers/CreateEnvironmentMapper.java
```diff
@@ -0,0 +1,36 @@
...
+import org.mapstruct.Mapping;
+
+@Mapper(uses = InstanceGroupMapper.class)
+public interface CreateEnvironmentMapper {
+  @Mapping(target = "environmentId.accountId", source = "context.accountId")
+  @Mapping(target = "environmentId.cluster", source = "cluster")
+  @Mapping(target = "environmentId.environmentName", source = "request.environmentName")
+  CreateEnvironmentRequest toDataServiceRequest(
+      ApiGatewayRequestContext context,
+      String cluster,
```
narehayrapetyan, line 30: createenvrequest doesn't contain cluster?

frontend-service/src/main/java/com/amazonaws/blox/frontend/mappers/DescribeEnvironmentRevisionMapper.java
```diff
@@ -0,0 +1,38 @@
...
+@Mapper(uses = EnvironmentRevisionMapper.class)
+public interface DescribeEnvironmentRevisionMapper {
+
+  @Mapping(target = "environmentId.accountId", source = "context.accountId")
+  @Mapping(target = "environmentId.cluster", source = "cluster")
+  @Mapping(target = "environmentId.environmentName", source = "environmentName")
+  DescribeEnvironmentRevisionRequest toDataServiceRequest(
+      ApiGatewayRequestContext context,
+      String cluster,
+      String environmentName,
```
narehayrapetyan, line 33: why are we taking these as separate fields?

frontend-service/src/main/java/com/amazonaws/blox/frontend/operations/CreateEnvironment.java
```diff
@@ -41,12 +46,12 @@
...
-      @ApiParam(required = true) @RequestBody CreateEnvironmentRequest request) {
+      @ApiParam(required = true) @RequestBody CreateEnvironmentRequest request)
+      throws InvalidParameterException, ServiceException, ResourceExistsException {
 
-    // TODO: Stub implementation, replace with dataservice call
-    String id = UUID.randomUUID().toString();
-
-    return CreateEnvironmentResponse.builder().environmentRevisionId(id).build();
+    return mapper.fromDataServiceResponse(
+        dataService.createEnvironment(
```
narehayrapetyan, line 53: don't we need to map the exceptions into frontend exceptions? and log?

frontend-service/src/main/java/com/amazonaws/blox/frontend/operations/EnvironmentController.java
```diff
@@ -24,27 +25,19 @@
...
-  }
+  @Autowired DataService dataService;
 
   /**
    * Get the API gateway request context from the current request.
    *
    * @return
    */
-  private ApiGatewayRequestContext getApiGatewayRequestContext() {
+  protected ApiGatewayRequestContext getApiGatewayRequestContext() {
```
narehayrapetyan, line 39: why protected?

frontend-service/src/main/java/com/amazonaws/blox/frontend/mappers/CreateEnvironmentMapper.java
wjbuys: Nope, it's a Spring MVC `@PathVariable`, since it's part of the URL:

https://github.com/wjbuys/blox/blob/dcc96ab5a9d5d4d36001db984acab2aab47c4075/frontend-service/src/main/java/com/amazonaws/blox/frontend/operations/CreateEnvironment.java#L45-L55

I originally tried to move the `@PathVariable String cluster` argument into the `CreateEnvironmentRequest` object as a field. However, even though the Spring MVC docs suggest that this will correctly be bound, the Swagger generator we're using doesn't support it, and totally omits `cluster` from the generated Swagger model, and so we don't get it in the generated client.

frontend-service/src/main/java/com/amazonaws/blox/frontend/mappers/DescribeEnvironmentRevisionMapper.java
wjbuys: Same problem as above: since they are part of the URL, they must be passed in as simple `@PathVariable` parameters to the frontend SpringMVC handler method.

frontend-service/src/main/java/com/amazonaws/blox/frontend/operations/CreateEnvironment.java
wjbuys: Spring exception mapping is complicated, and this PR was already pretty big, so I opted to do it as a followup to this PR. There's an open task for it on #280, which we'll probably have to split into a separate issue, since it's a big chunk of work.

frontend-service/src/main/java/com/amazonaws/blox/frontend/operations/EnvironmentController.java
wjbuys: It's used by subclasses of EnvironmentController to get at the API gateway request context. Previously subclasses only used the `getAccountId()` method, which is now no longer used.

narehayrapetyan (approved):

PR #312, opened 2017-12-20
New PR on describe environment steps

xjhe:
TODO
#### Summary

This is the complete implementation of describe environment steps.

#### Implementation details

One thing to check/discuss is that to make sure the steps have the same environment name, I generate the random name postfix at the beginning. Not sure if this is the correct way to do it. 

#### Testing

New tests cover the changes: 

#### Licensing


integ-tests/src/cucumber/java/steps/dataservice/DescribeEnvironmentSteps.java
```diff
@@ -0,0 +1,203 @@
...
+@ContextConfiguration(classes = CucumberConfiguration.class)
+public class DescribeEnvironmentSteps implements En {
+
+  @Autowired private DataServiceWrapper dataServiceWrapper;
+  private static final String ACCOUNT_ID = "accountID";
+  private static final String TASK_DEFINITION_ARN =
+      "arn:aws:ecs:us-east-1:" + ACCOUNT_ID + ":task-definition/sleep";
+  private static final String ROLE_ARN = "arn:aws:iam::" + ACCOUNT_ID + ":role/testRole";
+  private static final String CLUSTER_NAME_PREFIX = "cluster";
+  private String postfix = null;
```
narehayrapetyan: I don't think you need to store it in a class variable since you can retrieve the name from the memoized wrapper

integ-tests/src/cucumber/java/steps/dataservice/DescribeEnvironmentSteps.java
```diff
@@ -0,0 +1,203 @@
...
+
+  public DescribeEnvironmentSteps() {
+    Given(
+        "^I create an environment named \"([^\"]*)\"$",
```
narehayrapetyan: So you actually don't need to implement this step since it's implemented in the createSteps already. Cucumber should detect step definitions in all steps

integ-tests/src/cucumber/java/steps/dataservice/DescribeEnvironmentSteps.java
```diff
@@ -0,0 +1,203 @@
...
+    When(
+        "^I describe the environment named \"([^\"]*)\"$",
+        (final String environmentNamePrefix) -> {
+          dataServiceWrapper.describeEnvironment(describeEnvironmentRequest(environmentNamePrefix));
+        });
+
+    Then(
+        "^the created and described environments match$",
+        () -> {
+          final CreateEnvironmentResponse createEnvironmentResponse =
```
narehayrapetyan: so currently we're not saving the responses. if you look at the memoizedwrapper implementation you'll see we're only saving the request objects. You should extend it to save both then this will work.

integ-tests/src/cucumber/java/steps/dataservice/DescribeEnvironmentSteps.java
```diff
@@ -0,0 +1,203 @@
...
+        "^I try to describe a non-existent environment named \"([^\"]*)\"$",
+        (final String environmentNamePrefix) -> {
+          dataServiceWrapper.tryDescribeEnvironment(
+              describeEnvironmentRequest(environmentNamePrefix));
+        });
+
+    Given(
+        "^I delete the created environment named \"([^\"]*)\"$",
+        (final String environmentNamePrefix) -> {
+          dataServiceWrapper.tryDeleteEnvironment(deleteEnvironmentRequest(environmentNamePrefix));
```
narehayrapetyan: if the step is not described with "try" you shouldn't call the wrapper try methods since they will expect the call to fail.

integ-tests/src/cucumber/java/steps/dataservice/DescribeEnvironmentSteps.java
```diff
@@ -0,0 +1,203 @@
...
+
+    When(
+        "^I try to describe the environment named \"([^\"]*)\"$",
+        (final String environmentNamePrefix) -> {
+          dataServiceWrapper.tryDescribeEnvironment(
+              describeEnvironmentRequest(environmentNamePrefix));
+        });
+  }
+
+  private DescribeEnvironmentRequest describeEnvironmentRequest(
```
narehayrapetyan: We can move the request wrappers into the dataservicewrapper class and reuse them from all steps

integ-tests/src/cucumber/java/steps/dataservice/DescribeEnvironmentSteps.java
```diff
@@ -0,0 +1,203 @@
...
+import steps.wrappers.DataServiceWrapper;
+import static org.junit.Assert.assertTrue;
+
+import java.util.UUID;
+
+@ContextConfiguration(classes = CucumberConfiguration.class)
+public class DescribeEnvironmentSteps implements En {
+
+  @Autowired private DataServiceWrapper dataServiceWrapper;
+  private static final String ACCOUNT_ID = "accountID";
```
narehayrapetyan: this will fail when we add input validation since accountids need to be 12 digit numbers. We can use the accountId that the tests are going to run as. for now, we can just use 12345678912

narehayrapetyan (changes requested):
Nice job! You also need to implement the exception steps right?

integ-tests/src/cucumber/java/steps/dataservice/DescribeEnvironmentSteps.java
xjhe: Sure! good to know! I'll change this.

integ-tests/src/cucumber/java/steps/dataservice/DescribeEnvironmentSteps.java
xjhe: I see. Yes getting it from the wrapper will be cleaner.

integ-tests/src/cucumber/java/steps/dataservice/DescribeEnvironmentSteps.java
xjhe: I didn't know that cucumber is that smart :-D  Then I'll just leave out this.

integ-tests/src/cucumber/java/steps/dataservice/DescribeEnvironmentSteps.java
xjhe: That's true. I'll modify the wrapper to save both.

integ-tests/src/cucumber/java/steps/dataservice/DescribeEnvironmentSteps.java
xjhe: Sure!

integ-tests/src/cucumber/java/steps/dataservice/DescribeEnvironmentSteps.java
xjhe: Sure. I can move them into the wrappers.

integ-tests/src/cucumber/java/steps/dataservice/DescribeEnvironmentSteps.java
xjhe: Just realize that 12345678912 is only 11 digit. I'll use 123456789012 instead.

xjhe:
Had a discussion with Nare and summarize the fix for this PR (that are not covered in the comments)
1. Don't use "name" for some cases anymore; Just get the full name (prefix + UUID) from the memoized wrapper;
2. Move helper methods into a separate class like inputCreator;
3. Create an exception wrapper to store exceptions.

xjhe:
Update this PR with my changes:
1. Address comments on naming/minor changes;
2. Move most of the helper methods from CreateEnviornmentSteps/DescribeEnvironmentSteps into InputCreator to allow reuse of these methods.
3. Also saves responses in the DataSerfviceWrapper. 
4. Create the exception handling steps in CommonStep. Currently I only wrote the steps for handling ResourceNotFoundException.

integ-tests/build.gradle
```diff
@@ -11,6 +11,7 @@ addCucumberSuite 'cucumber'
 
 cucumber {
     tags = ["~@ignore"]
+    stepDefinitionRoots = ['steps']
```
narehayrapetyan: if you pull in my changes this won't be necessary anymore

integ-tests/src/cucumber/java/steps/dataservice/CommonSteps.java
```diff
@@ -0,0 +1,56 @@
...
+        (final String exceptionName) -> {
+          assertNotNull("Expecting an exception to be thrown", exceptionContext.getException());
+          assertEquals(exceptionName, exceptionContext.getException().getClass().getSimpleName());
+        });
+
+    And(
+        "^the resourceType is \"([^\"]*)\"$",
+        (final String resourceType) -> {
+          assertNotNull("Expecting an exception to be thrown", exceptionContext.getException());
+          final ResourceNotFoundException exception =
```
narehayrapetyan: so this is not always a resourceNotFoundException since other exceptions like resourceExists also have this structure with type and id. You'll have to get the thrown exception and cast it to that type.

integ-tests/src/cucumber/java/steps/dataservice/DescribeEnvironmentSteps.java
```diff
@@ -0,0 +1,119 @@
...
+              createEnvironmentResponse.getEnvironment(),
+              describeEnvironmentResponse.getEnvironment());
+        });
+
+    Given(
+        "^I update the created environment$",
+        (final String environmentNamePrefix) -> {
+          final EnvironmentId environmentId = getEnvironmentIdFromCreatedEnvironment();
+          dataServiceWrapper.updateEnvironment(
+              inputCreator.updateEnvironmentRequest(
```
narehayrapetyan: doesn't this need to contain a field that you're updating? it might make sense to pass in an environment object so you don't have to add all possible updatable fields to the method signature

narehayrapetyan (commented):
This looks great. Can you merge in the latest changes and address the comments and then you'll be good to go.

integ-tests/src/cucumber/java/steps/dataservice/CommonSteps.java
xjhe: The exception with resourceId/resourceType are ResourceNotFoundException, ResourceExistException, ResourceInUseException. I can check the type of the thrown exception and cast it into one of the types

wjbuys: If that turns out to be a bit icky, consider using [assertJ](http://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html) for assertions (see examples in `scheduling-manager` cucumber tests), which will let you do assertions on properties without casting:

```java
assertThat(exceptionContext.getException()).hasFieldOrPropertyWithValue("resourceType", resourceType);
```

integ-tests/src/cucumber/java/steps/dataservice/DescribeEnvironmentSteps.java
xjhe: Looks like I can't pass the whole environment object in? The parameter of the step definition need to be the same as the one in the feature definitions, which can only be string/numbers. 

I can modify this step into "I update the created environment with new task definition "new task definition" "

wjbuys (approved):
LGTM

PR #313, opened 2018-01-05
Implementing create environment steps and fixing cucumber config

narehayrapetyan:
./gradlew build


integ-tests/src/cucumberTest/java/cucumber/steps/wrappers/MemoizedWrapper.java
```diff
@@ -12,32 +12,37 @@
...
 
-  @Override
-  public <T> T getFromHistory(Class<T> type, Function<List<Object>, Object> fn) {
-    return null;
-  }
-
   @Override
   public <T> T getLastFromHistory(Class<T> type) {
-    return null;
+    return (T) memory.get(type).get(memory.size() - 1);
```
xjhe: Should this line be memory.get(type).get(memory.get(type).size() - 1) ?
the memory.size() will be the number of all entries for all keys.

narehayrapetyan: you're right! great catch

wjbuys (approved):

PR #314, opened 2018-01-09
Create environment and environment revision

narehayrapetyan:
./gradlew deploy and ./gradlew cucumberTest: create environment and validate answer invokes dataservice in lambda and the test passes.

Manually tested partial failure cases by modifying values in dynamodb:

* environment does not exist
* environment exists but environment revision doesn't
* environment and revision exist but environment is not valid
* environment and revision exist and environment is valid

Most of the logic happens in the environmentRepositoryDDB


data-service/src/main/java/com/amazonaws/blox/dataservice/api/CreateEnvironmentApi.java
```diff
@@ -0,0 +1,86 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Component;
+
+@Slf4j
+@Component
+@AllArgsConstructor
+public class CreateEnvironmentApi {
+
+  @NonNull private final ApiModelMapper apiModelMapper;
+  @NonNull private EnvironmentRepository environmentRepository;
```
xjhe: Can environmentRepository be "final" as well?

data-service/src/main/java/com/amazonaws/blox/dataservice/api/CreateEnvironmentApi.java
```diff
@@ -0,0 +1,86 @@
...
+          .environment(apiModelMapper.toWrapperEnvironment(createdEnvironment))
+          .environmentRevision(apiModelMapper.toWrapperEnvironmentRevision(environmentRevision))
+          .build();
+
+    } catch (final ResourceExistsException | InternalServiceException e) {
+      log.error(e.getMessage(), e);
+      throw e;
+    } catch (final Exception e) {
+      log.error(e.getMessage(), e);
+      throw new InternalServiceException(e.getMessage(), e);
```
xjhe, line 83: Here we put all other type of exceptions into "InternalServiceException", while one exception that we catch from environmentRepository.createEnvironmentAndEnvironmentRevision() is also "InternalServiceException". Should we separate these two cases into different types of exceptions?

data-service/src/main/java/com/amazonaws/blox/dataservice/model/EnvironmentId.java
```diff
@@ -39,11 +39,20 @@ public String generateAccountIdCluster() {
     return new StringJoiner(UNDERSCORE).add(accountId).add(cluster).toString();
   }
 
+  public String generateAccountIdClusterEnvironmentName() {
+    return new StringJoiner(UNDERSCORE).add(accountId).add(cluster).add(environmentName).toString();
+  }
+
   public static String getAccountIdFromAccountIdCluster(String accountIdCluster) {
```
xjhe, line 46: In the EnvironmentMapper.java (line 67) we only use EnvironmentId.getAccountIdFromAccountIdClusterEnvironmentName(). So do we still need the getAccountIdFromAccountIdCluster() method here?

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
```diff
@@ -37,8 +51,98 @@
...
+      @NonNull final EnvironmentRevision environmentRevision)
       throws ResourceExistsException, InternalServiceException {
+
+    createEnvironment(environment);
+    createEnvironmentRevision(environmentRevision);
+
+    return updateEnvironmentToValid(environment);
+  }
+
+  private Environment updateEnvironmentToValid(@NonNull final Environment environment)
```
xjhe: This may be beyond the scope of this PR, but I would like to clarify -- there is an updateEnvironment in DataServiceApi, so do we need a general updateEnvironment utility here in our later implementations? Looks like this one only update the environment to valid, which is quite specific.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/model/EnvironmentDDBRecord.java
```diff
@@ -38,20 +38,21 @@
 @NoArgsConstructor
 public class EnvironmentDDBRecord {
 
-  public static final String ENVIRONMENT_ID_HASH_KEY = "environmentId";
+  public static final String ACCOUNT_ID_CLUSTER_HASH_KEY = "accountIdCluster";
```
xjhe, line 41: Why do we change the hash key from environmentId into accountIdCluster? Is it likely that under the same accountId and cluster there are multiple environments with different environment names?

xjhe (commented):
I make some comments and questions to this PR -- thank you for finishing up EnvironmentRepositoryDDB 💯

data-service/src/main/java/com/amazonaws/blox/dataservice/model/EnvironmentId.java
xjhe: Please ignore this comment -- I figure that the accountIDCluster is the new key so there's no more questions on this line.

data-service/src/main/java/com/amazonaws/blox/dataservice/api/CreateEnvironmentApi.java
narehayrapetyan: yep

data-service/src/main/java/com/amazonaws/blox/dataservice/api/CreateEnvironmentApi.java
narehayrapetyan: we certainly can have another layer of internal exceptions and not use external exceptions like resourceNotFound inside the data service but it's a question of balance between separation and too many object conversions. It's unlikely the shape of the exception will change between external and internal and as long as it's clear from the stack where the exception occurs, I decided to go with less exception types and conversions.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
narehayrapetyan: Good point, refactored into a generic update.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/model/EnvironmentDDBRecord.java
narehayrapetyan: So this was always composed of accountId and cluster but it was confusing because it was called environmentId and did not match the environmentId of the environment object.

data-service/src/main/java/com/amazonaws/blox/dataservice/api/CreateEnvironmentApi.java
xjhe: Understand! This makes sense.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/model/EnvironmentDDBRecord.java
xjhe: Yes that's true. So what we really want as the key here is not the actual EnvironmentId object. Thank you for clarifying this :-D

data-service/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Configuration packages="com.amazonaws.services.lambda.runtime.log4j2.LambdaAppender">
```
wjbuys: If you rebase this on the latest `dev` commit, you can omit this configuration (it's included in the `lambda-logging` project).

wjbuys (approved):

PR #319, opened 2018-01-10
Implemented DescribeEnvironmentApi

xjhe:
TODO
#### Summary
Describe Environment API! Quite simple change. 

#### Implementation details
Takes in EnvironmentID and return the DescribeEnvironmentResponse

#### Testing
Unit test. Need to merge the integ test PR for integration test. 

New tests cover the changes: 

#### Licensing

### DIFF
--- a/data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/exception/ResourceNotFoundException.java
+++ b/data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/exception/ResourceNotFoundException.java
@@ -20,7 +20,7 @@ public class ResourceNotFoundException extends ClientException {
   private String resourceId;
 
   public ResourceNotFoundException(String resourceType, String resourceId) {
-    super(String.format("%s with id %s could not be found"));
+    super(String.format("%s with id %s could not be found", resourceType, resourceId));
 
     this.resourceType = resourceType;
     this.resourceId = resourceId;
--- a/data-service/src/main/java/com/amazonaws/blox/dataservice/api/DescribeEnvironmentApi.java
+++ b/data-service/src/main/java/com/amazonaws/blox/dataservice/api/DescribeEnvironmentApi.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may
+ * not use this file except in compliance with the License. A copy of the
+ * License is located at
+ *
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "LICENSE" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.blox.dataservice.api;
+
+import com.amazonaws.blox.dataservice.mapper.ApiModelMapper;
+import com.amazonaws.blox.dataservice.model.Environment;
+import com.amazonaws.blox.dataservice.model.EnvironmentId;
+import com.amazonaws.blox.dataservice.repository.EnvironmentRepository;
+import com.amazonaws.blox.dataservicemodel.v1.exception.InternalServiceException;
+import com.amazonaws.blox.dataservicemodel.v1.exception.ResourceNotFoundException;
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.DescribeEnvironmentRequest;
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.DescribeEnvironmentResponse;
+
+import lombok.NonNull;
+import lombok.AllArgsConstructor;
+import lombok.Value;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+@Value
+@AllArgsConstructor
+public class DescribeEnvironmentApi {
+  @NonNull private ApiModelMapper apiModelMapper;
+  @NonNull private EnvironmentRepository environmentRepository;
+
+  public DescribeEnvironmentResponse describeEnvironment(
+      final DescribeEnvironmentRequest describeEnvironmentRequest)
+      throws ResourceNotFoundException, InternalServiceException {
+    final com.amazonaws.blox.dataservicemodel.v1.model.EnvironmentId environmentIdFromRequest =
+        describeEnvironmentRequest.getEnvironmentId();
+    final EnvironmentId environmentId =
+        apiModelMapper.toModelEnvironmentId(environmentIdFromRequest);
+    try {
+      final Environment environment = environmentRepository.getEnvironment(environmentId);
+      return DescribeEnvironmentResponse.builder()
+          .environment(apiModelMapper.toWrapperEnvironment(environment))
+          .build();
+    } catch (final ResourceNotFoundException | InternalServiceException e) {
+      log.error(e.getMessage(), e);
+      throw e;
+    } catch (final Exception e) {
+      log.error(e.getMessage(), e);
+      throw new InternalServiceException(e.getMessage(), e);
+    }
+  }
+}
--- a/data-service/src/main/java/com/amazonaws/blox/dataservice/api/DescribeEnvironmentApi.java
+++ b/data-service/src/main/java/com/amazonaws/blox/dataservice/api/DescribeEnvironmentApi.java
@@ -24,19 +24,17 @@ import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.DescribeEnvironment
 import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.DescribeEnvironmentResponse;
 
 import lombok.NonNull;
-import lombok.AllArgsConstructor;
 import lombok.Value;
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 @Value
-@AllArgsConstructor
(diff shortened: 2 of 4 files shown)


data-service/src/main/java/com/amazonaws/blox/dataservice/api/DescribeEnvironmentApi.java
```diff
@@ -0,0 +1,58 @@
...
+import lombok.NonNull;
+import lombok.AllArgsConstructor;
+import lombok.Value;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+@Value
+@AllArgsConstructor
+public class DescribeEnvironmentApi {
+  @NonNull private ApiModelMapper apiModelMapper;
```
wbingli: with `final`?

data-service/src/main/java/com/amazonaws/blox/dataservice/api/DescribeEnvironmentApi.java
```diff
@@ -0,0 +1,58 @@
...
+  @NonNull private EnvironmentRepository environmentRepository;
```
wbingli: same with `final`

data-service/src/test/java/com/amazonaws/blox/dataservice/api/DescribeEnvironmentApiTest.java
```diff
@@ -0,0 +1,142 @@
...
+    describeEnvironmentApi.describeEnvironment(describeEnvironmentRequest);
+  }
+
+  @Test(expected = InternalServiceException.class)
+  public void describeEnvironmentInternalServiceException() throws Exception {
+    when(environmentRepository.getEnvironment(isA(EnvironmentId.class)))
+        .thenThrow(new InternalServiceException(""));
+    describeEnvironmentApi.describeEnvironment(describeEnvironmentRequest);
+  }
+}
```
wbingli: Should add one more test case to cover catching unknown exception and wrap to InternalServiceException.

data-service/src/test/java/com/amazonaws/blox/dataservice/api/DescribeEnvironmentApiTest.java
```diff
@@ -0,0 +1,142 @@
...
+            .deploymentMethod(DEPLOYMENT_METHOD)
+            .validEnvironment(true)
+            .build();
+    describeEnvironmentRequest =
+        DescribeEnvironmentRequest.builder().environmentId(environmentIdWrapper).build();
+  }
+
+  @Test
+  public void describeEnvironmentSuccess() throws Exception {
+    when(environmentRepository.getEnvironment(isA(EnvironmentId.class))).thenReturn(environment);
```
wbingli: could minimal the condition to `when(environmentRepository.getEnvironment(environmentId))`

data-service/src/main/java/com/amazonaws/blox/dataservice/api/DescribeEnvironmentApi.java
xjhe: Sure!

data-service/src/main/java/com/amazonaws/blox/dataservice/api/DescribeEnvironmentApi.java
xjhe: Sure. Will also check other variables.

data-service/src/main/java/com/amazonaws/blox/dataservice/api/DescribeEnvironmentApi.java
narehayrapetyan: so @value actually ensures that all fields are final so it's okay to omit it

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/exception/ResourceNotFoundException.java
```diff
@@ -20,7 +20,7 @@
   private String resourceId;
 
   public ResourceNotFoundException(String resourceType, String resourceId) {
-    super(String.format("%s with id %s could not be found"));
+    super(String.format("%s with id %s could not be found", resourceType, resourceId));
```
narehayrapetyan, line 23: whoops good catch

data-service/src/main/java/com/amazonaws/blox/dataservice/api/DescribeEnvironmentApi.java
```diff
@@ -0,0 +1,58 @@
...
+
+  public DescribeEnvironmentResponse describeEnvironment(
+      final DescribeEnvironmentRequest describeEnvironmentRequest)
```
narehayrapetyan: you can add a @nonnull here

data-service/src/test/java/com/amazonaws/blox/dataservice/api/DescribeEnvironmentApiTest.java
```diff
@@ -0,0 +1,142 @@
...
+    final DescribeEnvironmentResponse describeEnvironmentResponse =
+        describeEnvironmentApi.describeEnvironment(describeEnvironmentRequest);
+
+    verify(environmentRepository).getEnvironment(environmentIdArgumentCaptor.capture());
```
narehayrapetyan: I think you can omit the captor and pass in the expected object directly and that will compare the fields

narehayrapetyan (approved):
Looks great! Minor comments

data-service/src/main/java/com/amazonaws/blox/dataservice/api/DescribeEnvironmentApi.java
xjhe: That's true...I should pay more attention to the difference between @data and @ value

data-service/src/main/java/com/amazonaws/blox/dataservice/api/DescribeEnvironmentApi.java
```diff
@@ -0,0 +1,58 @@
...
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.DescribeEnvironmentResponse;
+
+import lombok.NonNull;
+import lombok.AllArgsConstructor;
+import lombok.Value;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+@Value
+@AllArgsConstructor
```
gongmax: Is this needed? Since @Value will introduce a AllArgsConstructor if no explicit constructor is defined.

xjhe: That's correct! Thank you for pointing out. I was reading the page on @value and figure that it's actually quite different from @ data... 
@ value equals 
 @Getter @FieldDefaults(makeFinal=true, level=AccessLevel.PRIVATE) @AllArgsConstructor @ToString @EqualsAndHashCode

data-service/src/test/java/com/amazonaws/blox/dataservice/api/DescribeEnvironmentApiTest.java
xjhe: That's right!

data-service/src/main/java/com/amazonaws/blox/dataservice/api/DescribeEnvironmentApi.java
xjhe: Sure!

data-service/src/test/java/com/amazonaws/blox/dataservice/api/DescribeEnvironmentApiTest.java
xjhe: Good point. Done.

data-service/src/test/java/com/amazonaws/blox/dataservice/api/DescribeEnvironmentApiTest.java
xjhe: Fixed!

data-service/src/test/java/com/amazonaws/blox/dataservice/api/DescribeEnvironmentApiTest.java
```diff
@@ -0,0 +1,144 @@
...
+  @Test(expected = InternalServiceException.class)
+  public void describeEnvironmentInternalServiceException() throws Exception {
+    when(environmentRepository.getEnvironment(isA(EnvironmentId.class)))
+        .thenThrow(new InternalServiceException(""));
+    describeEnvironmentApi.describeEnvironment(describeEnvironmentRequest);
+  }
+
+  @Test(expected = InternalServiceException.class)
+  public void describeEnvironmentInternalServiceExceptionWithUnknownException() throws Exception {
+    when(environmentRepository.getEnvironment(isA(EnvironmentId.class))).thenReturn(null);
```
wbingli: `Return(null)` looks not intuitive to understand what this test about, kind of hack as following code will cause NPE.   Would be good to use `when(environmentRepository.getEnvironment(isA(EnvironmentId.class))).thenThrow(new IllegalStateException(""))`

xjhe: I was thinking of letting environmentRepository.getEnvironment() throw some other exceptions. Just not sure if we should also test the exceptions that are not from environmentRepository.getEnvironment(). 
I can keep both unit test cases and rename them into something more intuitive like

describeEnvironmentUnknownException()

and 

describeEnvironmentOtherExceptions() ?

narehayrapetyan:
Also you should add unit tests for the environmentRepository methods you're using. Sorry didn't have time to finish those up.

wbingli (approved):

PR #321, opened 2018-01-11
Add missing fields to FE model/mappers

wjbuys:
Merging #308 after #314 without rebasing and retesting first introduced
some logical conflicts that caused the build to fail. Specifically, #314
introduced new mandatory fields on various data service model classes,
which weren't being set in the frontend service unit tests or mappers.

This adds those fields, which fixes the build.


wbingli (approved):
LGTM

PR #322, opened 2018-01-11
Add getEnvironmentRevision DDB operation

wbingli:
#### Summary
This is the DDB level implementation of `getEnvironmentRevision`.  We will need this method for both DescribeEnvironmentRevision and StartDeployment.  

The StartDeployment will need this method to make validation and throw not found if given revision doesn't exist.

#### Implementation details
Besides the DDB implementation, I also fix a few mapper warning for some of them we don't want to map.

#### Testing
`./gradlew build`

#### Licensing


data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
```diff
@@ -0,0 +1,129 @@
...
+            .accountId(ACCOUNT_ID)
+            .cluster(CLUSTER)
+            .environmentName(ENVIRONMENT_NAME)
+            .build();
+  }
+
+  @Test
+  public void testGetEnvironmentRevision() throws Exception {
+    // Given
+    final String environmentRevisionId = "revision-id";
```
xjhe: Could you bring the environmentRevisionId into a static member variable, instead of instantiating it in every test? looks like it's the same across all three unit tests.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
```diff
@@ -68,6 +68,32 @@ public Environment updateEnvironment(Environment environment)
...
+              environmentRevisionId);
+      if (environmentRevisionRecord == null) {
+        throw new ResourceNotFoundException(
+            ResourceType.ENVIRONMENT_REVISION, environmentRevisionId);
+      }
+      return environmentMapper.toEnvironmentRevision(environmentRevisionRecord);
+    } catch (final AmazonServiceException e) {
+      throw new InternalServiceException(
+          String.format(
+              "Could not load record with environment revision key %s and environment revision id %s",
```
xjhe, line 91: Can we log the environment name here as well? It may give the user more information

data-service/src/main/java/com/amazonaws/blox/dataservice/model/Environment.java
```diff
@@ -38,7 +38,7 @@
   @NonNull private EnvironmentStatus environmentStatus;
   @NonNull private String latestEnvironmentRevisionId;
   // environment and revision were both successfully created
-  @NonNull private boolean validEnvironment;
+  private boolean validEnvironment;
```
xjhe, line 41: Why do we remove the @NonNull for this variable

wbingli: It's a primitive type, so never be null and it cause a build warning.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
wbingli: The key here actually is an accountIDClusterEnvionrment hash key,  so it contains all the information including environment name.

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
wbingli: make sense, will do.

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/exception/ResourceNotFoundException.java
```diff
@@ -14,6 +14,9 @@
  */
 package com.amazonaws.blox.dataservicemodel.v1.exception;
 
+import lombok.EqualsAndHashCode;
+
+@EqualsAndHashCode
```
gongmax, line 19: Why this is needed

wbingli: Good catch. I need to verify the exception in the unit test.  The `@EqualsAndHashCode` could make the verify simpler to just use 

```
thrown.expect(is(new ResourceNotFoundException(
                ResourceType.ENVIRONMENT_REVISION, environmentRevisionId)));
```

gongmax: I think you can use thrown.expectMessage(....) to check the exception message.

wbingli: yes, the `expectMessage` method can verify the exception message. But this exception message is implemented inside the `ResourceNotFoundException` class, not part of the logic in class for testing.

Use the `thrown.expect` it can only verify the exception instance I created in the `EnvironmentRepositoryDDB` class, and make sure it pass the right arguments to build the exception.

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
```diff
@@ -0,0 +1,129 @@
...
+            environmentId.generateAccountIdClusterEnvironmentName(),
+            ENVIRONMENT_REVISION_ID);
+    verify(environmentMapper).toEnvironmentRevision(environmentRevisionDDBRecord);
+    assertThat(result, is(environmentRevision));
+  }
+
+  @Test
+  public void testGetEnvironmentRevisionNotFound() throws Exception {
+    // Given
+    final String environmentRevisionId = "revision-id";
```
xjhe: Should probably remove this and use ENVIRONMENT_REVISION_ID instead? Same for the other unit test.

xjhe (approved):
I think this looks good.

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/exception/ResourceNotFoundException.java
gongmax: Make sense!

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
wbingli: Oh, forgot this. Just Updated.

PR #323, opened 2018-01-12
Make listClusters return a List<Cluster>

wjbuys:
A cluster name doesn't uniquely identify a cluster without the owning
account ID, so we have to return both values from the listClusters API
call in the data service.


gongmax (approved):

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/Cluster.java
```diff
@@ -0,0 +1,32 @@
...
+
+@Data
+@Builder
+// required for builder
+@AllArgsConstructor
+// required for mapstruct
+@NoArgsConstructor
+public class Cluster {
+  @NonNull private String accountId;
+  @NonNull private String cluster;
```
xjhe: Can we change the name of this field into "clusterName" ? I was a little bit confused when looking at this change as both the class name and this attribute are called "Cluster". 
Also want to confirm, in the "cluster" field of an environment, we should still just keep the cluster name instead of the whole Cluster object?

xjhe: Well if we still just use the cluster name in the Environment object, I think we can just keep it as "cluster" here to avoid further confusion....or change both into "clusterName" :-D

wjbuys: Yeah, it's really inconsistent right now. I'm in favor of being clearer; let's use:

- `clusterName` if it's only the name
- `clusterArn` if it's actually a full ARN
- `cluster` if it's a String in the same format as what the ECS API accepts (i.e. either name of fully qualified ARN), or if it's a Cluster object.

wjbuys: > Also want to confirm, in the "cluster" field of an environment, we should still just keep the cluster name instead of the whole Cluster object?

Environment is a bit of a special case, since it contains `accountId` itself. I think the MapStruct mapping between types might become really annoying if we make EnvironmentId be something like `(cluster: Cluster, environmentName: String)`

PR #324, opened 2018-01-12
Implement DS StartDeployment API

wbingli:
#### Summary
This is implementation for data-service for start-deployment API.   Not actually create Deployment, use a fake Deployment id,  but it will set the active environment revision.

Not add too many validation for the API, such as whether we should allow start-deployment with the same active environment revision. 

The rollback consider as a different story, we may use the same API but with extra API parameter to indicate it's a rollback action.

#### Implementation details
 - Added 'org.assertj:assertj-core:3.8+' for testCompile to use assertj in the data-service.
 - Need to fetch both environment and environment-revision DDB record so it make sure those resource exists. Otherwise it will thrown ResourceNotFound exception.

#### Testing
`./gradlew build`

#### Licensing


data-service/src/main/java/com/amazonaws/blox/dataservice/api/StartDeploymentApi.java
```diff
@@ -0,0 +1,73 @@
...
+
+@Slf4j
+@Component
+@AllArgsConstructor
+public class StartDeploymentApi {
+
+  @NonNull private final ApiModelMapper apiModelMapper;
+  @NonNull private final EnvironmentRepository environmentRepository;
+
+  public StartDeploymentResponse startDeployment(final StartDeploymentRequest request)
```
gongmax: NonNull?

data-service/src/main/java/com/amazonaws/blox/dataservice/api/StartDeploymentApi.java
```diff
@@ -0,0 +1,73 @@
...
+      final EnvironmentId environmentId =
+          apiModelMapper.toModelEnvironmentId(request.getEnvironmentId());
+      final Environment environment = environmentRepository.getEnvironment(environmentId);
+
+      final EnvironmentRevision environmentRevision =
+          environmentRepository.getEnvironmentRevision(
+              environmentId, request.getEnvironmentRevisionId());
+
+      environment.setActiveEnvironmentRevisionId(environmentRevision.getEnvironmentRevisionId());
+
```
gongmax, line 53: Is the "lastUpdatedTime" in Environment needed to be updated?

data-service/src/main/java/com/amazonaws/blox/dataservice/api/StartDeploymentApi.java
```diff
@@ -0,0 +1,73 @@
...
+      environmentRepository.updateEnvironment(environment);
```
gongmax, line 54: This is predated your CR but I'm thinking about this: in environmentRepository.updateEnvironment(environment), dynamoDBMapper.load is called once to get the lastest recordVersion. This can truly get the optimistic lock but also add an extra transaction to DDB. Since we have to do a load in startDeploymentApi, which is the immediate upper level of environmentRepository.updateEnvironment, do you think we need load it again in environmentRepository.updateEnvironment?

data-service/src/main/java/com/amazonaws/blox/dataservice/api/StartDeploymentApi.java
wbingli: yes, will update

data-service/src/main/java/com/amazonaws/blox/dataservice/api/StartDeploymentApi.java
wbingli: It should be updated, but I would like this update will be done on DB API level, which the `EnvironmentRepositoryDDB#updateEnvironment' will update this field.

I will do other PR for the missing part of the updateEnvironment method.

gongmax: I kind of lean to have all of the fields that need to be updated be set in the data-service api level, and leave the DB layer only take care of interacting with DB. But open to discussion.

data-service/src/main/java/com/amazonaws/blox/dataservice/api/StartDeploymentApi.java
wbingli: Good question.  Looked at current optimistic locking implementation in  `EnvironmentRepositoryDDB#updateEnvironment `,  I think it does not do what we want.

Considering following scenario:
 - Request A: start-deployment on environment E
 - Request B: Do a delete-environment at the same time
 - Request A gets E revision 1, passed validation (we will do it in the StartDeployment API ), environment looks good and ready to update environment
 - Request B get E revision 1, make it as deleted and update DDB. the E record revision is now 2
 - **ISSUE**: in current implementation, say `environmentRepository.updateEnvironment` fetch the E with revision 2, it will make a success call to update the DDB record.

To make optimistic locking actually does it job, we need to make sure the data used for request validation is the one we finally used for update.

I think we should have recordRevision in the Environment model and it will map to the environment record for optimistic checking.

wbingli: Since it's a different issue of this PR, should we have a different PR for the optimistic locking issue?

gongmax: Agree, so I think remove the setRecordVersion piece of code from environmentDepositoryDDB and add recordRevision in environment model mapper will 1: fix the locking issue, 2: avoid extra read from DDB.

data-service/src/main/java/com/amazonaws/blox/dataservice/api/StartDeploymentApi.java
wbingli: I think any call to make a "updateEnvironment" DDB call should update the `lastUpdatedTime`. In this case, we don't have to duplicate the `setLastUpdatedTime(...)`  before every `updateEnvironment` call.   Unless we have different propose of the latestUpdatedTime with the record, it's more like a way to simplify our code.

gongmax (approved):

PR #325, opened 2018-01-12
Add listEnvironments DDB operation

gongmax:
#### Summary
This is the DDB level implementation of listEnvironments. Cluster.java in dataServiceModel is stolen from blox/blox#323.

#### Testing
gradlew build
gradlew check

#### Licensing


data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/Cluster.java
```diff
@@ -0,0 +1,32 @@
...
+@Data
+@Builder
+// required for builder
+@AllArgsConstructor
+// required for mapstruct
+@NoArgsConstructor
+public class Cluster {
+  @NonNull private String accountId;
+  @NonNull private String cluster;
+}
```
wbingli: I'm thinking whether github PR could have code review diff not have to against with the target branch. Like this case,  you may want to only have diff for your change without the Cluster change.  Otherwise, you will have to resolve conflict later on.

data-service/src/main/java/com/amazonaws/blox/dataservice/model/Cluster.java
```diff
@@ -0,0 +1,40 @@
...
+import lombok.NonNull;
+
+@Data
+@Builder
+// required for builder
+@AllArgsConstructor
+// required for mapstruct
+@NoArgsConstructor
+public class Cluster {
+  private static final String UNDERSCORE = "_";
```
wbingli, line 32: Not in scope of this PR, but not sure using underscore as delimiter is good idea, that means we will not allow cluster and environment name to have underscore.   

It will hurt customer as lots of people like to use underscore to name stuff.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepository.java
```diff
@@ -32,6 +33,8 @@ Environment createEnvironmentAndEnvironmentRevision(
   Environment getEnvironment(EnvironmentId environmentId)
       throws ResourceNotFoundException, InternalServiceException;
 
+  List<Environment> listEnvironments(Cluster cluster) throws InternalServiceException;
```
wbingli, line 36: Should also throw `ResourceNotFoundException` for cluster not found?

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
```diff
@@ -306,6 +307,25 @@ public Environment getEnvironment(@NonNull final EnvironmentId environmentId)
...
+      throws InternalServiceException {
+    try {
+      return dynamoDBMapper
+          .query(
+              EnvironmentDDBRecord.class,
+              new DynamoDBQueryExpression<EnvironmentDDBRecord>()
+                  .withHashKeyValues(
+                      EnvironmentDDBRecord.withHashKeys(cluster.generateAccountIdCluster())))
+          .stream()
+          .map(e -> environmentMapper.toEnvironment(e))
```
wbingli: Should be able to use method reference `environmentMapper::toEnvironment`

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/model/EnvironmentDDBRecord.java
```diff
@@ -42,6 +42,10 @@
   public static final String ENVIRONMENT_NAME_RANGE_KEY = "environmentName";
   public static final String LATEST_ENVIRONMENT_REVISION_ID = "latestEnvironmentRevisionId";
 
+  public static EnvironmentDDBRecord withHashKeys(final String accountIdCluster) {
+    return EnvironmentDDBRecord.builder().accountIdCluster(accountIdCluster).build();
```
wbingli, line 46: units?

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
```diff
@@ -124,4 +141,44 @@ public void testGetEnvironmentRevisionInternalError() throws Exception {
...
+  @Test
+  public void testListEnvironments() throws Exception {
+    when(environmentDDBRecords.stream()).thenReturn(Stream.of(environmentDDBRecord));
+    when(dynamoDBMapper.query(eq(EnvironmentDDBRecord.class), any(DynamoDBQueryExpression.class)))
+        .thenReturn(environmentDDBRecords);
+    when(environmentMapper.toEnvironment(environmentDDBRecord)).thenReturn(environment);
+
+    final List<Environment> result = environmentRepositoryDDB.listEnvironments(cluster);
+
+    verify(dynamoDBMapper).query(eq(EnvironmentDDBRecord.class), any());
```
wbingli: It would be better to capture the DynamoDBQueryExpression and you can verify the passed hash key:
```
verify(dynamoDBMapper).query(eq(EnvironmentDDBRecord.class), queryExpCaptor.capture());
assertThat(queryExpCaptor.getValue().getHashKeyValues().getAccountIdCluster()).isEqualsTo(CLUSTER. generateAccountIdCluster())
```

frontend-service/api/swagger.yml
```diff
@@ -428,4 +428,4 @@ definitions:
         type: "string"
       newTargetRevisionId:
         type: "string"
-x-generated-at: "2018-01-09T22:55:42.535Z"
+x-generated-at: "2018-01-09T22:55:42.535Z"
```
xjhe: Do we need to update this line?

wbingli: why this change?  I would think having a new line at the end of file is good practice.  You can enable the "Ensure line feed at file end on save" if using intellij.

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/Cluster.java
gongmax: I copied PR 323 only because this can unblock some change in my PR. And since I copied it, I don't think there will be conflict.

data-service/src/main/java/com/amazonaws/blox/dataservice/model/Cluster.java
gongmax: Good catch, but I don't think this disallow customer to use underscore in their cluster/environment name. We only use the underscore to form the hashkey in DDB, and it seems we do not need to get the cluster/environment name/id from the hashkey.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepository.java
gongmax: What do you mean by "cluster not found"? If there is no record associated with this cluster and account in DDB, this api just return an empty list.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
gongmax: yeah will update

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/model/EnvironmentDDBRecord.java
gongmax: Can you explain more?

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
gongmax: will update

frontend-service/api/swagger.yml
gongmax: I don't know. This is auto-generated, and I don't know what change here is. It seems to me there is no code change here and no line added or deleted.

data-service/src/main/java/com/amazonaws/blox/dataservice/model/Cluster.java
xjhe: I think the actual place that this underscore will matter is in the EnvironmentId object? Because we do try to decompose the accountID and cluster from the key. We may think about the impact there. https://github.com/blox/blox/blob/dev/data-service/src/main/java/com/amazonaws/blox/dataservice/model/EnvironmentId.java#L46

gongmax: Oh you are right. We should reconsider this.

frontend-service/api/swagger.yml
wjbuys: This is auto-updated by [this filter](https://github.com/blox/blox/blob/dev/buildSrc/src/main/java/com/amazonaws/blox/swagger/GenerationTimestampFilter.java), which seemed like a good idea at the time, but now seems like it just makes annoying noise in our diffs.

I think we can remove that filter in a later PR to avoid this issue in future.

data-service/src/main/java/com/amazonaws/blox/dataservice/model/Cluster.java
wjbuys: Agreed. I'd say we should use the same separators as ARNs, so identifiers that are valid in ARNs will also be valid in our DB key.

data-service-model/src/main/java/com/amazonaws/blox/dataservicemodel/v1/model/Cluster.java
wjbuys: Sorry about that, I forgot to actually push that yesterday. I think you'll be able to rebase your PR on that change now.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepository.java
wbingli: I think they are two different things. 

- Like ResouceNotFound for cluster, which means no such clsuter defined. 
- Emtpy list means there is no environment in this cluster, but cluster does exist.

My idea is following the same rule of list api, say list-services:
```
 aws ecs list-services --cluster not-exist

An error occurred (ClusterNotFoundException) when calling the ListServices operation: Cluster not found.
```

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/model/EnvironmentDDBRecord.java
wbingli: oh, just mention we should add unit test for the code change.   It's obvious but it won't hurt to have a simple unit test.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepository.java
wbingli: The implementation would be different as we don't track the cluster.  Would be a front-end service validation thing to validate the cluster with ECS before invoke data service.

data-service/src/main/java/com/amazonaws/blox/dataservice/model/Cluster.java
gongmax: So we will use cluster name not cluster arn across all of our apis and ddb tables?

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepository.java
gongmax: I agree we need validation for non-exist cluster but I don't think we should take care of it at this level. I'm leaning to add it in data-service api rather than this DDB interacting level.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/model/EnvironmentDDBRecord.java
gongmax: Fixed

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepository.java
wbingli: yes, the cluster validation should be on higher level

wbingli (approved):
LGTM

PR #327, opened 2018-01-12
quick fix on the warnings on the exceptions objects

xjhe:
TODO
#### Summary
Fix the warning generated by the @value annotation. 

#### Testing

New tests cover the changes: 

#### Licensing


wbingli (commented):
I think we plan to use @Getter and assertj for comparing the exception fields.

xjhe:
Ah. I changed the annotation to @Getter and I think the ExpectedException will just do the job

wjbuys (approved):

PR #329, opened 2018-01-15
Implement ListEnvironments DS API

gongmax:
TODO
#### Summary
ListEnvironments API data-service side implementation 

#### Testing
./gradlew build
./gradlew check

#### Licensing

### DIFF
--- a/data-service/src/main/java/com/amazonaws/blox/dataservice/api/ListEnvironmentsApi.java
+++ b/data-service/src/main/java/com/amazonaws/blox/dataservice/api/ListEnvironmentsApi.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may
+ * not use this file except in compliance with the License. A copy of the
+ * License is located at
+ *
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "LICENSE" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.blox.dataservice.api;
+
+import com.amazonaws.blox.dataservice.mapper.ApiModelMapper;
+import com.amazonaws.blox.dataservice.model.Cluster;
+import com.amazonaws.blox.dataservice.model.Environment;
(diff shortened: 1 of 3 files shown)


data-service/src/main/java/com/amazonaws/blox/dataservice/api/ListEnvironmentsApi.java
```diff
@@ -0,0 +1,63 @@
...
+  @NonNull private EnvironmentRepository environmentRepository;
+
+  public ListEnvironmentsResponse listEnvironments(@NonNull final ListEnvironmentsRequest request)
+      throws InvalidParameterException, InternalServiceException {
+
+    // TODO validate the cluster
+    final Cluster cluster = apiModelMapper.toModelCluster(request.getCluster());
+
+    try {
+      final List<Environment> environments = environmentRepository.listEnvironments(cluster);
```
xjhe, line 47: This is prior to this PR -- should environmentRepository.listEnvironments throw a ResourceNotFoundException when the query result is null?

xjhe: And if the result is empty, should we return an empty list of environments, or should we throw an exception

xjhe (commented):
Minor comments! Otherwise this looks good

data-service/src/main/java/com/amazonaws/blox/dataservice/api/ListEnvironmentsApi.java
gongmax: if query result is null, environmentRepository.listEnvironments should just return an empty list. ResourceNotFoundException is not suitable here since it's a List* call.
Same for the ListEnvironments DS API, if an empty list of environments is returned from environmentRepository, we just return an empty list.

xjhe: I see. This makes sense!

xjhe (approved):

data-service/src/main/java/com/amazonaws/blox/dataservice/api/ListEnvironmentsApi.java
wbingli: From customer user experience perspective, ListEnvironment API should have a ResourceNotFoundException check on provided cluster.  So we will eventually validate the cluster info.  But we don't store the cluster info, only way we can do is via call ECS api. There will be two options in my mind:

 - FE call ECS API to validate cluster
 - ECS proxy app do the validate before call blox service. 

My first impression should be do the validation on FE  as it make blox service self-contained.  But none of them will be DS service, we can discuss about that on FE API.

PR #331, opened 2018-01-15
Replace the Joiner in EnvironmentId and Cluster from Underscore to sl…

gongmax:
#### Summary
1. Replace the Joiner in Class EnvironmentId, Cluster and AttributeDDBConverter from underscore to slash.
2. Re-organize class ApiModelMapper's import, so now all the imported classes that have the same name in both data-service and data-service-model are from com.amazonaws.blox.dataservice.model.

#### Testing
./gradlew build
./gradlew check

#### Licensing


data-service/src/main/java/com/amazonaws/blox/dataservice/model/Cluster.java
```diff
@@ -29,12 +29,12 @@
 // required for mapstruct
 @NoArgsConstructor
 public class Cluster {
-  private static final String UNDERSCORE = "_";
+  private static final String JOINER = "_";
```
wbingli: Still underscore?

wbingli: maybe to use `DELIMITER` as the var name?

data-service/src/main/java/com/amazonaws/blox/dataservice/model/EnvironmentId.java
```diff
@@ -29,30 +29,30 @@
 @NoArgsConstructor
 public class EnvironmentId {
 
-  private static final String UNDERSCORE = "_";
+  private static final String JOINER = "/";
```
wbingli: Actually AWS ARN doesn't forbid to use "/" in the resource even slash is used for ARN format like `arn:partition:service:region:account-id:resourcetype/resource`, you may see some services use '/' in their resource ID.

But inside ECS,   we don't allow '/' in the name, e..g in the cluster name, it says "The name of your cluster. If you do not specify a name for your cluster, you create a cluster named default. Up to 255 letters (uppercase and lowercase), numbers, hyphens, and underscores are allowed."

I think use slash is good idea so far.  The UNDERSCORE is definitely buggy as cluster name allowed to contain underscore.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/model/AttributeDDBConverter.java
```diff
@@ -20,21 +20,18 @@
 
 public class AttributeDDBConverter implements DynamoDBTypeConverter<String, Attribute> {
 
-  private static final String UNDERSCORE = "_";
+  private static final String JOINER = "_";
```
wbingli: still underscore?

data-service/src/main/java/com/amazonaws/blox/dataservice/model/EnvironmentId.java
xjhe: Should we add this message in Blox front end as well, so people know that '/' is not allowed in names? 
We can do this in another sim...not necessarily in this PR.

wbingli: I don't think we allow it. We can follow the same name requirement as cluster, like "Up to 255 letters (uppercase and lowercase), numbers, hyphens, and underscores are allowed.". In this way,  it basically doesn't allow slash.

xjhe: That's true. My suggestion is that we should output this similar message to the user when they try to use names with slash :-D

data-service/src/main/java/com/amazonaws/blox/dataservice/model/Cluster.java
gongmax: Yeah, that's the word... Fixed.

data-service/src/main/java/com/amazonaws/blox/dataservice/model/EnvironmentId.java
gongmax: We are getting these cluster names only from ECS, so there is no need for Blox to validate the cluster name. But yeah, we definitely need to add validation and return some message to indicate the invalid parameters when customer call createEnvironment in FE. Will check and address it in future PR.

wbingli:
Looks build failed, style issue.

gongmax:
Fix the style issue

wbingli (approved):
LGTM

PR #332, opened 2018-01-16
Merge all step definitions into one file

xjhe:
#### Summary
1. Merge all step definitions into one file;
2. Remove some redundant codes;

#### Testing
./gradlew clean build 
 ./gradlew check
 ./gradlew cucumberTest  <--- This test still fails, but I added one line to check if the return value of the data service api is null, so we can take more time to see if the failure comes in the api or in the test itself

New tests cover the changes: 

#### Licensing


wjbuys (approved):
Yep, I think having one file for all the dataservice steps makes it a lot clearer.

PR #333, opened 2018-01-16
Fix cucumber initialization failure when using non-default region/profile

wjbuys:
This passes the AWS region/profile to use from the build logic into the cucumber integration tests through the default AWS system properties.

Before, the integration tests would use the default region/profile regardless of what the build logic configured. If the default profile was invalid, then the `cucumberTest` task would fail silently in spring context initialization where it tries to construct a Lambda client.

This should fix that.

#### Testing
Removed default profile from AWS config, and configured build to use a non-default profile. Before the change, `gradle integ-tests:cucumberTest` would fail before any tests were executed.

After this change, it actually fails because the cucumber tests themselves fail. I'm intentionally not fixing those in this PR, since @xjhe is already working on that.


xjhe (approved):
Good to go!

PR #334, opened 2018-01-16
Add implementation for frontend StartDeployment API and integration test

wbingli:
#### Summary
Add implementation for StartDeployment API.  

Aadd unit test for mapper, so we can focus our API unit test only on API.

I added integration test for FE component, this integration test will exercise all the FE code starting from lambda call. 

#### Testing
`./gradlew build`

- Add "compileOnly "org.mapstruct:mapstruct-processor" in the gradle to support IDE, so we can run the unit test for mapper in IDE. The IntelliJ doesn't support gradle `apt` well, it would not include the apt in the class path. 

#### Licensing


frontend-service/src/test/java/com/amazonaws/blox/frontend/integration/StartDeploymentIntegrationTest.java
```diff
@@ -0,0 +1,59 @@
...
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+
+public class StartDeploymentIntegrationTest extends IntegrationTestBase {
+
+  private static final String DEPLOYMENT_ID = "deployment-id";
+  private static final String ENVIRONMENT_REVISION_ID = "1";
+
+  @Test
+  public void testStartDeployment() throws Exception {
```
xjhe, line 35: Do we need to test any failure cases as well? Like when the response fails.

xjhe (approved):
Overall this looks good to me

frontend-service/src/test/java/com/amazonaws/blox/frontend/integration/StartDeploymentIntegrationTest.java
wbingli: yes, definitely, like one exception for ResourceNotFoundException.

wbingli: I added one failure case. However, we don't handle exception properly.  It's a big topic as we need to probably do Spring error handling or AWS lambda error handling, and may related with API and client SDK change.

Therefore, add a stub test here and TODO to update this test once we handle error properly.

wjbuys: Yeah, there's still an open task for this in #280, which we should break out for post M1.

frontend-service/build.gradle
```diff
@@ -36,6 +36,8 @@ dependencies {
     compile "io.swagger:swagger-annotations:${swaggerVersion}"
 
     compile "org.mapstruct:mapstruct-jdk8"
+
+    compileOnly "org.mapstruct:mapstruct-processor"
```
wjbuys, line 40: Hm, is this needed along with the `apt` declaration on L41?

frontend-service/src/test/java/com/amazonaws/blox/frontend/mappers/StartDeploymentMapperTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+public class StartDeploymentMapperTest {
+
+  private static final String ACCOUNT_ID = "123456789012";
+  private static final String CLUSTER = "mycluster";
+  private static final String ENVIRONMENT_NAME = "mynev";
+  private static final String ENVIRONMENT_REVISION_ID = "1";
+  private static final String DEPLOYMENT_ID = "deployment-id";
+
+  private final StartDeploymentMapper mapper = Mappers.getMapper(StartDeploymentMapper.class);
```
wjbuys, line 36: I wonder if it might be a bit more straightforward to test the mapper implicitly from the StartDeploymentTest, and to then avoid mocking out the mapper there. Thoughts?

frontend-service/src/test/java/com/amazonaws/blox/frontend/operations/StartDeploymentTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+  @Before
+  public void setUp() {
+    when(request.getAttribute(AwsProxyHttpServletRequestReader.API_GATEWAY_CONTEXT_PROPERTY))
+        .thenReturn(context);
+  }
+
+  @Test
+  public void testStartDeployment() throws Exception {
+    // Given
```
wjbuys, line 58: I think the comments can be omitted, or otherwise given a textual description similar to what we'd put in the cucumber test. If you're following the example from the existing e2e junit tests, that was mostly done to illustrate how those tests could be mapped to cucumber, not as a general pattern.

frontend-service/src/test/java/com/amazonaws/blox/frontend/operations/StartDeploymentTest.java
```diff
@@ -0,0 +1,80 @@
...
+  @Mock private HttpServletRequest request;
+  @Mock private DataService dataService;
+  @Mock private StartDeploymentMapper mapper;
+  @Mock private ApiGatewayRequestContext context;
+
+  @InjectMocks private StartDeployment api;
+
+  @Before
+  public void setUp() {
+    when(request.getAttribute(AwsProxyHttpServletRequestReader.API_GATEWAY_CONTEXT_PROPERTY))
```
wjbuys, line 52: There's an `EnvironmentControllerTest` base class that takes care of a lot of these concerns. Any reason to avoid using that?

frontend-service/src/test/java/com/amazonaws/blox/frontend/integration/StartDeploymentIntegrationTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+import com.amazonaws.blox.dataservicemodel.v1.exception.ResourceNotFoundException;
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.StartDeploymentResponse;
+import com.amazonaws.blox.frontend.operations.StartDeployment;
+import com.amazonaws.serverless.proxy.internal.model.AwsProxyRequest;
+import com.amazonaws.serverless.proxy.internal.model.AwsProxyResponse;
+import com.amazonaws.serverless.proxy.internal.testutils.AwsProxyRequestBuilder;
+import org.junit.Test;
+
+public class StartDeploymentIntegrationTest extends IntegrationTestBase {
```
wjbuys, line 29: Looking at the amount of coupling we have to do to our web framework's implementation details here, I don't know if this level of testing for each operation is worthwhile.

It's definitely useful to have a fast feedback test that makes sure our spring config is sane, and that all the web framework stuff is wired up correctly in terms of error handling, etc. But if we have to use raw URLs and manual marshaling/unmarshaling for *every* feature we're testing, I'm not sure if the added coupling is worth it.

Do you think it might make sense to keep one fairly generic test here that exercises our routing/error mapping through the spring MVC stuff, and then use the existing testing approach in `operations.*Test` to test actual functionality?

What I've done in other MVC frameworks, is to extract most of the functionality into a controller base class, and then to create a TestController subclass that we can control the behaviour of (e.g. a URL that always throws exceptions, a URL that always redirects, etc) to test these cross-controller concerns. 

Thoughts? Also happy to discuss this in person, if that's more helpful.

wjbuys (commented):
I think this is shippable (with some optional minor fixes), but I have some questions around the integration testing approach here that I think we should discuss before adding more integration tests that follow the same pattern.

frontend-service/build.gradle
wbingli: See my comment in the PR summary, basically it's  "Add "compileOnly "org.mapstruct:mapstruct-processor" in the gradle to support IDE, so we can run the unit test for mapper in IDE. The IntelliJ doesn't support gradle apt well, it would not include the apt in the class path."

wjbuys: Oh! I totally missed that, sorry.

If `compileOnly` works, do we actually gain anything by using the `apt` plugin? Should we just switch everything back to `compileOnly`?

frontend-service/src/test/java/com/amazonaws/blox/frontend/integration/StartDeploymentIntegrationTest.java
wbingli: My initial idea is to use Spring MVC integration test framework, see (https://docs.spring.io/spring-security/site/docs/current/reference/html/test-mockmvc.html)

We could use Spring MVC integration test to test our routing and error handling. But I don't think it will exercise the whole component enough and any effort we save.  The FE routing is not complex as a website, we mostly handle the request just like a Coral API. There is no redirects etc.  The integration test is actual one-one mapping to test each API call. 

If the effort to test the routing is the same, why we don't exercise the whole component from the entry point, lambda handler?  It will not only exercise the Spring mvc routing and also make sure the wiring between lambda to Spring works fine. 

The "raw URLs and manual marshaling/unmarshaling for every feature" is somehow annoying. But I see it's actual the input and output of this component, lambda input and output.   The URL are actually good to use here, as it's how we expose the API. Even we use Spring MVC test, we need to specify the same URL as well.

frontend-service/src/test/java/com/amazonaws/blox/frontend/mappers/StartDeploymentMapperTest.java
wbingli: I think it's what we want to discuss f-2-f.  Just summarize what I think about them:

### Unit Test Practice 1:  API Unit test with real Mapper 
Example see (https://github.com/blox/blox/blob/dev/frontend-service/src/test/java/com/amazonaws/blox/frontend/operations/CreateEnvironmentTest.java).  It will use a real mapper inside API unit test. 

I can see this unit test can gain more confidence it pass the real object and get some concrete result to check.  However, the downside I see are:
 - Not a pure unit test for API test only. It actually depends on the implementation of mapper. If any mapper model change or rule change, it most likely we need to fix the API test even no change on API class. It validates one of the rule that one bug should be only be caught in one test.
- No Mapper unit test:  Most likely we think we exercise the mapper in API test, but API unit test is not focus on the mapper test. Without mapper unit test, we cannot make sure we have cover all the cases in the mapper. 

### Unit Test Practice 2:  Unit test for mapper and Mock mapper/object in API Unit test
Example see:
 - Mapper unit test: https://github.com/blox/blox/blob/dev/frontend-service/src/test/java/com/amazonaws/blox/frontend/mappers/StartDeploymentMapperTest.java
 - API unit test: https://github.com/blox/blox/blob/dev/frontend-service/src/test/java/com/amazonaws/blox/frontend/operations/StartDeploymentTest.java

The good part is testing each class separately without any dependency.  The argument is always goes to:
 - Mapper consider as an implementation detail in API. If any refactoring on mapper, say change API name from one to other.  We don't need to update the test, and it should expect the same result.  (The practice 1 may argue that how would you make sure you call the right mapper API, what if you make a call to the wrong API but just got right result )

It's kind of the argument on __Testing Behavior__  vs. __Testing Implementation__.   

The first test practice is more on testing behavior, the test says "I don't care about what is mapper,  with this input and I expect this result."

The second test practice is on testing implementation, the test says "This class calls this API of mapper, and then calls data service with the result from mapper."    

It's may a personal taste thing, but I'm open to have a discussion.

frontend-service/src/test/java/com/amazonaws/blox/frontend/operations/StartDeploymentTest.java
wbingli: I see them as different test. This class is more as a simple unit test with mocks. But `EnvironmentControllerTest` kind of like small integration test with Spring test runner.So want to have a different solution here to show an example for discussion.

frontend-service/src/test/java/com/amazonaws/blox/frontend/operations/StartDeploymentTest.java
wbingli: It's kind of my personal way if writing a long unit test or any other test.  The `Given-When-Then` is nice to break  the test code into three sections and easy to read the test.

I'm happy to remove it if it's annoying.

wjbuys: Given-when-then is perfectly fine (I also like breaking up my tests these way, usually just with white space), just wanted to make sure it's your choice and not a copy of a pattern that was kind of throwaway.

frontend-service/build.gradle
wbingli: We still need `apt` for `gradle` to process the annotation. The IntelliJ has different configuration for annotation processing. It's more like the  gradle + IntelliJ integration doesn't support `apt`  so far.

The IDE support for `apt` plugin is here: https://github.com/tbroyer/gradle-apt-plugin#intellij-idea   I haven't went though all of them, but `compileOnly` is the way from the sample config of this plugin.

wjbuys: Yeah, I've got that too, but it's hit or miss whether the IDE actually applies it.

I did try just getting rid of the plugin completely in favor of just using `compileOnly/testCompileOnly` instead of `apt`, and that seemed to work... Let's come back to this later.

frontend-service/src/test/java/com/amazonaws/blox/frontend/integration/StartDeploymentIntegrationTest.java
wjbuys: So I went down the MockMVC rabbit hole, and I couldn't get it to work with the aws-serverless-java-container code, because it doesn't actually stand up a full servlet stack from just our Application Config (it [manually wires some spring stuff in the handler](https://github.com/awslabs/aws-serverless-java-container/blob/master/aws-serverless-java-container-spring/src/main/java/com/amazonaws/serverless/proxy/spring/SpringLambdaContainerHandler.java#L52-L63)). But if we could get that working, it does sound like it would be the best of both worlds.

wbingli: You mean you want to start from our` LambdaHandler` class?   The integration test actually write the Lambda container handler as here : https://github.com/blox/blox/blob/eb90b05d93f250dca8de3c2074c47cb52ff514f3/frontend-service/src/test/java/com/amazonaws/blox/frontend/integration/IntegrationTestConfig.java#L46-L47

We cannot provide a configuration Class (a test one) to start because LambadaContainerHandler will initialize the Spring app context but not provide a way to get it out, so we cannot wire our test into the context.

Issue #315, opened 2018-01-09
Exception mapping error between JsonRPCServer and client


narehayrapetyan:
Exception mapping does not work correctly. The return payload on the JsonRPCLambdaServer side returns:
`
{"jsonrpc":"2.0","id":"1380430143","error":{"code":-32001,"message":"environment with id EnvironmentId(environmentName=testfd33348b-b0d2-4c9d-863b-c29792719376, accountId=12345678912, cluster=clustere57c7631-8441-4b27-990a-afd4bfef8b19) already exists","data":{"exceptionTypeName":"com.amazonaws.blox.dataservicemodel.v1.exception.ResourceExistsException","message":"environment with id EnvironmentId(environmentName=testfd33348b-b0d2-4c9d-863b-c29792719376, accountId=12345678912, cluster=clustere57c7631-8441-4b27-990a-afd4bfef8b19) already exists"}}}`

but the client interprets it as:

```
java.lang.NullPointerException
        at com.googlecode.jsonrpc4j.JsonRpcClient.handleErrorResponse(JsonRpcClient.java:270)
        at com.googlecode.jsonrpc4j.JsonRpcClient.readResponse(JsonRpcClient.java:196)
        at com.googlecode.jsonrpc4j.JsonRpcClient.readResponse(JsonRpcClient.java:529)
        at com.amazonaws.blox.jsonrpc.JsonRpcLambdaClient.readResponse(JsonRpcLambdaClient.java:150)
        at com.amazonaws.blox.jsonrpc.JsonRpcLambdaClient.lambda$invokeAsync$2(JsonRpcLambdaClient.java:112)
        at java.util.concurrent.CompletableFuture.uniApply(CompletableFuture.java:602)
        at java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:577)
        at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:474)
        at java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:1962)
        at software.amazon.awssdk.http.pipeline.stages.AsyncRetryableStage$RetryExecutor.handle(AsyncRetryableStage.java:152)
  
```

wjbuys:
It looks like the default exception mapper in jsonrpc4j only supports exceptions with no-argument or message-argument constructors:

[...]

PR #335, opened 2018-01-17
Use a common prefix to namespace resources created in a single scenario

wjbuys:
This changes the InputCreator to prefix the same UUID to all resources
it creates, and sets up the DataServiceSteps as a DirtiesContext. The
effect of this is that all resources that are created in the same
scenario run will have the same prefix, and that prefix will be
different from all other scenario runs.

This lets us run the tests more than once, without needing to clean up
data.

#### Testing
Apart from `gradle check`, the `gradle integ-tests:cucumberTest` target passes except for the following:

```
Feature: Create environment

  Scenario: Create an environment                  # /Users/buys/workspaces/blox/blox/integ-tests/src/cucumberTest/resources/features/dataservice/create_environment.feature:5
    When I create an environment                   # DataServiceSteps.java:47
    Then the created environment response is valid # DataServiceSteps.java:53

  Scenario: Create an environment that already exists                                          # /Users/buys/workspaces/blox/blox/integ-tests/src/cucumberTest/resources/features/dataservice/create_environment.feature:9
    Given I create an environment named "test" in the cluster "testCluster"                    # DataServiceSteps.java:98
    When I try to create another environment with the name "test" in the cluster "testCluster" # DataServiceSteps.java:105
    Then there should be a ResourceExistsException thrown                                      # DataServiceSteps.java:188
      org.junit.ComparisonFailure: expected:<R[esourceExists]Exception> but was:<R[untime]Exception>
        at org.junit.Assert.assertEquals(Assert.java:115)
        at org.junit.Assert.assertEquals(Assert.java:144)
        at cucumber.steps.dataservice.DataServiceSteps.lambda$new$13(DataServiceSteps.java:192)
        at ✽.Then there should be a ResourceExistsException thrown(/Users/buys/workspaces/blox/blox/integ-tests/src/cucumberTest/resources/features/dataservice/create_environment.feature:12)

    And the resourceType is "environment"                                                      # DataServiceSteps.java:195
    And the resourceId contains "test"                                                         # DataServiceSteps.java:214
```

This still fails because of #315 (exceptions from JSON-RPC aren't mapped correctly), and can be tackled as part of that issue.


integ-tests/src/cucumberTest/java/cucumber/steps/helpers/InputCreator.java
```diff
@@ -20,95 +20,100 @@
...
-        .taskDefinition(TASK_DEFINITION_ARN)
+        .environmentId(id)
+        .role(getRoleArn())
+        .taskDefinition(getTaskDefinitionArn())
         .environmentType(EnvironmentType.Daemon)
         .deploymentMethod("ReplaceAfterTerminate")
         .build();
   }
 
+  private EnvironmentId environmentId(final String environmentName, final String cluster) {
```
xjhe, line 79: It's clearer to name this method as "createEnvironmentId" given that you're generating a new ID with the name and cluster?

integ-tests/src/cucumberTest/java/cucumber/steps/dataservice/DataServiceSteps.java
```diff
@@ -143,7 +145,7 @@ public DataServiceSteps() {
           final EnvironmentId environmentId = getEnvironmentIdFromCreatedEnvironment();
           dataServiceWrapper.updateEnvironment(
               inputCreator.updateEnvironmentRequestWithNewCluster(
-                  environmentId.getEnvironmentName(), newCluster));
+                  environmentId.getEnvironmentName(), inputCreator.prefixName(newCluster)));
```
xjhe, line 148: This is not in the scope of this PR, but should we change this update into "I update the created environment with new task definition" ? For the purpose of testing, updating the cluster name should be fine. But I'm not sure in reality will there be a use case of updating the cluster.

integ-tests/src/cucumberTest/java/cucumber/steps/helpers/InputCreator.java
```diff
@@ -20,95 +20,100 @@
 import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.CreateEnvironmentRequest;
 import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.DeleteEnvironmentRequest;
 import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.DescribeEnvironmentRequest;
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.ListClustersRequest;
 import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.UpdateEnvironmentRequest;
-
+import java.util.Collections;
+import java.util.StringJoiner;
 import java.util.UUID;
+import lombok.Getter;
```
xjhe, line 27: Are we using Getter/Setter here?

integ-tests/src/cucumberTest/java/cucumber/steps/helpers/InputCreator.java
```diff
@@ -20,95 +20,100 @@
...
-  private static final int ACCOUNT_ID_SIZE = 12;
-  private static final String ACCOUNT_ID = generateAccountId();
-  private static final String TASK_DEFINITION_ARN =
-      "arn:aws:ecs:us-east-1:" + ACCOUNT_ID + ":task-definition/sleep";
-  private static final String ROLE_ARN = "arn:aws:iam::" + ACCOUNT_ID + ":role/testRole";
-  private static final String CLUSTER_NAME_PREFIX = "cluster";
+  private static final String DEFAULT_ACCOUNT_ID = "123456789012";
+  private static final String DEFAULT_CLUSTER_NAME = "Cluster";
+  private static final String DEFAULT_ENVIRONMENT_NAME = "Environment";
+  private final String sharedPrefix = "blox-integ-tests";
```
xjhe: Should we make sharedPrefix static as well?

integ-tests/src/cucumberTest/java/cucumber/steps/helpers/InputCreator.java
wjbuys: Should be on the `accountId` field, yep. Didn't want to apply it to the whole class.

integ-tests/src/cucumberTest/java/cucumber/steps/dataservice/DataServiceSteps.java
wjbuys: Definitely, in a followup PR. Updating the cluster name actually makes no sense; we shouldn't support moving Environments between Clusters, or Weird Things May Happen.

integ-tests/src/cucumberTest/java/cucumber/steps/helpers/InputCreator.java
wjbuys: Whoops, yes, will fix.

integ-tests/src/cucumberTest/java/cucumber/steps/helpers/InputCreator.java
wjbuys: The other methods on this class are named for the type they create (compare `DescribeEnvironmentRequest describeEnvironmentRequest`), so I thought this naming would make sense in the context of this class; with an appropriately named variable, all invocations look something like:

```
create.describeEnvironmentRequest(...)
create.environmentId(...)
```

as opposed to:

```
creator.createDescribeEnvironmentRequest()
creator.createEnvironmentId()
```

Thoughts?

xjhe: Yes if we follow the pattern of the other methods, I think naming this as environmentId is fine.

wbingli: I would prefer shorter naming,  `create.describeEnvironmentRequest`  looks good

wbingli (approved):
LGTM

PR #336, opened 2018-01-17
Implement ListEnvironments FE API, implement ListEnvironments with en…

gongmax:
#### Summary
 Implement ListEnvironments FE API, implement ListEnvironments with environmentNamePrefix. Will add FE integ-test in future PR

#### Testing
./gradlew build

#### Licensing


data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepository.java
```diff
@@ -33,7 +34,8 @@ Environment createEnvironmentAndEnvironmentRevision(
   Environment getEnvironment(EnvironmentId environmentId)
       throws ResourceNotFoundException, InternalServiceException;
 
-  List<Environment> listEnvironments(Cluster cluster) throws InternalServiceException;
+  List<Environment> listEnvironments(Cluster cluster, Optional<String> environmentNamePrefix)
```
wjbuys: Every time I try to do this, I keep getting IDE/compile warnings that `Optional<>` in method parameters is not recommended, and [the internets seem to agree](https://stackoverflow.com/questions/31922866/why-should-java-8s-optional-not-be-used-in-arguments).

I'm not sure I agree, but does this actually buy us less complexity over a normal null-check for the prefix parameter?

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
```diff
@@ -148,13 +155,49 @@ public void testGetEnvironmentRevisionInternalError() throws Exception {
...
+
+    verify(dynamoDBMapper)
+        .query(eq(EnvironmentDDBRecord.class), ddbQueryExpressionCaptor.capture());
+    verify(environmentMapper).toEnvironment(environmentDDBRecord);
+
+    final EnvironmentDDBRecord queriedEnvironmentDDBRecord =
+        (EnvironmentDDBRecord) ddbQueryExpressionCaptor.getValue().getHashKeyValues();
+    final Map queryConditions = ddbQueryExpressionCaptor.getValue().getRangeKeyConditions();
+
+    assertThat(
```
wjbuys: How do you feel about moving to AssertJ for all test assertions? It seems like it would simplify some of these assertions quite a bit, especially the Map comparisons.

frontend-service/src/main/java/com/amazonaws/blox/frontend/operations/ListEnvironments.java
```diff
@@ -30,22 +36,29 @@
...
   @ApiOperation(value = "List all environments")
   public ListEnvironmentsResponse listEnvironments(
       @PathVariable("cluster") String cluster,
       @RequestParam(value = "environmentNamePrefix", required = false) String environmentNamePrefix,
-      @ModelAttribute RequestPagination pagination) {
+      @ModelAttribute RequestPagination pagination)
+      throws ServiceException, ClientException {
 
-    return ListEnvironmentsResponse.builder().environmentNames(Collections.emptyList()).build();
+    // TODO: Add validation for cluster. Add Validation for environmentNamePrefix
```
wjbuys, line 49: Do we have an issue to track adding external resource validation?

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepository.java
wbingli: The actual interesting code is following `@NonNull final Optional<String> environmentNamePrefix)`. You need to make sure the Optional is not null, :).    I would prefer to pass a nullable value to the method. And inside the method. you can use Optional.ofNullable to handle two different cases.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
```diff
@@ -308,15 +312,23 @@ public Environment getEnvironment(@NonNull final EnvironmentId environmentId)
   }
 
   @Override
-  public List<Environment> listEnvironments(@NonNull final Cluster cluster)
+  public List<Environment> listEnvironments(
```
wbingli: You may keep the original method to have the helper method to list all Enviornment without prefix, such as 
```
public List<Environment> listEnvironments(@NonNull final Cluster cluster) {
  this(....);
}
```

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepository.java
gongmax: No, it doesn't. I'm just used to use Optional to avoid null input to the method. Since you call this out, I will remove it.

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
gongmax: Sure. Done!

frontend-service/src/main/java/com/amazonaws/blox/frontend/operations/ListEnvironments.java
gongmax: I've created one.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepository.java
gongmax: @wbingli well, the intention of use Optional is not checking whether the Optional is null or not. So if the Optional itself is null, a NPE is thrown. But I removed it anyway since you have concerns.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
gongmax: Yeah, good catch!

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDBTest.java
```diff
@@ -164,29 +161,28 @@ public void testListEnvironmentsWithEnvironmentNamePrefix() throws Exception {
...
-    assertThat(result.size(), is(1));
-    assertThat(result.get(0), is(environment));
+    final Map<String, Condition> queryConditions =
+        ddbQueryExpressionCaptor.getValue().getRangeKeyConditions();
+
+    assertThat(queriedEnvironmentDDBRecord.getAccountIdCluster())
+        .isEqualTo(cluster.generateAccountIdCluster());
+    assertThat(queryConditions).isNotEmpty().hasSize(1);
+    assertThat(queryConditions).containsKey(ENVIRONMENT_NAME_RANGE_KEY);
+    assertThat(queryConditions)
```
wjbuys, line 179: Protip: For comparing maps, you can use `assertThat(queryConditions).contains(entry(...key..., ...value...)`

wjbuys (approved):

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepository.java
wjbuys: Yeah, to be clear, using Optional in return types is totally okay, it's just the usage as method parameters (especially when the type itself is nullable) that's counter-indicated.

The exception to this is that certain service libraries will treat `Optional` parameters specially (e.g. the Swagger generator will make those `required=false` in the frontend model), so for those cases it may be useful.

PR #337, opened 2018-01-17
Add delete environment DataService API

xjhe:
#### Summary
This PR implements the DeleteEnvironemntApi for data service. 
As agreed with the team, this Api currently does the job of a reaper, which deletes the actual record of Environment/EnvironmentRevision from the DDB.

#### Testing
./gradlew check
./gradlew clean build

New tests cover the changes: 

#### Licensing

### DIFF
--- a/data-service/src/main/java/com/amazonaws/blox/dataservice/api/DeleteEnvironmentApi.java
+++ b/data-service/src/main/java/com/amazonaws/blox/dataservice/api/DeleteEnvironmentApi.java
@@ -42,6 +42,9 @@ public class DeleteEnvironmentApi {
     final EnvironmentId environmentId =
         apiModelMapper.toModelEnvironmentId(request.getEnvironmentId());
     try {
+      // Get the environment to check it exists
+      final Environment environment = environmentRepository.getEnvironment(environmentId);
+
       // List and delete all environment revisions
       final List<EnvironmentRevision> environmentRevisions =
           environmentRepository.listEnvironmentRevisions(environmentId);
@@ -51,8 +54,6 @@ public class DeleteEnvironmentApi {
         environmentRepository.deleteEnvironmentRevision(revision);
       }
 
-      // Get the environment to check it exist
-      final Environment environment = environmentRepository.getEnvironment(environmentId);
       // Delete the environment
       environmentRepository.deleteEnvironment(environmentId);
       return DeleteEnvironmentResponse.builder()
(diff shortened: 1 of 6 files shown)


data-service/src/main/java/com/amazonaws/blox/dataservice/api/DeleteEnvironmentApi.java
```diff
@@ -0,0 +1,70 @@
...
+      final List<EnvironmentRevision> environmentRevisions =
+          environmentRepository.listEnvironmentRevisions(environmentId);
+      // TODO: Need to stop all the running tasks from the scheduler manager
+
+      for (final EnvironmentRevision revision : environmentRevisions) {
+        environmentRepository.deleteEnvironmentRevision(revision);
+      }
+
+      // Get the environment to check it exist
+      final Environment environment = environmentRepository.getEnvironment(environmentId);
```
wbingli: How about move this API call to the beginning? If the environment doesn't exist, we don't need to call the `listEnviromentRevisions`.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
```diff
@@ -366,4 +366,21 @@ public void deleteEnvironmentRevision(@NonNull final EnvironmentRevision environ
...
   }
+
+  @Override
+  public void deleteEnvironment(@NonNull final EnvironmentId environmentId)
+      throws InternalServiceException {
+    try {
+      dynamoDBMapper.delete(
+          EnvironmentDDBRecord.withKeys(
+              environmentId.generateAccountIdCluster(), environmentId.getEnvironmentName()),
+          SaveBehavior.CLOBBER.config());
```
wbingli: Why need a `SaveBehavior.CLOBBER.config()` instead of the default config?  The SaveBehavior probably be only useful for save and update?

data-service/src/main/java/com/amazonaws/blox/dataservice/api/DeleteEnvironmentApi.java
xjhe: Good point! Let me do that

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
xjhe: I was thinking about using the default one, but I see that the "deleteEnvironmentRevision" also uses this SaveBahavior. I can just change both back to the default

xjhe: After reading about the SaveBehavior I think it should be used for deleteEnvironmentRevision but not deleteEnvironment. The SaveBehavior.CLOBBER.config() is to delete when the version is out of date. We don't need this feature for deleteEnvironment I think.

wjbuys (approved):

wbingli (approved):

PR #338, opened 2018-01-17
Add buildspec file to build on AWS CodeBuild

wbingli:
#### Summary
The build spec for code build. It will allow people to use CodeBuild to build the project
 
There should be more artifacts we should publish for deployment, but I would like to leave it for the next step.

#### Testing
See https://github.com/wbingli/blox/commits/aws_code_build,  it passed in my code build project.

#### Licensing


buildspec.yml
```diff
@@ -0,0 +1,17 @@
+version: 0.2
+
+phases:
+  build:
+    commands:
+      - ./gradlew build packageLambda
```
wjbuys: The canonical target for verifying the build is `gradle check`, which also does static analysis.

buildspec.yml
```diff
@@ -0,0 +1,17 @@
...
+
+artifacts:
+  files:
+    - 'frontend-service/build/distributions/frontend-service.zip'
```
wjbuys: We don't actually vend/use these anywhere, don't think we need to actually archive them.

buildspec.yml
```diff
@@ -0,0 +1,17 @@
...
+    - 'data-service/build/distributions/data-service.zip'
+    - 'scheduling-manager/build/distributions/scheduling-manager.zip'
+cache:
+  paths:
+    - '/root/.gradle/caches/**/*'
```
wjbuys: Ah neat, I was wondering how this would work. Just watch out, some of these files aren't cacheable, and must be deleted before caching. See https://docs.travis-ci.com/user/languages/java/#Caching for details.

wbingli: What I observed for codebuild caching, it basically zip all the files declared and upload to S3.  And download the S3 object and unzip for next build.

buildspec.yml
wbingli: I can remove it for now. I initially idea is we have S3 bucket as our artifact repo and used for further deployment, so we don't have to re-build and reload to different S3 bucket.

buildspec.yml
wbingli: should the `build` is the full build, including `check` task?

buildspec.yml
wjbuys: Right, so the problem is that gradle leaves a lockfile in there, and that could interfere with subsequent gradle runs. We work around this in Travis with this snippet:

https://github.com/blox/blox/blob/653e896c0a056ab2a97210a4a091a1c2a7ee6bc2/.travis.yml#L7-L11

We should do something similar here.

wbingli: I think the intention to remove the lock file and other stuff is not for cleaning the cache for subsequent gradle. As far as I see, without removing those files, build with previous cache file work fine.

The actual propose of removing them is to avoid re-uploading the cache, see (https://github.com/travis-ci/docs-travis-ci-com/issues/358).   Because those lock file will be changed for every build, the build see the change and have to re-upload the whole cache even there is no actual gradle dependency change.

CodeBuild doesn't provide such `before_cache` phase, but I can try it in the `post-build` phase to see if CodeBuild actual compare the cache to avoid re-uploading.

wbingli:
## Update for travis configuration

While trying to fix the cache issue for CodeBuild, I found that our Travis confguration for cache is actually broken, see any previous build (e.g. https://travis-ci.org/blox/blox/builds/330022734). Files insides `fileHashes` and `jars-3` are changed even no code change. (Try to re-build the same build).

I think `jars-3` is used by gradle for custom build task, so it actually rebuild out `buildSrc`  every time. Once removing those filed, the cache in Travis now is doing  right job. See https://travis-ci.org/blox/blox/builds/330226041

I also deleted all the Travis caches.  The old caches is about ~500M, because it includes other version of gradle (4.0.3). I think we use this version before, therefore, it comes with the cache all the way to now.  Once deleted all the caches, now the cache is about ~280M.

wbingli:
## Update for the build spec for CodeBuild:

- Disable gradle daemon:  Same best practice from Traivs,  the build is running in container and gradle daemon will never be used again.
- Cache is still __BROKEN__:  I applied the same configuration as Travis. Ideally it should work as Travis. However, it just re-upload every time.  I doubt CodeBuild will validate whether to upload comparing the cache file changes. Also CodeBuild doesn't give much info about which files are changed in cache for re-uploading. Travis does a good job here so we can easily fix it.
- The build task is `./gradlew build`. I think it is actually better than Travis' default way.  Gradle `build` task = `assemble` + `check`. I see Travis actually separate it as two commands.  It will bring performance issue as it needs to start two JVM process, gradle is really slow, groovy.. hmm..

buildspec.yml
wjbuys: Ah, I see, I misunderstood. Thanks for clarifying.

wjbuys (approved):
Nicely done 👍

buildspec.yml
```diff
@@ -0,0 +1,22 @@
+version: 0.2
+
+phases:
+  pre_build:
+    commands:
+      - mkdir -p /root/.gradle
+      - echo "org.gradle.daemon=false" > /root/.gradle/gradle.properties
+  build:
+    commands:
+      - ./gradlew build
```
xjhe, line 10: When I run the build from my command line, I usually run ./gradlew clean build to make sure it didn't show me wrong error messages. Should we also change this into clean build?

wbingli: We don't need the `clean`. The CodeBuild or Travis will always start a build from a fresh container.

xjhe: Good to know that! Then I think this is fine. 
It's better to let others check this PR as well, given that I'm not that familiar with CodeBuild.

wbingli: yep, Jacob has reviewed this PR.

PR #339, opened 2018-01-18
Fix errors in describe environment cucumber tests

xjhe:
#### Summary
Fix some errors in DescribeEnvironment cucumber test, so it accepts the correct environmentId. 

The test still fails for the exception messages (because of the JSON RPC problem), and for the case that requires an update environment, given that we haven't implemented update yet. 

#### Testing
./gradlew clean build
./gradlew cucumberTest

New tests cover the changes: 

#### Licensing


wbingli (approved):
LGTM

PR #340, opened 2018-01-18
Add DataService.listClusters() API

wjbuys:
This is a first-pass implementation of the listClusters API that lists the clusters tied to environments that Blox knows about.

### Known issues to be fixed in the next PR:

- We're using a DDB scan instead of querying the index.
  Unfortunately, the current index doesn't work, because the attributes
  it indexes on were removed. This is a big enough chunk of work that it
  needs a separate PR
- No unit tests for the EnvironmentRepositoryDDB.listClusters
  These would have to be thrown away immediately when adding the index
  above, so leaving them out and relying on the integ tests for now.


data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
```diff
@@ -383,4 +389,31 @@ public void deleteEnvironmentRevision(@NonNull final EnvironmentRevision environ
           e);
     }
   }
+
+  @Override
+  public List<Cluster> listClusters(String accountId, String clusterNamePrefix) {
+    if (clusterNamePrefix != null) {
+      Validate.notNull(accountId, "accountId must be specified if clusterNamePrefix is specified");
```
wbingli, line 396: Not sure if we want to enforce an accountId if clusterNamePrefix is provided.  The `listCluster` is mostly used by reconciler to schedule work per cluster.  With only clusterNamePrefix, it's a way to partition the clusters and potential run multiple reconciler with different assigned name prefix.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
```diff
@@ -383,4 +389,31 @@ public void deleteEnvironmentRevision(@NonNull final EnvironmentRevision environ
...
+
+    Stream<Cluster> stream =
+        records
+            .stream()
+            .map(EnvironmentDDBRecord::getAccountIdCluster)
+            .distinct()
+            .map(Cluster::fromAccountIdCluster);
+
+    if (accountId != null) {
+      stream = stream.filter(c -> c.getAccountId().equals(accountId));
```
wbingli, line 411: How about chain with the stream?  
```
.map(..)
.filter(c -> accountId==null || c.getAccountId().equals(accountId)))
```

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
```diff
@@ -383,4 +389,31 @@ public void deleteEnvironmentRevision(@NonNull final EnvironmentRevision environ
...
+
+  @Override
+  public List<Cluster> listClusters(String accountId, String clusterNamePrefix) {
+    if (clusterNamePrefix != null) {
+      Validate.notNull(accountId, "accountId must be specified if clusterNamePrefix is specified");
+    }
+
+    // TODO: rewrite this to use the index instead of scanning the table
+    PaginatedScanList<EnvironmentDDBRecord> records =
+        dynamoDBMapper.scan(EnvironmentDDBRecord.class, new DynamoDBScanExpression());
```
xjhe, line 401: We're doing a full table scan to get all clusters. Do we know how slow this is in the extreme case?

wjbuys: See the review description; this is a short-term hack, working on fixing the data model/index now.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
wjbuys: I'm keeping this separate so that I can change it over to the DDB Query expression code quickly in the followup PR.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
wjbuys: This API is based on the assumption that the data model will have accountID as hash key and cluster name as range key. In that schema, DDB will not allow queries for only the range key without scanning.

I suspect that we'll have to partition along some completely synthetic axis (a consistent hash perhaps?) if we do partition here, so maybe even (accountId, clusterName) isn't the right key schema.

At least for tests, allowing this query pattern is pretty handy.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
xjhe: Makes sense....I should have read the description first :-D Waiting for your next PR

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
wbingli: y, make sense.

wbingli (approved):
Shipit

PR #342, opened 2018-01-19
Add Data Service integ test for ListEnvironments

gongmax:
#### Summary
Add Data Service integ test for ListEnvironments

#### Testing
./gradlew cucumberTest

#### Licensing


integ-tests/src/cucumberTest/resources/features/dataservice/list_environments.feature
```diff
@@ -1,26 +1,41 @@
-@ignore
 @dataservice
 @list-environments
 Feature: List environments
 
+  @ignore
```
xjhe, line 5: Do we need to add a "background" with the accountID 123456789012?

integ-tests/src/cucumberTest/resources/features/dataservice/list_environments.feature
```diff
@@ -1,26 +1,41 @@
...
   Scenario: List environments on a non-existent cluster
-    When I list environments on a cluster that does not exist
+    When I list environments in a cluster that does not exist
```
xjhe, line 7: We still need a step definition for this one

integ-tests/src/cucumberTest/resources/features/dataservice/list_environments.feature
```diff
@@ -1,26 +1,41 @@
...
     Then 0 environments are returned
```
xjhe, line 8: If we change this line into "there environment are returned" and with empty return list, we can skip implementing this step, and can reuse the "there environment are returned" step.

integ-tests/src/cucumberTest/resources/features/dataservice/list_environments.feature
gongmax: Actually 123456789012 is the default value in inputCreator. Add background has is kind of duplicated but can make the test easy to read. I'm leaning to not add it.

integ-tests/src/cucumberTest/resources/features/dataservice/list_environments.feature
gongmax: I think this scenario is testing listEnvrionments with a cluster that is not exist, in which case, the API will throw an exception. Since we haven't implement validation for the cluster, so I mark this scenario as ignore and will leave the detail implementation in future.

xjhe: Sure. Let's save this one for future implementations.

integ-tests/src/cucumberTest/resources/features/dataservice/list_environments.feature
xjhe: That's true -- the "background" is just for readability. I'm fine with this.

xjhe (approved):
All is well

PR #343, opened 2018-01-19
Fix froentend service lambda permission and async call permission issue

wbingli:
#### Summary
For the end-2-end test working,  this PR fix following problem to make FE working:
 - Grant FE lambda permission to call data service lambda
 - The SDK async client issue inside lambda
 - Timeout is too short for FE to start, which will cause all request failure. It looks lambda will only cache the handler object if it handle the request successful.

#### Implementation details
There is one issue of template processing, basically you cannot use `!GetAtt DataServiceHandler.Arn` but have to be `Fn::GetAtt: [DataServiceHandler, Arn]`.  It looks there is some issue inside cloudformation process `PostProcessCloudformation`, it doesn't handle the `!GetAtt` well.  I haven't really spent time on it, I doubt  the exclamation is special case during json/template processing.

#### Testing
Run the E2E test to call the API successful now.  The E2E test will be other PR.

```
282f7b3 2018-01-19 05:56:51 8b4d2484-fcdd-11e7-add1-01463dd32d62 INFO  LambdaHandler:37 - Handling request: 8b4cd61e-fcdd-11e7-809e-f3f5162a1eab
/aws/lambda/blox-libruce-alpha-us-west-2-FrontendHandler-Z0HV9IB54DTL 2018/01/19/[$LATEST]1ac7a03a02cd4813bca5d246b282f7b3 2018-01-19 05:56:51 8b4d2484-fcdd-11e7-add1-01463dd32d62 INFO  LambdaHandler:41 - Completed request: 8b4cd61e-fcdd-11e7-809e-f3f5162a1eab, with response: 200
```

#### Licensing


wjbuys:
> There is one issue of template processing, basically you cannot use !GetAtt DataServiceHandler.Arn but have to be Fn::GetAtt: [DataServiceHandler, Arn]. It looks there is some issue inside cloudformation process PostProcessCloudformation, it doesn't handle the !GetAtt well. 

Ah, yeah, it's because the Java library we're using for parsing and postprocessing the YAML files unfortunately doesn't support the YAML directive/type tag syntax that makes the `!Ref` shorthands work.

wjbuys (approved):

PR #344, opened 2018-01-19
End2end test

wbingli:
#### Summary
This PR updates the end-2-end test for the first milestone goal.  This test can pass for all the API calls to make a deployment. However,  the final pieces of scheduler is not finished yet. Once it's done, this e2e  should work.

This PR depends on PR(#343)  to work togather.

#### Implementation details
- Use Junit for the integration test:  Not spend effort to setup cucumber, since our main focus is to make the e2e testing running and we most likely will move away from cucumber
- Update the frontend client by `gradlew updateClient`
- Remove the stage in  the FrontEnd API URL. The SDK generated actually include stage in the path.  (Will need more discussion this for the API/SDK versioning mgmt)

#### Testing
- `gradlew testEndToEnd` Passed now
Observed a new tasked is started and stopped after the test.

#### Licensing

#### Before merging
This PR depends on PR (https://github.com/blox/blox/pull/343), better to have PR(#343) merge first.


end-to-end-tests/src/main/java/com/amazonaws/blox/integ/BloxTestStack.java
```diff
@@ -29,20 +31,68 @@
...
+        ProfileCredentialsProvider.builder().profileName(this.credentialProfile).build();
+
+    this.cloudFormationClient =
+        CloudFormationClient.builder()
+            .credentialsProvider(profileCredentialsProvider)
+            .region(Region.of(this.region))
+            .build();
+
+    this.dynamoDBClient =
+        DynamoDBClient.builder()
```
wjbuys: I don't think we need this client anymore, since this PR also deletes the code that directly interacted with DDB to clear tables.

end-to-end-tests/build.gradle
```diff
@@ -46,6 +46,7 @@ task testEndToEnd(type: Test) {
     dependsOn ":scheduling-manager:deploy"
 
     doFirst {
+        systemProperty 'blox.tests.region', stack.region.toString()
         systemProperty 'blox.tests.awsProfile', stack.profile.toString()
```
wjbuys: For the integ tests, I actually started using the AWS SDK override properties instead of plumbing these through manually: 

https://github.com/blox/blox/blob/d4017f64a0be92a1e83db34956c9926c8d7ded43/integ-tests/build.gradle#L12-L21

That should let us just create default clients everywhere, which should simplify `BloxStack` somewhat. Do you think that pattern is worth adopting here?

wjbuys (approved):
LGTM, with some questions/suggestions

end-to-end-tests/build.gradle
wbingli: That looks good. Let me adopt this approach.

end-to-end-tests/src/main/java/com/amazonaws/blox/integ/BloxTestStack.java
wbingli: I was originally deleting this code because we are planning for delete environment.  However, it looks DeleteEnvironment is not ready yet. I will actually revert the dynamodb clean table code back.

wbingli:
## Update for this PR
I have merged the latest code about scheduler change.  I have successfully passed the E2E test in my local, but there are few fixes will be updated for this PR:

- Revert the clean dynamo tables code back.  The `delete-environment` API is not ready yet,  we still need this code to clean the test environment. (Also avoid the scheduler check for this env)
- Some fixes on the scheduler, some issues to pass the right value to start task.
- Per PR comment, adopt sdk default system properties for sdk config.

end-to-end-tests/src/test/java/com/amazonaws/blox/CreateDaemonEnvironmentTest.java
```diff
@@ -0,0 +1,118 @@
...
+import com.amazonaws.blox.model.Environment;
+import com.amazonaws.blox.model.EnvironmentRevision;
+import com.amazonaws.blox.model.StartDeploymentRequest;
+import java.util.UUID;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class CreateDaemonEnvironmentTest {
+  private static final String ENVIRONMENT_NAME = "EndToEndTestEnvironment_" + UUID.randomUUID();
```
wjbuys: Making this `static final` will reuse the name for all tests, so we have to teardown after each test. If we made it non-static, then each test would get its own environment name, and we'd only have to clean up once after the entire suite has run (but we can revisit this later).

wjbuys (approved):

end-to-end-tests/src/test/java/com/amazonaws/blox/CreateDaemonEnvironmentTest.java
wbingli: y, update environment name to be per test case in new commit.

PR #345, opened 2018-01-19
Add .gradletasknamecache to gitignore for using gradle command completion

wbingli:
#### Summary
This meta file is used to cache gradle task if using gradle auto completion.

#### Licensing


wjbuys (approved):

wjbuys:
Oooh, how does one use the task name auto-completion?

wbingli:
It can auto complete or suggest for tasks and arguments.  It will do if type tab key, see https://github.com/gradle/gradle-completion .

PR #346, opened 2018-01-22
Scheduler-manager change to use updated data service model and APIs

gongmax:
#### Summary
Wiring up updated data service model and APIs to schedule manager. Use mocked DataService instead of FakeDataService in scheduler manager's unit tests. 

#### Testing
./gradlew build
./gradlew cucumberTest

#### Licensing


scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/state/ECSState.java
```diff
@@ -16,5 +16,5 @@
 
 public interface ECSState {
 
-  ClusterSnapshot snapshotState(String clusterArn);
+  ClusterSnapshot snapshotState(String clusterName);
```
wbingli, line 19: This interface will have to change once we do a real job, to use assume role to get the customer cluster state. With only the cluster name, it's not possible to know customer service role arn. 

But as long as for M1, this PR should be fine to only focus on using new DS API.

scheduling-manager/src/test/java/com/amazonaws/blox/scheduling/manager/ManagerHandlerTest.java
```diff
@@ -37,9 +39,25 @@
...
-  public static final String FIRST_ENVIRONMENT_ARN = "arn:::::environment1";
-  public static final String SECOND_ENVIRONMENT_ARN = "arn:::::environment2";
+  public static final String ACCOUNT_ID = "12345678012";
+  public static final String CLUSTER_NAME = "cluster1";
+  public static final String FIRST_ENVIRONMENT_NAME = "environment1";
+  public static final String SECOND_ENVIRONMENT_NAME = "environment2";
+
+  private static final Cluster CLUSTER =
+      Cluster.builder().accountId(ACCOUNT_ID).clusterName(CLUSTER_NAME).build();
+  private final EnvironmentId firstEnvironmentId =
```
wbingli: could be also a static for environment ID since we don't intent to be different for different test case.

wbingli (approved):
Minor comments, LGTM.

scheduling-manager/src/test/java/com/amazonaws/blox/scheduling/manager/ManagerHandlerTest.java
gongmax: Good catch. Fixed

PR #347, opened 2018-01-22
add @ignore to those temporarily failed cucumber tests

xjhe:
#### Summary

Simple change! Add @ignore to cucumber tests that failed because of the exception messages/unimplemented DS APIs, so the failure won't distract people when adding new cucumber tests. 

I was intended to submit this change together with other changes, but I'll have to work on my oncall tasks these two days, so I just submit this change separately first, so it won't block others. 

#### Testing
./gradlew check
./gradlew build
./gradlew cucumberTest

New tests cover the changes: 

#### Licensing


wjbuys (approved):

gongmax (commented):
How about the "Create an environment" scenario?

gongmax (approved):

PR #350, opened 2018-01-23
Cleanup old dataservice API

wbingli:
#### Summary
After migrated for scheduler, the old DS API is no long useful. 

#### Testing
`gradle build`

#### Licensing


xjhe (approved):
This looks fine. Have we confirmed that we have moved any usage of this old API to the new one?

wjbuys (approved):
Probably also worth doing a `grep v1.old -R .` in addition to the `gradle build` to make sure it's all gone...

wbingli:
Yes, I have used search tool and make sure no source code has the old code.

PR #351, opened 2018-01-23
Add delete-environment FE API

wbingli:
#### Summary
This PR add the delete-environment FE API implementation.  Some of the work is not covered in this PR, like swagger update, client SDK update etc.   Will be covered in next PR.

#### Implementation details
- API implementation
- controller unit test
- API integration test
- Refactoring on controller test to use spring for controller instance

#### Testing
`gradle check`

#### Licensing


frontend-service/src/main/java/com/amazonaws/blox/frontend/mappers/DeleteEnvironmentMapper.java
```diff
@@ -0,0 +1,37 @@
...
+package com.amazonaws.blox.frontend.mappers;
+
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.DeleteEnvironmentRequest;
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.DeleteEnvironmentResponse;
+import com.amazonaws.blox.frontend.operations.DeleteEnvironment;
+import com.amazonaws.serverless.proxy.internal.model.ApiGatewayRequestContext;
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+
+@Mapper(uses = EnvironmentMapper.class)
```
xjhe, line 24: Do we need a unit test for the mapper (even though this is quite simple)?

wbingli: Little bit tricky here to unit test the DeleteEnvironmentMapper. 
tl;td    The mapper test is inside the `DeleteEnvironmentTest` controller test.

I personally prefer to have unit test per one class only. As you may see example like `StartDeploymentMapper`.  However, the `DeleteEnvironmentMapper` is a little bit different. This mapper depends on the `EnvrionmentMapper`, and we are using spring model to generate the code,  we need to load those mappers into spring context to wire them up.

Instead of having other Spring runner to test the mapper only,  I refactored current controller unit test a little bit, so the mapper will be tested with the controller.

xjhe: Right, I can see that in order to test the mapper we need to wire up the EnvironmentMapper. I think doing it via the controller is fine. As long as there's a way to test the mapper it should be fine.

xjhe (approved):

frontend-service/src/test/java/com/amazonaws/blox/frontend/operations/EnvironmentControllerTestCase.java
```diff
@@ -51,13 +60,29 @@
...
   protected static final String ACCOUNT_ID = "1234567890";
-  protected MockHttpServletRequest servletRequest;
-  ApiGatewayRequestContext requestContext = new ApiGatewayRequestContext();
-  DataService dataService = mock(DataService.class);
+  @Autowired DataService dataService;
+  @Autowired HttpServletRequest servletRequest;
+
+  @Configuration
+  @Import(MapperConfiguration.class)
+  @ComponentScan("com.amazonaws.blox.frontend.operations")
```
wjbuys, line 68: Hm. Is there a reason that this is setting up a totally new config, rather than importing the main Application config and overriding the necessary beans?

frontend-service/src/test/java/com/amazonaws/blox/frontend/integration/DeleteEnvironmentIntegrationTest.java
```diff
@@ -0,0 +1,99 @@
...
+            .build();
+  }
+
+  @Test
+  public void testDeleteEnvironment() throws Exception {
+    when(dataService.deleteEnvironment(any()))
+        .thenReturn(DeleteEnvironmentResponse.builder().environment(environment).build());
+
+    final AwsProxyRequest request =
+        new AwsProxyRequestBuilder("/v1/myCluster/environments/myEnv", "DELETE").build();
```
wjbuys, line 61: Hm. Yeah, I'm not convinced that we want this layer of testing for every API. I think it's okay to merge in as a starting point, but we should consider turning this integration test into a functional test of just the FE behaviours.

wjbuys (approved):

frontend-service/src/test/java/com/amazonaws/blox/frontend/operations/EnvironmentControllerTestCase.java
wbingli: The Spring runner cannot work with both `@ContextConfiguration` and `@ComponentScan`. So need to use a Configuration file to load all configurations. 

As for directly using the main Application config `Application`, that one actually includes too many dependencies,  and have system properties for lambda configuration, enable mvc etc.   The `EnvironmentControllerTestCase` is still lightweight unit test, so use a new config just for controller dependencies.

frontend-service/src/test/java/com/amazonaws/blox/frontend/integration/DeleteEnvironmentIntegrationTest.java
wbingli: Sure. I will merge this one, and let's update the test later per our integration discussion.

PR #352, opened 2018-01-23
Use compile/compileOnly instead of net-ltgt-apt plugin

wjbuys:
The net-ltgt-apt plugin does not seem to play well with IntelliJ. Even
after following [the instructions on the website][1] to the letter,
lombok would not correctly show up in the Gradle project's module
dependencies without being declared as a full compile-time dependency.

Since neither lombok nor mapstruct's annotation processors actually have
any dependencies that could cause problems by being included on the
compile classpath, we're not actually getting much value out of
explicitly declaring these dependencies.

This does mean that we're potentially rebuilding everything when the
classpath changes, due to the issue explained [here][2]. However, I
think having a working IDE is more valuable than that particular
performance optimization.

[1]: https://github.com/tbroyer/gradle-apt-plugin#intellij-idea
[2]: https://blog.gradle.org/incremental-compiler-avoidance#about-annotation-processors

#### Testing
Clean build from command line:
```
gradle clean
gradle check
```
Rebuild from command line:
```
gradle check
```

Clean build from IDE:
```
gradle clean
Build -> Rebuild Project from IntelliJ menu
```


frontend-service/build.gradle
```diff
@@ -37,12 +36,10 @@ dependencies {
     compile "org.mapstruct:mapstruct-jdk8"
 
     compileOnly "org.mapstruct:mapstruct-processor"
-    apt "org.mapstruct:mapstruct-processor"
```
wbingli, line 40: y, the `compileOnly` is required to make IntelliJ work with annotation processor.  It looks with `compileOnly`, it will make gradle compile work as well. I was thinking the `apt` will actual configure some compile options to make gradle make the compile right. 

If `compileOnly` works for gradle to process annotation, it makes sense to only have it.

wbingli (approved):

frontend-service/build.gradle
wjbuys: The links in the PR description have more information on what the `apt` dependency type gets us. (tldr: better transitive dependency isolation, faster incremental compile in some edge cases)

PR #353, opened 2018-01-24
Use a GSI query instead of scan to list clusters by accountId/name

wjbuys:
This uses a GSI to power listClusters queries that have to filter by
accountId and clusterNamePrefix.

Unfortunately, it seems like using a separate POJO for the GSI
(EnvironmentClusterIndexDDBRecord) doesn't work as the docs say it does.
In particular, trying to do any queries using that POJO fails due to a
missing hash key attribute annotation. Therefore this change moves the
indexing annotations back into EnvironmentDDBRecord.

Also unfortunately, DDB does not deduplicate records in a GSI by
hash/range key, so we still have one record per Environment in the
cluster index, and we have to manually deduplicate the values we return.

To make testing easier, the unit tests for listClusters use DDBLocal so
that we don't have to construct accurate mocks for the DDB API's
behaviour.  For more testing convenience, this also introduces Test Data
Builders (see [this article][1] for definition) to make it easier to
construct valid model POJOs for our tests.

[1]: http://blog.codeleak.pl/2014/06/test-data-builders-and-object-mother.html

#### Testing
```
gradle check
gradle cucumberTest
```


data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
```diff
@@ -396,25 +395,36 @@ public void deleteEnvironmentRevision(@NonNull final EnvironmentRevision environ
...
+          dynamoDBMapper.scan(
+              EnvironmentDDBRecord.class,
+              new DynamoDBScanExpression()
+                  .withIndexName(EnvironmentDDBRecord.ENVIRONMENT_CLUSTER_GSI_NAME));
+    } else {
+      DynamoDBQueryExpression<EnvironmentDDBRecord> query =
+          new DynamoDBQueryExpression<EnvironmentDDBRecord>()
+              .withIndexName(EnvironmentDDBRecord.ENVIRONMENT_CLUSTER_GSI_NAME)
+              .withHashKeyValues(EnvironmentDDBRecord.builder().accountId(accountId).build())
+              .withConsistentRead(false);
```
wbingli, line 411: Dynamo GSI don't support consistent read,  not sure if this configuration is necessary.

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryTestCase.java
```diff
@@ -0,0 +1,59 @@
...
+import com.amazonaws.blox.dataservice.repository.model.EnvironmentRevisionDDBRecord;
+import com.amazonaws.blox.dataservice.test.rules.LocalDynamoDb;
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapper;
+import com.amazonaws.services.dynamodbv2.model.CreateTableRequest;
+import com.amazonaws.services.dynamodbv2.model.ProvisionedThroughput;
+import org.junit.Before;
+import org.junit.Rule;
+import org.mapstruct.factory.Mappers;
+
+public class EnvironmentRepositoryTestCase {
```
wbingli: `TestCase` -> `TestBase`?

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/ListClustersTest.java
```diff
@@ -0,0 +1,104 @@
...
+  private static final String OTHER_ENVIRONMENT_NAME = "EnvironmentTwo";
+  ModelBuilders models = ModelBuilders.builder().build();
+
+  List<EnvironmentId> ids =
+      Arrays.asList(
+          models.environmentId("1", "ClusterOne", "EnvironmentOne"),
+          models.environmentId("2", "ClusterTwo", "EnvironmentTwo"),
+          models.environmentId("2", "ClusterThree", "EnvironmentThree"));
+
+  public void givenEnvironments(List<EnvironmentId> ids) throws Exception {
```
wbingli: should be private? Would be better to move at the end of the class, as it's helper methods.

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/ListClustersTest.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+  }
+
+  public void givenEnvironments(EnvironmentId... ids) throws Exception {
+    givenEnvironments(Arrays.asList(ids));
+  }
+
+  @Test
+  public void itDoesntIncludeDuplicateClusters() throws Exception {
+    givenEnvironments(ids);
```
wbingli: Looks `givenEnvironments(ids);` is used for all test cases, maybe move into a @Before

data-service/src/test/java/com/amazonaws/blox/dataservice/test/data/ApiModelBuilders.java
```diff
@@ -0,0 +1,17 @@
...
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "LICENSE" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.blox.dataservice.test.data;
+
+public class ApiModelBuilders {}
```
wbingli: not used class?

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/ListClustersTest.java
```diff
@@ -0,0 +1,104 @@
...
+    assertThat(repo.listClusters(ids.get(0).getAccountId(), null))
+        .extracting("accountId", "clusterName")
+        .containsExactlyInAnyOrder(tuple(ids.get(0).getAccountId(), ids.get(0).getCluster()));
+  }
+
+  @Test
+  public void itFiltersClustersByAccountIdAndClusterPrefixIfGiven() throws Exception {
+    givenEnvironments(ids);
+
+    assertThat(repo.listClusters(ids.get(1).getAccountId(), "ClusterT"))
```
wbingli: How about directly use the value instead of `ids.get(1).getAccountId()`?    like`repo.listClusters("1", "Cluster")`.  Otherwise, we have to go back and read the test data to know what's the value it use. 

Or even we could name the account name as "account_id_with_two_clusters" to have more readable code.

wbingli (commented):
Some minor comments.

data-service/src/main/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryDDB.java
wjbuys: It is; the default is `true`, and if this is omitted, the call fails.

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryTestCase.java
wjbuys: Hm. I've always named my test base classes `*TestCase` because of ye Olden Days where all JUnit tests had to derive from the `org.junit.TestCase` class, but I've not really thought about that in a long time. How about `AbstractEnvironmentRepositoryTest`?

wjbuys: Which reminds me, this should be abstract...

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/ListClustersTest.java
wjbuys: I had it that way before, and then I had to use slightly different givens for each case, and it was weird that some givens were in the before block and some were in the test case. I think this is starting to smell of the Large Shared Fixture antipattern, let me try your suggestion below instead.

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/ListClustersTest.java
wjbuys: I like it! Makes sense.

data-service/src/test/java/com/amazonaws/blox/dataservice/test/data/ApiModelBuilders.java
wjbuys: Ah, hm. I wanted to establish the naming pattern for the Builders we'd use for the `data-service-model` classes. Perhaps a bit premature, will remove.

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/EnvironmentRepositoryTestCase.java
wbingli: Y,  we have to make it as abstract if the test case class end with `Test`. Junit will fail to say no test case if a non-abstract class ends with `Test` without actual test case.

data-service/src/test/java/com/amazonaws/blox/dataservice/repository/ListClustersTest.java
wjbuys: I'm thinking something like this for the test cases:

```java
  @Test
  public void itDoesntIncludeDuplicateClusters() throws Exception {
    EnvironmentIdBuilder sameCluster = models.environmentId().accountId("1").cluster("TestCluster");
    givenEnvironments(
        sameCluster.environmentName("EnvironmentOne").build(),
        sameCluster.environmentName("EnvironmentTwo").build());

    assertThat(repo.listClusters("1", "TestCluster"))
        .extracting("accountId", "clusterName")
        .containsExactlyInAnyOrder(tuple("1", "TestCluster"));
  }
```

For short cases like these I think you're right, not extracting the constants does make it easier to read the test case.

wbingli (approved):
Nice

PR #354, opened 2018-01-25
Generate swagger from a package and update swagger for delete-environ…

wbingli:
#### Summary
Update the swagger file for delete-environment API.

-  update the gradle task to generate the swagger based on package scan.
- Remove the x-generated-at in the swagger file

#### Testing
`gradle build`

#### Licensing


buildSrc/src/main/java/com/amazonaws/blox/tasks/GenerateSwaggerModel.java
```diff
@@ -76,18 +77,23 @@
...
+      Reflections reflections =
+          new Reflections(
+              new ConfigurationBuilder()
+                  .addClassLoaders(loader)
+                  .addUrls(ClasspathHelper.forPackage(pkg, loader))
+                  .setExpandSuperTypes(false)
+                  .filterInputsBy(new FilterBuilder().includePackage(pkg)));
+      Set<Class<?>> springMVCClasses =
+          reflections.getTypesAnnotatedWith(
+              org.springframework.web.bind.annotation.RestController.class);
```
wjbuys, line 95: Hm, we don't have a direct dependency on this in the `build.gradle`, is it coming in as a transitive dependency? Might be worth adding an explicit dependency in that case.

frontend-service/build.gradle
```diff
@@ -60,15 +60,8 @@ task swagger(type: GenerateSwaggerModel, dependsOn: classes) {
     scanClasspath = project.sourceSets.main.runtimeClasspath
 
     // TODO: Specify API classes by package rather than name
```
wjbuys: You can probably remove this TODO now ;)

buildSrc/src/main/java/com/amazonaws/blox/tasks/GenerateSwaggerModel.java
```diff
@@ -53,10 +56,8 @@
 public class GenerateSwaggerModel extends DefaultTask {
   private Log log = new SystemStreamLog();
 
-  public static final SwaggerFilter DEFAULT_FILTER = new GenerationTimestampFilter();
-
-  /** A list of class names to scan for Swagger annotations */
-  @Input private List<String> apiClasses = new ArrayList<>();
+  /** A list of package names to scan for swagger annotations */
+  @Input private List<String> apiPackages = new ArrayList<>();
```
wjbuys, line 60: 🎉

wjbuys (approved):
LGTM, optional minor fixes as you see fit

buildSrc/src/main/java/com/amazonaws/blox/tasks/GenerateSwaggerModel.java
wbingli: Yes, it comes from the "swagger-maven-plugin", which we use to generate swagger from Spring mvc.  

Not sure what's the value we add for explicit dependency for spring-web. I would see it's better to use the transitive dependency from swagger to avoid potential problem to maintain the same spring version within swagger or different spring versions.

frontend-service/build.gradle
wbingli: Will do

PR #358, opened 2018-01-26
Add Junit integration test for data service createEnvironment

xjhe:
#### Summary

Create integration test for data service for CreateEnvironment.
The task was initially to create the integ test for DescribeEnvironmentRevision, but that test requires us to implement the listEnvironmentRevison api, so I switched to CreateEnvironment. 

#### Implementation details

1. I follow the scenario described in the original cucumber feature. 
2. I only do spring wiring for some inputs, since I couldn't figure out how to put the dynamoDB local into spring. 
3. I took some of the table creation steps from the ListClusterTest, but not sure if we should factor it out into a common class, since the ListClusterTest is intended for unit test and this is integration test. Can they still use the same utility? 

#### Testing
./gradlew check
./gradlew build clean

Not very slow! It's similar to the speed of the unit test. 

New tests cover the changes: 

#### Licensing


data-service/src/test/java/com/amazonaws/blox/dataservice/integration/DataServiceIntegrationTestConfig.java
```diff
@@ -0,0 +1,35 @@
...
+package com.amazonaws.blox.dataservice.integration;
+
+import com.amazonaws.blox.dataservice.mapper.ApiModelMapper;
+import com.amazonaws.blox.dataservice.mapper.EnvironmentMapper;
+import org.mapstruct.factory.Mappers;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class DataServiceIntegrationTestConfig {
```
wbingli: You can load most of the data service configuration directly.   I'm not sure if we can directly import the `Application.class` and replace only the `@Bean public AmazonDynamoDB dynamoDBClient()` like below. If we can override it,  then we load almost the same as real runtime. Otherwise, you can load other beans by ComponentScan.
```
  @Bean
  public AmazonDynamoDB dynamoDBClient() {
    amazonDynamoDBLocal = DynamoDBEmbedded.create();
    return amazonDynamoDBLocal.amazonDynamoDB();
  }
```

data-service/src/test/java/com/amazonaws/blox/dataservice/integration/DataServiceIntegrationTestBase.java
```diff
@@ -0,0 +1,93 @@
...
+        new DataServiceApi(
+            new CreateEnvironmentApi(apiModelMapper, repo),
+            new DescribeEnvironmentApi(apiModelMapper, repo),
+            new StartDeploymentApi(apiModelMapper, repo),
+            new ListEnvironmentsApi(apiModelMapper, repo),
+            new DescribeEnvironmentRevisionApi(apiModelMapper, repo),
+            new ListClustersApi(apiModelMapper, repo),
+            new DeleteEnvironmentApi(apiModelMapper, repo));
+
+    ProvisionedThroughput throughput =
```
wbingli: The tables setup is duplicated with `EnvironmentRepositoryTestBase`.  We could create a new class, such as `DynamoDBLocalSetup` for the table creation.

data-service/src/test/java/com/amazonaws/blox/dataservice/integration/CreateEnvironmentIntegrationTest.java
```diff
@@ -0,0 +1,140 @@
...
+    // The second environment is also valid
+    checkEnvironmentValid(ENVIRONMENT_NAME, ACCOUNT_ID, CLUSTER_TWO, createEnvironmentResponse2);
+  }
+
+  private void checkEnvironmentValid(
+      final String environmentName,
+      final String accountId,
+      final String cluster,
+      final CreateEnvironmentResponse createEnvironmentResponse) {
+    assertEquals(
```
wbingli: Better to use assertj, `assertThat(..).isEqualTo(..)`. It provides both type safe, easy to read and easy to use with fluent api.

data-service/src/test/java/com/amazonaws/blox/dataservice/integration/CreateEnvironmentIntegrationTest.java
```diff
@@ -0,0 +1,140 @@
...
+  @Test
+  public void testCreateEnvironmentSuccessful() throws Exception {
+    final EnvironmentId createdEnvironmentId =
+        models
+            .environmentId()
+            .accountId(ACCOUNT_ID)
+            .environmentName(ENVIRONMENT_NAME)
+            .cluster(CLUSTER_ONE)
+            .build();
+    // I create an environment
```
wbingli: This comment repeat below code,   I don't see the value to keep the cucumber sentences.

data-service/src/test/java/com/amazonaws/blox/dataservice/integration/CreateEnvironmentIntegrationTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+    // The environment is valid
+    checkEnvironmentValid(ENVIRONMENT_NAME, ACCOUNT_ID, CLUSTER_ONE, createEnvironmentResponse);
+  }
+
+  @Test
+  public void testCreateAnEnvironmentAlreadyExist() throws Exception {
+    final EnvironmentId createdEnvironmentId =
+        models
+            .environmentId()
```
wbingli: EnvironmentId is the same everywhere and not really important to have it for each test case.  How about moving it as field?

xjhe: It's only on the third test that the id is different (having two different clusters). Do you suggest just create these two IDs as fields?

data-service/src/test/java/com/amazonaws/blox/dataservice/integration/CreateEnvironmentIntegrationTest.java
xjhe: Sure

data-service/src/test/java/com/amazonaws/blox/dataservice/integration/DataServiceIntegrationTestBase.java
xjhe: I was thinking about that, but not sure if it's clear enough to use this common class for both unit test (the repository test) and integ test ( the data service test)

data-service/src/test/java/com/amazonaws/blox/dataservice/integration/DataServiceIntegrationTestConfig.java
xjhe: Sure, I can load the other configs from Application.class. The problem is with the dynamoDB local -- the DDBlocal should be created in the @before method and should be shutdown in the @after method. I tried to create the dynamoDBClient here, but I can't shutdown the amazonDynamoDBLocal later. Any suggestions on this?

wbingli: You don't have to create the `DynamoDBEmbedded` instance during the `@Before`. Create it as a bean will create an instance for the Spring context.  You only need to create all the tables and **may** need to delete all tables after each test case. But you can also refresh the whole Spring context to generate all new beans, which will force to create a new DynamoDBEmbedded instance.

As for the DynamoDBEmbedded instance shutdown,  I'm not sure whether it's required, if it's in-memory db, it should be gone once the process done.  If a shutdown provides value for the DynamoDBEmbedded instance, you may use Spring `@PreDestroy` to call the shutdown.  That means you can create a wrap class on the DynamoDBEmbedded, which could be used for both setup tables and shutdown instance.

data-service/src/test/java/com/amazonaws/blox/dataservice/integration/CreateEnvironmentIntegrationTest.java
wbingli: y, could have a default environment ID. As most of the test case don't care it too much, the creation of environment ID (7 lines of code) is quite distraction to read the test.

data-service/src/test/java/com/amazonaws/blox/dataservice/integration/DataServiceIntegrationTestConfig.java
xjhe: I think the purpose for shutdown() is just to delete all the created tables. If we don't call it at the end, we can either manually delete all tables after each test, or write the test in a way that it can use the content from the previous tests. I can try deleting the table and see if it works fine.

wbingli: I don't think it will delete the tables. I  should stop all the jobs and release resources. It may only have value if you create a DynamoDBEmbedded to a local file as `DynamoDBEmbedded.create(...File...)`, as the shutdown can release the resource on the file. But for in-memory, we should be safe.

xjhe: Tried with overriding the dynamoDBClient bean. The creation was fine, but the next test will complain that the table has already been created. I think it's because the resource only gets released after we're done with all the unit tests (at the end of the process), but not after each unit test. 
I go with deleting the tables after each test and it's fine...

wjbuys: Could we use a `@DirtiesContext` to spin up a new context for each test case?

data-service/src/test/java/com/amazonaws/blox/dataservice/integration/CreateEnvironmentIntegrationTest.java
wjbuys: And I'm willing to bet that that will make this readable enough to inline the `checkEnvironmentValid` method.

data-service/src/test/java/com/amazonaws/blox/dataservice/integration/DataServiceIntegrationTestConfig.java
xjhe: Changing into @DirtiesContext also works!

data-service/src/test/java/com/amazonaws/blox/dataservice/integration/CreateEnvironmentIntegrationTest.java
```diff
@@ -0,0 +1,116 @@
...
+  @Test
+  public void testCreateEnvironmentSuccessful() throws Exception {
+    final CreateEnvironmentResponse createEnvironmentResponse =
+        dataService.createEnvironment(
+            models
+                .createEnvironmentRequest()
+                .taskDefinition(TASK_DEFINITION)
+                .environmentId(createdEnvironmentId1)
+                .build());
+    checkEnvironmentValid(ENVIRONMENT_NAME, ACCOUNT_ID, CLUSTER_ONE, createEnvironmentResponse);
```
wbingli: Should it be `assertThat(createEnvironmentResponse.getEnvironment(). getEnvironmentId()).isEqualTo(createdEnvironmentId1)`?   Therefore, we can remove the `checkEnvironmentValid` method.

data-service/src/test/java/com/amazonaws/blox/dataservice/integration/DataServiceIntegrationTestBase.java
```diff
@@ -0,0 +1,42 @@
...
+
+@ContextConfiguration(classes = DataServiceIntegrationTestConfig.class)
+@RunWith(SpringJUnit4ClassRunner.class)
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)
+public abstract class DataServiceIntegrationTestBase {
+  @Autowired private AmazonDynamoDB amazonDynamoDB;
+  @Autowired private DynamoDBMapper dynamoDBMapper;
+  @Autowired DataService dataService;
+
+  private final DynamoDBLocalSetup dynamoDBLocalSetup = new DynamoDBLocalSetup();
```
wbingli: It can merged into `setup` as local variable.

data-service/src/test/java/com/amazonaws/blox/dataservice/test/data/DynamoDBLocalSetup.java
```diff
@@ -0,0 +1,57 @@
...
+    CreateTableRequest createEnvironmentRevisions =
+        dynamoDBMapper
+            .generateCreateTableRequest(EnvironmentRevisionDDBRecord.class)
+            .withProvisionedThroughput(throughput);
+
+    amazonDynamoDB.createTable(createEnvironments);
+    amazonDynamoDB.createTable(createEnvironmentRevisions);
+  }
+
+  public void deleteTables(
```
wbingli: Don't think we use this method. As we use the `@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)`, we actually create new local dyanmodb instance for each test case.  The `deleteTables` is no longer required.    Probably should remove it unless we need it.

data-service/src/test/java/com/amazonaws/blox/dataservice/integration/CreateEnvironmentIntegrationTest.java
xjhe: Makes sense!

data-service/src/test/java/com/amazonaws/blox/dataservice/test/data/DynamoDBLocalSetup.java
xjhe: Deleted!

data-service/src/test/java/com/amazonaws/blox/dataservice/integration/CreateEnvironmentIntegrationTest.java
```diff
@@ -0,0 +1,106 @@
...
+  @Test
+  public void testCreateAnEnvironmentAlreadyExist() throws Exception {
+    dataService.createEnvironment(
+        models
+            .createEnvironmentRequest()
+            .taskDefinition(TASK_DEFINITION)
+            .environmentId(createdEnvironmentId1)
+            .build());
+
+    thrown.expect(ResourceExistsException.class);
```
wjbuys: No need to change for this PR, but since we already use AssertJ for other assertions, it might be worth using the `assertThatThrownBy` method to assert on exceptions: http://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html#exception-assertion (but only if it looks more readable).

data-service/src/test/java/com/amazonaws/blox/dataservice/integration/DataServiceIntegrationTestBase.java
```diff
@@ -0,0 +1,41 @@
...
+@RunWith(SpringJUnit4ClassRunner.class)
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)
+public abstract class DataServiceIntegrationTestBase {
+  @Autowired private AmazonDynamoDB amazonDynamoDB;
+  @Autowired private DynamoDBMapper dynamoDBMapper;
+  @Autowired DataService dataService;
+
+  @Before
+  public void setup() {
+    final DynamoDBLocalSetup dynamoDBLocalSetup = new DynamoDBLocalSetup();
```
wjbuys: If the `DynamoDbLocalSetup` class takes the mapper and ddb instance as constructor args, then it can also be injected into this test context (rather than constructing it inline).

wjbuys (approved):
Nice! This looks nice and readable.

wbingli (approved):

data-service/src/test/java/com/amazonaws/blox/dataservice/integration/DataServiceIntegrationTestBase.java
xjhe: Good point. I think Wenbing also pointed this out last week, but I forgot that comment :-( Made the change. I leave EnvironmentRepositoryTestBase as it is, but I think it can be turn into a similar format with the DataServiceIntegrationTestBase? I can do that in a later PR

data-service/src/test/java/com/amazonaws/blox/dataservice/integration/CreateEnvironmentIntegrationTest.java
xjhe: Sure. I made the change.
I notice one thing during my debug: even if I didn't wrote the assertThatThrownBy correctly, it will still pass the Intellij test, and will only fail the gradle check. I may need to see if you have any special configuration in intelliJ for it...

PR #359, opened 2018-01-26
Update E2E test to use DeleteEnvironment API to cleanup environment

wbingli:
#### Summary
We have delete-environment API ready. This PR updates the e2e test with this API. Also remove all the dependencies for data-server, dynamodb and lambda.

A bug is fixed inside data-service for delete-environment.  In this stage, use the CLOBB config to delete the dynamo record anyway.  In a long term, it will be a different implementation, which frondend will only update the status somehow.

#### Implementation details
- Fix DeleteEnv DDB issue to delete record anyway
- Remove e2e project dependency on data-service, dynamo and lambda json rpc
- Remove Dyanmodb helper for e2e test

#### Testing
`gradle build`
`gradle testEndToEnd`

#### Licensing


xjhe (approved):
Looks good -- have we verified that we can actually delete the environment from the end to end test?

wbingli:
The delete environment is manually verified.  We should also used API to verified the environment is deleted. However, we need to fix the error handling of FE API. As for now, the resource not found exception will cause 500.

wjbuys (approved):

PR #360, opened 2018-01-26
Sort swagger model to make the swagger.yml generation unique per model

wbingli:
#### Summary
There is a problem that swagger.yml may change even no change on swagger model.  The root cause is swagger use `LinkedHashMap` for most the fields. It causes sometimes yaml file change while  

This fix will use a `TreeMap` (implementation of SortedMap) to replace existing map.  So the yaml file will be naturely sorted.

Other solution is loading current swagger.yml into Swagger model and compare them too before write to file, but that looks need external dependency for swagger parser. This approach looks good enough.

#### Testing
`gradle --rerun-tasks swagger`  several times, no swagger.yml file change anymore.

#### Licensing


wjbuys (approved):
OMG THANK YOU! This was so annoying.

buildSrc/src/main/java/com/amazonaws/blox/swagger/SortedModelFilter.java
```diff
@@ -0,0 +1,40 @@
...
+
+import io.swagger.models.Swagger;
+import java.util.TreeMap;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+/**
+ * Swagger model uses LinkedHashMap for data like paths, definitions. Due to no order guarantee, it
```
wjbuys: Don't doc comments have to go before annotations?

wbingli: Oh, will fix it.

PR #361, opened 2018-01-29
Refactor integration test to verify Lambda and Spring integration with a sample controller

wbingli:
#### Summary
As for our integration test,  we want to use the integration test to focus on following:
 - Lambda to Spring mapping 
 - Spring context wiring 
 - Exception handling

As for the API logic itself, we will test each API in the controller unit test.

In this PR, it removes the per API integration and replace with a sample controller. It will verify the Lambda request to the Spring controller mapping, such as Request mapping, response etc.  

It also keep the whole Spring context scan as the same as production runtime to verify Spring context wiring.

As for the next step, we will use this framework to test the exception handling. 

#### Implementation details
Notice that I don't use lombok for the `SampleResponse` bean. Because lombok is a compileOnly dependency. Instead of adding extra dependency,  the basic constructor and getter/setter makes more sense. 

#### Testing
`gradle check`

#### Licensing


frontend-service/src/test/java/com/amazonaws/blox/frontend/integration/IntegrationTestBase.java
```diff
@@ -40,28 +36,9 @@
 )
 public abstract class IntegrationTestBase {
```
wjbuys, line 40: Now that we'll only have the one integration test, is it still worth having the base class?

frontend-service/src/test/java/com/amazonaws/blox/frontend/integration/IntegrationTestConfig.java
```diff
@@ -48,6 +49,16 @@
...
   }
 
+  @Bean
+  public SampleController sampleController() {
+    return new SampleController();
+  }
+
+  @Bean
+  public Checker sampleControllerChecker() {
+    return mock(Checker.class);
```
wjbuys, line 59: Did you figure out if this actually resets mocks after every test?

frontend-service/src/test/java/com/amazonaws/blox/frontend/integration/SampleController.java
```diff
@@ -0,0 +1,59 @@
...
+    return checker.handle(path, query);
+  }
+
+  static class SampleResponse {
+    private String result;
+
+    public SampleResponse() {}
+
+    public SampleResponse(final String result) {
+      this.result = result;
```
wjbuys: I'd be totally fine with including lombok as a `testCompileOnly` dependency, I've been doing that for the JSON-RPC tests.

wjbuys (approved):
Nice and simple, I like it. Feel free ship this and address comments in your next PR.

frontend-service/src/test/java/com/amazonaws/blox/frontend/integration/IntegrationTestBase.java
wbingli: I plan to use this one for exception testing as well.  Will evaluate if it's still needed then.

frontend-service/src/test/java/com/amazonaws/blox/frontend/integration/SampleController.java
wbingli: OK, will update.

frontend-service/src/test/java/com/amazonaws/blox/frontend/integration/IntegrationTestConfig.java
wbingli: Nope, will still need a reset for the mock once we have more test cases.  Will update to address this issue.

PR #364, opened 2018-01-30
Move all group/version declarations to root project's build.gradle

wjbuys:
This moves the `group` and `version` configuration into the root project for all subprojects. No need to add it to subprojects again in future.

#### Testing
```
gradle check
```

#### Licensing


wbingli (approved):
👍

PR #365, opened 2018-01-30
Map FE Exception and generate related apigateway SDK exception

wbingli:
#### Summary
It takes ResourceNotFoundException from data service as an example for froendend and SDK client error handling, it covers:
 - Handle Spring exception with status code and json response
 - Config API to generate proper exception data model per status code
 - Configure apigateway SDK generation to have proper exception for exception
 - Update end-2-end test to verify the error handling

#### Implementation details
The json-rpc-core manages similar thing for exception serialization and deserialization.  I have a quick look the swagger generation, however, it doesn't really support provided mapper or conventer to translate Throwable to API model.   With a pojo DTO, it looks straight forward and apigateway generates the right exception model.  Potentially we can use mapstruct to do the exception to exception DTO mapping.

We directly use the exception from data service. The actual exception value is not that nice, the `resourceId` is actually `EnvironmentId#toString`.  But we can translate the exception with proper message.

#### Testing
`gradle clean build testEndToEnd`

#### Licensing


end-to-end-tests/src/test/java/com/amazonaws/blox/CreateDaemonEnvironmentTest.java
```diff
@@ -58,6 +60,24 @@ public void tearDown() {
                 .environmentName(environmentName));
     // Cleanup ECS tasks
     stack.reset();
+
+    // Describe the environment again, should be ResourceNotFoundException
+    Throwable thrown =
```
wjbuys: This should really be in a testcase, and not in the environment teardown. Otherwise all tests will fail in teardown if this is broken.

frontend-service-client/src/main/java/com/amazonaws/blox/model/ResourceNotFoundException.java
```diff
@@ -0,0 +1,85 @@
...
+*/
+package com.amazonaws.blox.model;
+
+import javax.annotation.Generated;
+
+/**
+ * 
+ */
+@Generated("com.amazonaws:aws-java-sdk-code-generator")
+public class ResourceNotFoundException extends com.amazonaws.blox.model.BloxException {
```
wjbuys, line 12: I'm kind-of surprised that this worked; how did the generator come up with this (correct) name, but the name of the object in the Swagger model and in the Frontend is ResourceNotFoundException**Response**?

If there's some trick to it, it might be worth documenting somewhere for when we add more exceptions.

frontend-service/src/test/java/com/amazonaws/blox/frontend/integration/ErrorHandlingIntegrationTest.java
```diff
@@ -0,0 +1,52 @@
...
+    final AwsProxyRequest request =
+        new AwsProxyRequestBuilder("/test/testPath/sample", "POST")
+            .queryString("query", "1")
+            .build();
+
+    final AwsProxyResponse response = handler.proxy(request, lambdaContext);
+
+    assertThat(response.getStatusCode()).isEqualTo(404);
+    final JsonNode jsonResponse = mapper.readTree(response.getBody());
+    assertThat(jsonResponse.get("resourceType").asText()).isEqualTo("env");
```
wjbuys, line 49: Should this assert on the message as well?

frontend-service/src/main/java/com/amazonaws/blox/frontend/exception/GlobalControllerExceptionHandler.java
```diff
@@ -0,0 +1,39 @@
...
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ControllerAdvice
+public class GlobalControllerExceptionHandler {
+
+  @ResponseStatus(HttpStatus.NOT_FOUND)
+  @ExceptionHandler(
+      com.amazonaws.blox.dataservicemodel.v1.exception.ResourceNotFoundException.class)
```
wjbuys: Any reason we can't import this long name here? Doesn't looks like there's a conflict anymore...

frontend-service/src/main/java/com/amazonaws/blox/frontend/operations/DescribeEnvironment.java
```diff
@@ -35,6 +38,15 @@
 
   @RequestMapping(path = "/{environmentName}", method = RequestMethod.GET)
   @ApiOperation(value = "Describe an Environment by name")
+  @ApiResponses(
+    value = {
+      @ApiResponse(
+        code = 404,
+        message = "Resource not found",
```
wjbuys: It might be worth extracting the code and message as constants on the ResourceNotFoundExceptionResponse class.

wjbuys (changes requested):
Good start. The main issue is the testing done in the `tearDown` method. In general, no assertions in `tearDown`, as failures there will suppress the actual test failure, IIRC.

The rest is fix-as-you-see-fit.

end-to-end-tests/src/test/java/com/amazonaws/blox/CreateDaemonEnvironmentTest.java
wbingli: It due to the delete-environment is part of the teardown.  When I was thinking about the delete-environment step, I put it into the teardown instead of inside the testcase, because it can try to cleanup the environment even   the test case fails in the middle.

Since we have verification for the delete-environment, it makes sense to be a part of the test case.   Even we potentially may lack data in dyanmodb if test case fail, but we use UUID for environment ID, it should be good for now.

frontend-service-client/src/main/java/com/amazonaws/blox/model/ResourceNotFoundException.java
wbingli: There is a configuration while generating the SDK, see ` "exceptionFor404=ResourceNotFoundException"`  (https://github.com/blox/blox/pull/365/files#diff-c197962302397baf3a4cc36463dce5eaR198)

I'm kind of glad to see apigateway leveraging the exception model we define in the swagger.

frontend-service/src/main/java/com/amazonaws/blox/frontend/exception/GlobalControllerExceptionHandler.java
wbingli: Good catch, will fix it.  It was because I use the same class name at some point.

frontend-service-client/src/main/java/com/amazonaws/blox/model/ResourceNotFoundException.java
wjbuys: Right, but that is referencing `ResourceNotFoundException`, right? How does it know to use the properties from the `ResourceNotFoundExceptionResponse` definition in the swagger?

frontend-service/src/main/java/com/amazonaws/blox/frontend/operations/DescribeEnvironment.java
wbingli: Good point, thanks, will update.

frontend-service/src/test/java/com/amazonaws/blox/frontend/integration/ErrorHandlingIntegrationTest.java
wbingli: Sure

frontend-service-client/src/main/java/com/amazonaws/blox/model/ResourceNotFoundException.java
wbingli: I think it's the interesting part of apigateway sdk generation.   It respects the exception model, but don't respect the exception name in the swagger. If not provided the  "exceptionFor404=ResourceNotFoundException", the default exception `NotFoundException` will be generated and with right data model as well.

wjbuys: Ah, I see, makes sense. Probably worth adding a comment to that effect here.

wjbuys (approved):

wjbuys:
Good to go once you make the test case changes

PR #366, opened 2018-01-30
DataService integration test: happy case!

xjhe:
#### Summary
Add some test for the happy case of DS integration test. More detail test and corner cases are already covered in unit tests.
I couldn't test DescribeEnvironmentRevision and StartDeployment because we don't have ListEnvironments yet (why ? we don't need it in our second milestone?)

Also, figure that in DeleteEnvironmentApi, if we don't use the configuration SaveBehavior.CLOBBER.config(), the deletion will fail. But why? I think the default config will still allow us to delete one environment with just one revision. 

#### Testing
./gradlew check
./gradlew clean build

#### Licensing


data-service/src/test/java/com/amazonaws/blox/dataservice/integration/DataServiceIntegrationTest.java
```diff
@@ -0,0 +1,133 @@
...
+package com.amazonaws.blox.dataservice.integration;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.amazonaws.blox.dataservicemodel.v1.model.EnvironmentId;
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.*;
+import java.util.Arrays;
+import org.junit.Test;
+
+public class DataServiceIntegrationTest extends DataServiceIntegrationTestBase {
```
wbingli: Not sure `DataServiceIntegrationTest` is a good name,  I would suggest we move separate each API into its own test class. Even they are just happy path, but we potentially could add more tests, such as exception handling cases.  Thoughts?

xjhe: That makes sense. I can break it down into separate tests for each API, so later people can add the non-happy case tests in, if necessary.

wjbuys (approved):
Looks good, although I wonder if it's helpful to have both the e.g. `ACCOUNT_ID` constants, as well as the default values on DataServiceModelBuilder. Perhaps we can either use e.g. `models.getAccountId()` in the tests, or set the default account ID on the model builder to `ACCOUNT_ID` where we construct it?

xjhe:
Don't know how to directly reply @wjbuys 's review message...so I post a new message here.
Yes I think getting the account ID from the model builder is fine. In our tests we only use this default account ID, so we don't need to customize it

data-service/src/test/java/com/amazonaws/blox/dataservice/integration/DescribeEnvironmentIntegrationTest.java
```diff
@@ -0,0 +1,51 @@
...
+package com.amazonaws.blox.dataservice.integration;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.amazonaws.blox.dataservicemodel.v1.model.EnvironmentId;
+import com.amazonaws.blox.dataservicemodel.v1.model.wrappers.DescribeEnvironmentResponse;
+import org.junit.Test;
+
+public class DescribeEnvironmentIntegrationTest extends DataServiceIntegrationTestBase {
+  private static final String ACCOUNT_ID = "123456789012";
```
wbingli: Looks the "ACCOUNT_ID", "ENVIRONMENT_NAME" and `envionrmentId` are pretty much the same for all the test,  we most likely could move into the base class.   Even some of the test case need different environment or cluster, we can have different name for those.

xjhe: The ENVIRONMENT_NAME and CLUSTER can be different, but I can use the customized value for them

xjhe:
The new update removes unnecessary values in account id, environment and cluster. Use most of the default values, unless the values are different from the default.

wbingli (approved):
/shipit

PR #367, opened 2018-02-03
Fix deprecation warning from `downloadClient` task

wjbuys:
Using `inputs.file` with a Task as argument was causing a warning:

> Using TaskInputs.file() with something that doesn't resolve to a File object has been deprecated and is scheduled to be removed in Gradle 5.0. Use TaskInputs.files() instead.

#### Testing
`gradle updateClient` no longer show the warning.

#### Licensing


wbingli (approved):
shipit

PR #368, opened 2018-02-05
Add stop task to scheduler

xjhe:
#### Summary
Allow scheduler to stop a task when the task version doesn't match the environment version. 

#### Implementation details
I made some assumptions in my implementation: 
1. When there is one task to stop on one instance and one task to launch on another instance, stop the task first. 
2. There can be multiple tasks per instance. Even though it should be one, but I didn't add any enforcement on this. 
3. We stop a task when its version doesn't match the environment's, regardless of the task status (if it's running or not).

#### Testing
./gradlew check
./gradlew clean build

#### Licensing

### DIFF
--- a/scheduling-manager/src/cucumber/resources/features/scheduling/replace_after_terminate.feature
+++ b/scheduling-manager/src/cucumber/resources/features/scheduling/replace_after_terminate.feature
@@ -43,9 +43,23 @@ Feature: Deploying a Daemon environment to a cluster with Replace After Terminat
       | instance | tasks          |
       | i-1      | t-1:v1:RUNNING |
       | i-2      | t-2:v2:RUNNING |
-      | i-3      |                |
     When the scheduler runs
     Then it should stop the following tasks:
       | clusterName | task           | reason                |
-      | TestCluster | v2             | replaceAfterTerminate |
+      | TestCluster | v2             | Stopped by deployment to DaemonEnvironment@v1 |
+    And it should not take any further actions
+
+  Scenario: Scheduling on a cluster with multiple actions
+    Given the cluster has the following instances and tasks:
+      | instance | tasks          |
+      | i-1      | t-1:v1:RUNNING |
+      | i-2      | t-2:v2:RUNNING |
+      | i-3      |                |
+    When the scheduler runs
+    Then it should start the following tasks:
+      | containerInstanceArn | taskDefinitionArn | group             |
+      | i-3                  | v1                | DaemonEnvironment |
+    Then it should stop the following tasks:
+      | clusterName | task           | reason                |
+      | TestCluster | v2             | Stopped by deployment to DaemonEnvironment@v1 |
     And it should not take any further actions
--- a/scheduling-manager/src/cucumber/resources/features/scheduling/replace_after_terminate.feature
+++ b/scheduling-manager/src/cucumber/resources/features/scheduling/replace_after_terminate.feature
@@ -37,3 +37,15 @@ Feature: Deploying a Daemon environment to a cluster with Replace After Terminat
       | containerInstanceArn | taskDefinitionArn | group             |
       | i-1                  | v1                | DaemonEnvironment |
     And it should not take any further actions
+
+  Scenario: Scheduling on a cluster with tasks that don't match the environment version
+    Given the cluster has the following instances and tasks:
+      | instance | tasks          |
+      | i-1      | t-1:v1:RUNNING |
+      | i-2      | t-2:v2:RUNNING |
+      | i-3      |                |
+    When the scheduler runs
+    Then it should stop the following tasks:
+      | clusterName | task           | reason                |
+      | TestCluster | v2             | replaceAfterTerminate |
+    And it should not take any further actions
--- a/scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
+++ b/scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
@@ -37,7 +37,7 @@ public class DaemonEnvironment {
   private final EnvironmentDescription environment;
 
   public boolean isMissingHealthyTask(List<Task> tasks) {
-    return tasks.stream().noneMatch(this::isTaskRunnable);
+    return tasks.stream().noneMatch(this::isMatchingTask);
   }
 
   public StartTask startTaskFor(ContainerInstance i) {
@@ -66,7 +66,7 @@ public class DaemonEnvironment {
         && HEALTHY_STATES.contains(t.getStatus());
   }
 
-  public boolean isTaskRunnable(Task t) {
+  public boolean isMatchingTask(Task t) {
     return t.getGroup().equals(environment.getEnvironmentName())
         && HEALTHY_STATES.contains(t.getStatus());
   }
--- a/scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
+++ b/scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
@@ -40,10 +40,6 @@ public class DaemonEnvironment {
     return tasks.stream().noneMatch(this::taskDesiredToRun);
   }
 
-  public boolean needsToStop(List<Task> tasks) {
-    return tasks.stream().anyMatch(this::taskToStop);
-  }
-
   public boolean noRunningTasks(List<Task> tasks) {
     return tasks.stream().noneMatch(t -> HEALTHY_STATES.contains(t.getStatus()));
   }
--- a/scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
+++ b/scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
@@ -23,7 +23,6 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.stream.Collectors;
 import lombok.RequiredArgsConstructor;
 
 /**
@@ -37,12 +36,16 @@ public class DaemonEnvironment {
 
   private final EnvironmentDescription environment;
 
-  public boolean hasMatchingTask(List<Task> tasks) {
-    return tasks.stream().noneMatch(this::matchesTask);
+  public boolean needsToAssign(List<Task> tasks) {
+    return tasks.stream().noneMatch(this::taskDesiredToRun);
   }
 
-  public boolean hasUnmatchingTask(List<Task> tasks) {
-    return tasks.stream().anyMatch(this::unmatchesTask);
+  public boolean needsToStop(List<Task> tasks) {
+    return tasks.stream().anyMatch(this::taskToStop);
+  }
+
+  public boolean noRunningTasks(List<Task> tasks) {
+    return tasks.stream().noneMatch(t -> HEALTHY_STATES.contains(t.getStatus()));
   }
 
   public StartTask startTaskFor(ContainerInstance i) {
@@ -54,31 +57,24 @@ public class DaemonEnvironment {
         .build();
   }
 
-  public List<StopTask> stopTaskFor(final List<Task> tasks) {
-    final List<String> tasksToStop =
-        tasks
-            .stream()
-            .filter(this::unmatchesTask)
-            .map(Task::getTaskDefinitionArn)
-            .collect(Collectors.toList());
-    return tasksToStop
-        .stream()
-        .map(
-            t ->
-                StopTask.builder()
-                    .clusterName(environment.getClusterName())
-                    .task(t)
-                    .reason("replaceAfterTerminate")
-                    .build())
-        .collect(Collectors.toList());
+  public StopTask stopTaskFor(final Task task) {
+    return StopTask.builder()
+        .clusterName(environment.getClusterName())
+        .task(task.getTaskDefinitionArn())
+        .reason(
+            String.format(
+                "Stopped by deployment to %s@%s",
+                environment.getEnvironmentName(), environment.getTaskDefinitionArn()))
+        .build();
   }
 
-  public boolean unmatchesTask(Task t) {
-    return !t.getGroup().equals(environment.getEnvironmentName())
-        || !t.getTaskDefinitionArn().equals(environment.getTaskDefinitionArn());
+  public boolean taskToStop(Task t) {
+    return t.getGroup().equals(environment.getEnvironmentName())
+        && !t.getTaskDefinitionArn().equals(environment.getTaskDefinitionArn())
+        && HEALTHY_STATES.contains(t.getStatus());
   }
 
-  public boolean matchesTask(Task t) {
+  public boolean taskDesiredToRun(Task t) {
     return t.getGroup().equals(environment.getEnvironmentName())
         && t.getTaskDefinitionArn().equals(environment.getTaskDefinitionArn())
         && HEALTHY_STATES.contains(t.getStatus());
--- a/scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
+++ b/scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
@@ -36,12 +36,8 @@ public class DaemonEnvironment {
 
   private final EnvironmentDescription environment;
 
-  public boolean needsToAssign(List<Task> tasks) {
-    return tasks.stream().noneMatch(this::taskDesiredToRun);
-  }
-
-  public boolean noRunningTasks(List<Task> tasks) {
-    return tasks.stream().noneMatch(t -> HEALTHY_STATES.contains(t.getStatus()));
+  public boolean isMissingHealthyTask(List<Task> tasks) {
+    return tasks.stream().noneMatch(this::isTaskRunnable);
   }
 
   public StartTask startTaskFor(ContainerInstance i) {
@@ -64,15 +60,14 @@ public class DaemonEnvironment {
         .build();
   }
 
-  public boolean taskToStop(Task t) {
+  public boolean isTaskStoppable(Task t) {
     return t.getGroup().equals(environment.getEnvironmentName())
         && !t.getTaskDefinitionArn().equals(environment.getTaskDefinitionArn())
         && HEALTHY_STATES.contains(t.getStatus());
   }
 
-  public boolean taskDesiredToRun(Task t) {
+  public boolean isTaskRunnable(Task t) {
     return t.getGroup().equals(environment.getEnvironmentName())
-        && t.getTaskDefinitionArn().equals(environment.getTaskDefinitionArn())
         && HEALTHY_STATES.contains(t.getStatus());
   }
 }
--- a/scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
+++ b/scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
@@ -16,12 +16,14 @@ package com.amazonaws.blox.scheduling.scheduler.engine.daemon;
 
 import com.amazonaws.blox.scheduling.scheduler.engine.EnvironmentDescription;
 import com.amazonaws.blox.scheduling.scheduler.engine.StartTask;
+import com.amazonaws.blox.scheduling.scheduler.engine.StopTask;
 import com.amazonaws.blox.scheduling.state.ClusterSnapshot.ContainerInstance;
 import com.amazonaws.blox.scheduling.state.ClusterSnapshot.Task;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 import lombok.RequiredArgsConstructor;
 
 /**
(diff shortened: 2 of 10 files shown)


wjbuys:
> We stop a task when its version doesn't match the environment's, regardless of the task status (if it's running or not).

I don't think this is a safe assumption to make; we should only emit stop events for tasks that aren't already terminating or terminated.

> When there is one task to stop on one instance and one task to launch on another instance, stop the task first.

Does this ordering matter within a single scheduler run? I think the scheduler should emit both events.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/ReplaceAfterTerminateScheduler.java
```diff
@@ -15,18 +15,29 @@
...
+        .forEach(i -> stopTaskActions.addAll(env.stopTaskFor(summary.tasksForInstance(i))));
+
+    List<SchedulingAction> startTaskActions =
+        summary
+            .getInstances()
+            .stream()
+            .filter(i -> env.hasMatchingTask(summary.tasksForInstance(i)))
+            .map(i -> env.startTaskFor(i))
+            .collect(Collectors.toList());
+    return (stopTaskActions.size() == 0) ? startTaskActions : stopTaskActions;
```
wjbuys: Why not return all of the actions? It should be safe to emit stop/start tasks in bulk, as long as they don't act on the exact same task ID.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
```diff
@@ -48,6 +54,30 @@ public StartTask startTaskFor(ContainerInstance i) {
...
+            .map(Task::getTaskDefinitionArn)
+            .collect(Collectors.toList());
+    return tasksToStop
+        .stream()
+        .map(
+            t ->
+                StopTask.builder()
+                    .clusterName(environment.getClusterName())
+                    .task(t)
+                    .reason("replaceAfterTerminate")
```
wjbuys: Is this string customer-visible? If so, we should plan for making it suitably descriptive, e.g. "Stopped by deployment to EnvironmentBlah@version2".

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
```diff
@@ -48,6 +54,30 @@ public StartTask startTaskFor(ContainerInstance i) {
...
+                    .build())
+        .collect(Collectors.toList());
+  }
+
+  public boolean unmatchesTask(Task t) {
```
wjbuys: From the name, I would expect `unmatchesTask(t) == !matchesTask(t)` for all `t`. This doesn't seem to be the case, so perhaps it's time to give both of these methods more descriptive names?

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/ReplaceAfterTerminateScheduler.java
```diff
@@ -15,18 +15,29 @@
...
+
+    List<SchedulingAction> stopTaskActions = new ArrayList<>();
+    summary
         .getInstances()
         .stream()
-        .filter(i -> env.hasMatchingTask(summary.tasksForInstance(i)))
-        .map(i -> env.startTaskFor(i))
-        .collect(Collectors.toList());
+        .filter(i -> env.hasUnmatchingTask(summary.tasksForInstance(i)))
+        .forEach(i -> stopTaskActions.addAll(env.stopTaskFor(summary.tasksForInstance(i))));
```
wjbuys: This can be written with a `flatMap` instead of imperatively modifying `stopTaskActions` using `forEach`.

scheduling-manager/src/cucumber/resources/features/scheduling/replace_after_terminate.feature
```diff
@@ -37,3 +37,15 @@ Feature: Deploying a Daemon environment to a cluster with Replace After Terminat
...
+    Given the cluster has the following instances and tasks:
+      | instance | tasks          |
+      | i-1      | t-1:v1:RUNNING |
+      | i-2      | t-2:v2:RUNNING |
+      | i-3      |                |
+    When the scheduler runs
+    Then it should stop the following tasks:
+      | clusterName | task           | reason                |
+      | TestCluster | v2             | replaceAfterTerminate |
+    And it should not take any further actions
```
wjbuys: Shouldn't this also start a task for `i-3`?

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
```diff
@@ -48,6 +54,30 @@ public StartTask startTaskFor(ContainerInstance i) {
         .build();
   }
 
+  public List<StopTask> stopTaskFor(final List<Task> tasks) {
+    final List<String> tasksToStop =
```
wjbuys: This method (`stopTaskFor`) is named just like `startTaskFor`, but it's doing more than just mapping tasks to `StopTask` actions as `startTaskFor` does for `StartTask` actions.

Maybe consider splitting out the filtering to a separate method, and have this method only do the mapping to `StopTask` objects.

xjhe:
I can add two checks: 
1. Check the task status, so we only stop a task when it's RUNNING or PENDING.
2. Allow the scheduler to start & stop tasks on the same step.

wjbuys (changes requested):

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
```diff
@@ -48,6 +54,30 @@ public StartTask startTaskFor(ContainerInstance i) {
...
+                StopTask.builder()
+                    .clusterName(environment.getClusterName())
+                    .task(t)
+                    .reason("replaceAfterTerminate")
+                    .build())
+        .collect(Collectors.toList());
+  }
+
+  public boolean unmatchesTask(Task t) {
+    return !t.getGroup().equals(environment.getEnvironmentName())
```
wbingli: We should only compare the task which is under the same environment.  If task is in different env name, they are managed by different environment and different config.   Should it be:
` return t.getGroup().equals(environment.getEnvironmentName()) && !t.getTaskDefinitionArn().equals(environment.getTaskDefinitionArn());`

wbingli (changes requested):

scheduling-manager/src/cucumber/resources/features/scheduling/replace_after_terminate.feature
xjhe: As discussed, the shceduler will stop the task on i-2 and start a task on i-3. And I will make it a separate case for testing this multiple step scheduling results.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
xjhe: Sure, will revise all the naming and restrucutre these actions.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/ReplaceAfterTerminateScheduler.java
xjhe: As discussed, I'll return all actions.
When checking for startTaskActions, I'll add a check to see if there is already a task running.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
xjhe: Sure. Good catch!

xjhe:
In this update, I tried my best to rename the methods in DaemonEnvironment, but some of them are still not that intuitive. If you have better suggestions please comment!

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/ReplaceAfterTerminateScheduler.java
```diff
@@ -15,18 +15,35 @@
...
+        summary
+            .getInstances()
+            .stream()
+            .filter(i -> env.needsToAssign(summary.tasksForInstance(i)))
+            .filter(i -> env.noRunningTasks(summary.tasksForInstance(i)))
+            .map(i -> env.startTaskFor(i))
+            .collect(Collectors.toList());
+
+    startTaskActions.addAll(stopTaskActions);
+    return startTaskActions;
```
wbingli: The return tasks are all actions, should be in a better name. or return a list with two actions list like `ImmutableList.Builder().addAll(startTaskActions).addAll(stopTaskActions).build()`

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/ReplaceAfterTerminateScheduler.java
```diff
@@ -15,18 +15,35 @@
...
-        .getInstances()
-        .stream()
-        .filter(i -> env.hasMatchingTask(summary.tasksForInstance(i)))
-        .map(i -> env.startTaskFor(i))
-        .collect(Collectors.toList());
+    List<ClusterSnapshot.Task> tasksToStop =
+        summary
+            .getInstances()
+            .stream()
+            .filter(i -> env.needsToStop(summary.tasksForInstance(i)))
```
wbingli: Not sure the filter on instance for `needsToStop` has value here. if the `env.taskToStop(t)` can find out all the task to stop, if there is no task in the instance, a empty list will return and it's good.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/ReplaceAfterTerminateScheduler.java
xjhe: Yes, this is clearer!

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/ReplaceAfterTerminateScheduler.java
xjhe: This is a good point -- stop task doesn't check instance status, so I should just use the filter on tasks.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/ReplaceAfterTerminateScheduler.java
```diff
@@ -15,18 +15,37 @@
...
-        .stream()
-        .filter(i -> env.hasMatchingTask(summary.tasksForInstance(i)))
-        .map(i -> env.startTaskFor(i))
-        .collect(Collectors.toList());
+    List<ClusterSnapshot.Task> tasksToStop =
+        summary
+            .getInstances()
+            .stream()
+            .flatMap(i -> summary.tasksForInstance(i).stream().filter(env::taskToStop))
+            .collect(Collectors.toList());
```
wjbuys: Perhaps avoid collecting this to a list here if we're just going to stream it again in the next statement.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
```diff
@@ -48,7 +53,24 @@ public StartTask startTaskFor(ContainerInstance i) {
...
+        .build();
+  }
+
+  public boolean taskToStop(Task t) {
+    return t.getGroup().equals(environment.getEnvironmentName())
+        && !t.getTaskDefinitionArn().equals(environment.getTaskDefinitionArn())
+        && HEALTHY_STATES.contains(t.getStatus());
+  }
+
+  public boolean taskDesiredToRun(Task t) {
```
wjbuys: How about `isTaskRunnable`?

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
```diff
@@ -48,7 +53,24 @@ public StartTask startTaskFor(ContainerInstance i) {
...
+        .clusterName(environment.getClusterName())
+        .task(task.getTaskDefinitionArn())
+        .reason(
+            String.format(
+                "Stopped by deployment to %s@%s",
+                environment.getEnvironmentName(), environment.getTaskDefinitionArn()))
+        .build();
+  }
+
+  public boolean taskToStop(Task t) {
```
wjbuys: How about `isTaskStoppable`?

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
```diff
@@ -35,8 +36,12 @@
...
   private final EnvironmentDescription environment;
 
-  public boolean hasMatchingTask(List<Task> tasks) {
-    return tasks.stream().noneMatch(this::matchesTask);
+  public boolean needsToAssign(List<Task> tasks) {
+    return tasks.stream().noneMatch(this::taskDesiredToRun);
+  }
+
+  public boolean noRunningTasks(List<Task> tasks) {
+    return tasks.stream().noneMatch(t -> HEALTHY_STATES.contains(t.getStatus()));
```
wjbuys: The method name says `Running`, but the implementation says `HEALTHY`, so maybe `noHealthyTasks`?

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/ReplaceAfterTerminateScheduler.java
```diff
@@ -15,18 +15,37 @@
...
+
+    List<SchedulingAction> stopTaskActions =
+        tasksToStop.stream().map(env::stopTaskFor).collect(Collectors.toList());
+
+    List<SchedulingAction> startTaskActions =
+        summary
+            .getInstances()
+            .stream()
+            .filter(i -> env.needsToAssign(summary.tasksForInstance(i)))
+            .filter(i -> env.noRunningTasks(summary.tasksForInstance(i)))
```
wjbuys: Any reason for splitting out `needsToAssign` and `noRunningTasks`? It doesn't look like they're ever used separately. Perhaps combining them into `isMissingHealthyTask(tasks)` would make this a bit clearer?

scheduling-manager/src/cucumber/resources/features/scheduling/replace_after_terminate.feature
```diff
@@ -37,3 +37,29 @@ Feature: Deploying a Daemon environment to a cluster with Replace After Terminat
...
+
+  Scenario: Scheduling on a cluster with tasks that don't match the environment version
+    Given the cluster has the following instances and tasks:
+      | instance | tasks          |
+      | i-1      | t-1:v1:RUNNING |
+      | i-2      | t-2:v2:RUNNING |
+    When the scheduler runs
+    Then it should stop the following tasks:
+      | clusterName | task           | reason                |
+      | TestCluster | v2             | Stopped by deployment to DaemonEnvironment@v1 |
```
wjbuys: This test case might be easier if we make a new feature file, where the current version of the environment is `v2`, so that versions are changing forward in time. Perhaps consider splitting these into:

```
features/replace_after_terminate/initial_deployment.feature
features/replace_after_terminate/update_deployment.feature
```

wjbuys (approved):
Looks pretty good, just some minor naming issues, apply as you see fit.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/ReplaceAfterTerminateScheduler.java
```diff
@@ -15,18 +15,37 @@
...
+            .map(env::startTaskFor)
+            .collect(Collectors.toList());
+
+    return new ImmutableList.Builder<SchedulingAction>()
```
wjbuys: You can avoid the dependency on Guava by doing:

```java
return Stream.concat(startTaskActions.stream(), stopTaskActions.stream()).collect(Collectors.toList());
```

scheduling-manager/src/cucumber/resources/features/scheduling/replace_after_terminate.feature
xjhe: I tried separating it out and figure that there are only two test scenarios in the new file. If we consider adding more to the update_deployment one that'll be a good idea

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
xjhe: Sure. (Yes I realize that there are two Healthy states...)

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
xjhe: Sure. I was struggling with naming. We should come up with some naming conventions?

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/ReplaceAfterTerminateScheduler.java
xjhe: Sure! And that'll fit in your next comment on using Stream.concat.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/ReplaceAfterTerminateScheduler.java
xjhe: Sure.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/DaemonEnvironment.java
wjbuys: Yeah. I think in general, the problem is that we are trying to interrogate properties of the instance/tasks. If we were to express these questions naturally, we would say (roughly)

```java
instances.filter((instance) -> instance.needsNewTask()).map(startTask);
tasks.filter((task) -> task.canBeStopped()).map(stopTask);
```

Since we don't have `needsNewTask` and `canBeStopped` methods on those entities (because we need the `DaemonEnvironment`'s data to make that decision), we have to basically invert the names, and prefix them with what the method receiver would have been:

```java
instance.filter((instance) -> env.instanceNeedsNewTask(instance)).map(startTask)
tasks.filter((task) -> env.taskCanBeStopped(task)).map(stopTask)
```

That gets tricky when e.g. `env.taskCanBeStopped(task)` depends on the instance that it's on, as will be the case for the next deployment method, TerminateAfterReplace.

xjhe: You point out my pain correctly :-D  I have the feeling that in DaemonEnvironment we're inverting the names. I can go with the current naming (of course taking your suggestion) and update them when we finish the TerminateAfterReplace.

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/engine/daemon/ReplaceAfterTerminateScheduler.java
xjhe: After thinking through the logic, we actually only need to check noRunningTasks() for start. 
If the task is HEALTHY, even if it belongs to another environment/ another version, we still won't start a new task. We'll just wait for the scheduler to stop this current task.
Therefore, we only need to check the status of the task, regardless of the environment name and task version.

wjbuys: > If the task is HEALTHY, even if it belongs to another environment/ another version, we still won't start a new task

That doesn't sound right; we *do* want to start a daemon task on an instance, even if there's already unrelated (i.e. from another env/no env) tasks running on it. Am I missing something here?

xjhe: I see! I was under the impression that the same instance won't be shared across different environments. If there can be multiple environments on an instance, then yes, we do need the check on matching environment name.

xjhe: The correct change should be, just check the matching environment name and RUNNING task status. We don't need to check task version, as we don't want to start a task when an older version is still running: we will stop it first.

wbingli (approved):

PR #369, opened 2018-02-05
Setup CodeBuild for Blox end-2-end test

wbingli:
#### Summary
This PR includes some configuration files and setup steps for whole CodeBuild setup. 

The permission for CodeBuild execution is still too board.  I'm afraid even we narrow down, it would still too much since it's not just about run the e2e, but also includes whole project deployment.    It means it includes all following,  1)deployment permissions; 2) application runtime permissions; 3) Test case permissions.  Once we have a pipeline to break down those steps, we can separate those permission into different user/role.

Only for this PR as an demo for the codebuild, I would suggest we don't enable it for pull request, but only  enable for commit "push", which are authorized code. 

#### Implementation details
- A cloudformation to setup the codebuild project and permission
- buildspec and properties for test and environment config
- README for the setup

#### Testing
 --> See This PR commit

#### Licensing


end-to-end-tests/code-build/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+We use CodeBuild to run quick end-to-end test. Below are steps to set it up:
+
+### Running `deploy` to setup blox application
+To avoid too broad permissions required for CodeBuild project, we setup blox project firstly.
+
+Make sure you have one AWS credential from the AWS account and used in **default** profile:
+
+```
+$ cat end-to-end-tests/code-build/end2end.properties >> ~/.gradle/gradle.properties
```
wjbuys: We probably don't want to make this permanent change. Can we use one of the alternatives outlined in [the docs](https://docs.gradle.org/current/userguide/build_environment.html) to change these gradle properties temporarily?

end-to-end-tests/code-build/README.md
```diff
@@ -0,0 +1,33 @@
...
+./gradlew createBucket
+./gradlew deploy
+```
+
+### Create CodeBuild Project
+Before we create the project, you must connect the AWS account to the GitHub account. 
+To do this, use the AWS CodeBuild console to create a build project. When you use the console to connect (or reconnect) with GitHub, on the GitHub Authorize application page, for Organization access, choose Request access next to each repository you want AWS CodeBuild to be able to access. Choose Authorize application.
```
wjbuys: This is kind of hard to parse; maybe split out into a bullet list?

wjbuys (approved):

wjbuys:
I wonder if we should add a CodeBuild artifact declaration for the test reports in `build/reports/`? That way we can actually see test failures a bit more clearly.

end-to-end-tests/code-build/README.md
wbingli: Sure.

end-to-end-tests/code-build/README.md
wbingli: yep, will update.

wbingli:
>I wonder if we should add a CodeBuild artifact declaration for the test reports in build/reports/? That way we can actually see test failures a bit more clearly.

Make sense, I will add it.

wbingli:
Add test report as artifact, See example:  https://s3-us-west-2.amazonaws.com/blox-codebuild-us-west-2/codebuild/report/a5e0af86-b78c-478c-b8a5-5e4818b365be/Blox-E2E-Test/end-to-end-tests/build/reports/tests/testEndToEnd/index.html

wjbuys:
How do I actually see that? I get Access Denied...

wbingli:
> How do I actually see that? I get Access Denied...

I don't set it as public. It's still private and need to go to console to access it.   Let me see we can make it as public available since it's not a secure.

end-to-end-tests/code-build/buildspec.yml
```diff
@@ -0,0 +1,26 @@
+version: 0.2
+
+phases:
+  pre_build:
+    commands:
+      - mkdir -p /root/.gradle
+      - echo "org.gradle.daemon=false" > /root/.gradle/gradle.properties
+      - cat end-to-end-tests/code-build/end2end.properties >> /root/.gradle/gradle.properties
+      - aws configure --profile default set region us-west-2
```
xjhe, line 9: Are we fixed with the region "us-west-2"? Do we allow customized regions

xjhe (approved):
I just add one minor question. Overall looks good

end-to-end-tests/code-build/buildspec.yml
wbingli: This one most likely will not be reused to setup multiple regions.  It's mostly for rebuild/recover/update the test environment. So far it's only for **us-west-2**.

Actually this command line is tricky one `aws configure --profile default set region us-west-2`, it will force to create a `default` profile.  Because if you specify `--profile default`, it requires a default profile exists even it will try to follow the credential chain to find credential.

wbingli:
@wjbuys 

>>How do I actually see that? I get Access Denied...

>I don't set it as public. It's still private and need to go to console to access it. Let me see we can make it as public available since it's not a secure.

Looked at more,  it looks we cannot easily have a public test report. CodeBuild will do a SSE-KMS (only) encryption for the artifact. It causes the issue that we cannot simply provide a HTTP URL to access even we make S3 object public accessible.  

Since we only do e2e test for push commit, it makes sense to only have authorized user to access the report from AWS console.  There may be some workaround to change the encryption but don't see the value on the effort.

end-to-end-tests/code-build/buildspec.yml
xjhe: Understood. 
I think the `aws configure --profile default set region us-west-2` will not create the default profile if it already exists? And if the user have already created the stacks, he/she should already have the default profile :-D

wbingli: yep, it will not create a new one if it exists.  This configuration file is used by CodeBuild, so every time is a completely new environment.  In this context, it will actually create a new one.

The reason we want to use `default` profile is we need to use the default credential chain, which will look for the instance profile (or technically it's container credential profile). See https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html#credentials-default

PR #370, opened 2018-02-07
Add state diagram for terminate then replace schduler

xjhe:
#### Summary
Add a state diagram

#### Testing
Not sure how to test the document? The preview doesn't generate the table for me.

#### Licensing


docs/diagrams/terminateThenReplaceSchedulerStateTransition.puml
```diff
@@ -0,0 +1,16 @@
+@startuml
+[*] --> state2
```
wbingli: From the sequence table, the state3 should be first one to start?  I'm not sure if this diagram better than the table list, it makes me really hard to understand like why `state2 -> state1` and `state5 -> state1` and `state4 -> state1`. It kinds of mixing the states and two tasks.

docs/deployment-ux.md
```diff
@@ -581,7 +581,37 @@ SomeEnvironment/Prod:1    Undeploying  3         4
...
+
+| T | Event                                    | i<sub>1</sub>                                            | action          |
+|:--|:-----------------------------------------|:---------------------------------------------------------|:----------------|
+| 0 | Create Environment(v1)                   |                                                          | start task      |
+| 1 | task t1:v1 running                       | :white_check_mark:<sub>v1</sub>                          | no action       |
+| 2 | Update Environment(v2)                   | :white_check_mark:<sub>v1</sub>                          | stop task       |
+| 3 | task t1:v1 terminating                   | :red_circle:<sub>v1</sub>                                | no action       |
+| 4 | task t1:v1 terminated                    | :no_entry:<sub>v1</sub>                                  | start task      |
+| 5 | task t2:v2 running                       | :white_check_mark:<sub>v2</sub>                          | no action       |
+| 6 | task t2:v2 terminating                   | :red_circle:<sub>v2</sub>                                | start task      |
```
wbingli: why the t2:v2 need to terminating after running?

wbingli (commented):
It looks to me the existing "Appendix A" describe clear enough for the TerminateThenReplace states change.

docs/diagrams/terminateThenReplaceSchedulerStateTransition.puml
xjhe: The table lists the sequence of how an example deployment works, while the diagram describes the steady state. So state1 ~ state5 doesn't have any sequential order. Maybe renaming them into stateA~E would be more helpful? 

I think what's missing from appendix A is what would be like in a steady state. Also, we should separate the "condition" (the cluster snapshot that we can read) and the "action" (the action that the scheduler take) in the sequence table. that's why I add one more column in the sequence table here.

docs/deployment-ux.md
xjhe: I want to describe the case when a task is accidentally terminated (unhealthy, or terminated by the user?), what action will the scheduler take. Task termination should be a valid case that we should consider

docs/diagrams/terminateThenReplaceSchedulerStateTransition.puml
wjbuys: Hm. Perhaps giving the states meaningful names will make this clearer? E.g. state2 is "Empty", state1 is "Active", state3 is "OutdatedRunning", state4 is "OutdatedStopped", state5 is "Failing".

docs/deployment-ux.md
wjbuys: In that case, the event would be "task t2:v2 terminated abnormally", and the state would be "terminated" (since the termination was not requested through the control plane).

docs/deployment-ux.md
```diff
@@ -581,7 +581,37 @@ SomeEnvironment/Prod:1    Undeploying  3         4
...
+The states and actions of each transition step on an environment update can be described in the following table. Just consider one instance `i1`. 
+
+| T | Event                                    | i<sub>1</sub>                                            | action          |
+|:--|:-----------------------------------------|:---------------------------------------------------------|:----------------|
+| 0 | Create Environment(v1)                   |                                                          | start task      |
+| 1 | task t1:v1 running                       | :white_check_mark:<sub>v1</sub>                          | no action       |
+| 2 | Update Environment(v2)                   | :white_check_mark:<sub>v1</sub>                          | stop task       |
+| 3 | task t1:v1 terminating                   | :red_circle:<sub>v1</sub>                                | no action       |
+| 4 | task t1:v1 terminated                    | :no_entry:<sub>v1</sub>                                  | start task      |
+| 5 | task t2:v2 running                       | :white_check_mark:<sub>v2</sub>                          | no action       |
```
wjbuys: I think you may need to include "Pending" (the state before "Running") here for completeness.

wjbuys (commented):
Thanks for taking the time to document this. Having it in writing makes it easier to talk about some of the edge cases. Some small naming suggestions should make this even clearer, though.

docs/diagrams/terminateThenReplaceSchedulerStateTransition.puml
xjhe: Let me change the name to the actual behavior of these states!

docs/deployment-ux.md
xjhe: Sure. The termination is not triggered by our actions. I can try to distinguish between the abnormal state and the normal state.

docs/deployment-ux.md
xjhe: Sure

PR #371, opened 2018-02-08
Add UpdateEnvironment API to frontend

wjbuys:
This adds the plumbing for the `updateEnvironment` API call on the Frontend. Since we're not sure how to handle role/instanceGroup updates yet, I've removed those from the DataService API, and omitted them from the frontend.

#### Testing
```
gradle check
```

#### Licensing

### DIFF
--- a/frontend-service/src/test/java/com/amazonaws/blox/frontend/operations/CreateEnvironmentTest.java
+++ b/frontend-service/src/test/java/com/amazonaws/blox/frontend/operations/CreateEnvironmentTest.java
@@ -20,14 +20,11 @@ import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 import com.amazonaws.blox.dataservicemodel.v1.model.DeploymentConfiguration;
-import com.amazonaws.blox.dataservicemodel.v1.model.Environment;
 import com.amazonaws.blox.dataservicemodel.v1.model.EnvironmentId;
-import com.amazonaws.blox.dataservicemodel.v1.model.EnvironmentRevision;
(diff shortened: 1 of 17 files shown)


frontend-service/src/test/java/com/amazonaws/blox/frontend/operations/CreateEnvironmentTest.java
```diff
@@ -51,26 +48,8 @@ public void mapsInputsAndOutputsCorrectly() throws Exception {
...
-                        .build())
-                .environmentRevision(
-                    EnvironmentRevision.builder()
-                        .environmentId(id)
-                        .environmentRevisionId(ENVIRONMENT_REVISION_ID)
-                        .instanceGroup(instanceGroup)
-                        .taskDefinition(TASK_DEFINITION)
-                        .createdTime(Instant.now())
-                        .build())
+                .environment(environmentDS(id, deploymentConfiguration))
```
xjhe, line 51: Not in the scope of this PR, but I think we can replace the construction of Environment/EnvironmentRevision with the ones from EnvironmentControllerTestCase in other FE operation tests as well.

xjhe (approved):

frontend-service/src/test/java/com/amazonaws/blox/frontend/operations/CreateEnvironmentTest.java
wjbuys: Agreed. I started down that route, and figured we'd want to use the ModelBuilders from the DS integ tests instead.

PR #372, opened 2018-02-10
Implement UpdateEnvironment in DataService

wjbuys:
This is the simplest possible implementation of UpdateEnvironment. It
doesn't take into account possible race conditions or partial failures
for updateing `latestEnvironmentRevisionId` on the environment, but as
long as callers only rely on the environmentRevisionId return from the
call itself, there should be no problem.

#### Known issues
- doesn't support updating InstanceGroup (yet)
- doesn't deal with concurrent updates or partial failures of updating `Environment.latestEnvironmentRevisionId`

#### Testing
New DataService integration tests to verify basic functionality; will add finer-grained unit tests when we revisit this to fix the known issues above.

#### Licensing


wbingli (approved):

xjhe (approved):
Looks good!

PR #373, opened 2018-02-12
Fix NullPointerException for inactive environments in scheduler handler

wjbuys:
The scheduler handler incorrectly assumed that an environment will
always have an active environment revision. This is not true for
newly-created environments, and would cause scheduler invocations to
fail with a NullPointerException.

Instead, it now returns an empty SchedulerOutput as soon as it's clear
that the environment has no active revisions.

#### Testing
New unit test passes.

#### Licensing


wbingli (approved):
lgtm

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/SchedulerHandler.java
```diff
@@ -59,6 +59,11 @@ public SchedulerOutput handleRequest(SchedulerInput input, Context context) {
 
     String activeEnvironmentId = environment.getActiveEnvironmentRevisionId();
```
xjhe: As a suggestion, we can change this into activeEnvironmentRevisonId to avoid confusion

xjhe (approved):
Looks good!

scheduling-manager/src/main/java/com/amazonaws/blox/scheduling/scheduler/SchedulerHandler.java
wjbuys: Oh, whoops totally a typo, will fix.

PR #374, opened 2018-02-12
Remove unused/broken updateEnvironment cucumber test case

wjbuys:
This test case didn't work, and was not being run. It also relies on
UpdateEnvironment returning full Environment/EnvironmentRevision
objects, which is no longer the case.

#### Testing
```
gradle cucumberTest
```

#### Licensing


wbingli (approved):

xjhe (approved):
Good to go

PR #375, opened 2018-02-12
Add end-to-end test for UpdateEnvironment

wjbuys:
#### Summary
The main goal of this PR is to add an end-to-end test for UpdateEnvironment.

#### Implementation details
Apart from adding the new test case, this also fixes the following issues that was preventing the end-to-end test from working:
- fix an FE bug where `environmentName` was present in both the URI path and the request body of `updateEnvironment`
- the test cluster now has multiple instances, and the tests verify that tasks are started on all of them
- the test stack has an additional task definition, so that we can change task definitions

#### Testing
```
gradle testEndToEnd
```
#### Licensing

### DIFF
--- a/end-to-end-tests/templates/test_cluster.yml
+++ b/end-to-end-tests/templates/test_cluster.yml
@@ -3,7 +3,7 @@ AWSTemplateFormatVersion: '2010-09-09'
 Parameters:
   DesiredInstances:
     Type: Number
-    Default: '1'
+    Default: '2'
     Description: Number of instances to launch in your ECS cluster.
   DesiredTasks:
     Type: Number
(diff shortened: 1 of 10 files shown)


end-to-end-tests/templates/test_cluster.yml
```diff
@@ -55,6 +55,21 @@ Resources:
...
+    Properties:
+      Family:
+        Fn::Join: ['', [{ Ref: 'AWS::StackName' }, -persistent-task]]
+      ContainerDefinitions:
+      - Name: sleep
+        Cpu: '10'
+        Memory: '300'
+        Essential: 'true'
+        Image: 'alpine:latest'
+        Command: ['/bin/sh', '-c', 'trap : TERM INT; sleep infinity & wait']
```
xjhe, line 69: Looks like this task just sleeps forever, so it will never end. Why do we want this behavior for this task? I think the transient task definition also works for testing the update environment.

wjbuys: Primarily, I justified this to myself by reasoning that the task should not accidentally exit (because its timeout expired) before it can be upgraded explicitly by the scheduler.

However, mainly I just needed another task definition that could be referenced from the tests. Rather than have a duplicate definition, I figured that having a task with a different behaviour could be useful to have as a test fixture.

xjhe: Makes sense! Actually, a daemon task should always be running in the background, so the new task that you added is closer to a real daemon task.

xjhe (approved):

PR #377, opened 2018-02-16
Allow `testEndToEnd` task takes configuration from command line options

wbingli:
#### Summary
Provide gralde task options to set configuration so we can test with different endpoint.

It also remove the depends on the deployment, so running test will only
run test without deployment. To run both deploy and test, you can do
`gradle deploy testEndToEnd`.  It will also be good for pipeline, which we could have the first step as test environment deployment `gradle end-to-end-tests:deploy` and then following steps to run test case. 

Still could simply run `testEndToEnd`, it will use the existing way for default profile and endpoint.

Refactor the e2e test to move the share part into abstract class.

#### Testing
- `./gradlew testEndToEnd`    - Default way
- `./gradlew testEndToEnd --aws-region us-west-2 --aws-profile blox-libruce-alpha-us-west-2 --endpoint https://mn2aq1kb4k.execute-api.us-west-2.amazonaws.com` - With configuration
- `./gradlew testEndToEnd --aws-region us-west-2 --aws-profile blox-libruce-alpha-us-west-2 --endpoint https://mn2aq1kb4k.execute-api.us-west-2.amazonaws.com --tests com.amazonaws.blox.CreateDaemonEnvironmentTest`   - Only test one test case

#### Licensing


end-to-end-tests/build.gradle
```diff
@@ -30,19 +32,13 @@ deployment {
     templateFile file("templates/test_cluster.yml")
 }
 
-task testEndToEnd(type: Test) {
+task testEndToEnd(type: EndToEndTest) {
```
xjhe, line 35: Here the dependency on doing a deployment is removed. Does it assume that the deployment has already been made in a previous step? If so, is there a way to check if the deployment has been done?

xjhe (commented):
Overall looks good to me. I'm not that familiar with the gradle task part, but other parts look fine.

end-to-end-tests/build.gradle
wbingli: Yes, we kind of assume the deployment should be done while taking the test.  The sequence is controlled in the pipeline. 

As for local development,  we don't have such check to verify if the deployment is done. But  you can always run `deploy`, as gradle can do cache and it won't run if it has cache in local.

wjbuys: A good way to solve this is to declare that that `testEndToEnd.mustRunAfter(deploy)` (see the docs: https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:ordering_tasks)

This will ensure that if both tasks are being run (e.g. `gradle deploy testEndToEnd`), then `testEndToEnd` will wait for `deploy`, but if only `testEndToEnd` is being run, `deploy` won't also run.

end-to-end-tests/src/test/java/com/amazonaws/blox/AbstractEndToEndTest.java
```diff
@@ -0,0 +1,67 @@
...
+        .deleteEnvironment(
+            new DeleteEnvironmentRequest()
+                .cluster(stack.getCluster())
+                .environmentName(environmentName));
+
+    // Cleanup ECS tasks
+    stack.reset();
+  }
+
+  void assertTaskStarted(final String environmentName, final String taskDefinition)
```
wjbuys: This method name isn't really accurate; perhaps `assertAllRunningTasksMatch()`?

wjbuys (approved):

end-to-end-tests/build.gradle
xjhe: That makes sense. In the pipeline the deployment will be kicked off before the integ test. My concern is that the pipeline is deploying to our blox_frontend account, while the e2e test is deploying to the e2e test account, so these two cannot be mixed together? 
We can of course trigger a deployment to the e2e test account in the TOD itself.

xjhe (approved):

end-to-end-tests/build.gradle
wbingli: There will be two kinds of deployment:
 - Service deployment:  it's about the deployment for service, like apitgateway, frontend, data-service tc.
 - End to end test deployment:  To run end to end test, we need setup some test resources.  Those test resources will be re-used by all test cases, such as ECS clusters and contain instances etc.  

In the end of day, they are running into different AWS accounts.

My idea is that our deployment in the pipeline will be just the service deployment.  The end-2-end deployment is not a **deployment** in pipeline context, it's like a setup for end to end test.  We could make it as the first step before we starts the end2end. In this way, even we want to  parallel the tests, we will only need to run the deployment for e2e test once.

xjhe: Yeah, I was a little bit confused on the deployments, but I'm cleared after your explanation. and Yes, I can see the benefit of doing one e2e deployment, and run the e2e tests multiple times.

PR #378, opened 2018-02-19
Fix the E2E test for codebuild to let deploy first

wbingli:
#### Summary

The gradle task order does not imply execution order, therefore the
default endpoint for testEndToEnd is null as deployment output file is
not ready yet.

For local development, it won't be a problem as the deployment output
file should exist already. We make the code build as two steps so we
could have endpoint information ready before run test task.

Once we have proper pipeline,  it won't be necessary to use codebuild for e2e testing.

#### Testing
Passed locally

#### Licensing


xjhe (approved):
Looks good to me

wjbuys:
Yeah, the problem is that the file is read in the gradle Configuration phase, while the file is written only during the Execution phase. Since Configuration happens first, the file won't exist during that phase (since Execution only happens later).

PR #381, opened 2018-03-09
Add generate uber jar tasks for end-to-end-tests

dramaticlly:
#### Summary
Add generate uber jar tasks for end-to-end-tests
- on blox release, copy the uber jar to correct path
- changed AbstractEndToEndTest to use either SystemProperty or EnvironmentVariable since we need to considering the possibility to run end-to-end tests on Lambda

#### Testing
bb release on the package root directory
Output uber jar copied to:
${Brazil_workspace_root}/build/blox-release

jar tf end-to-end-tests-0.1-SNAPSHOT-tests.jar |wc -l
10559

#### Licensing


end-to-end-tests/src/test/java/com/amazonaws/blox/AbstractEndToEndTest.java
```diff
@@ -31,7 +31,10 @@
 
   @Before
   public void setupBase() {
-    final String bloxEndpoint = System.getProperty("blox.tests.apiUrl");
+    String bloxEndpoint = System.getProperty("blox.tests.apiUrl");
+    if (bloxEndpoint == null || bloxEndpoint.isEmpty()) {
```
wjbuys: You could use the two-argument form of `System.getProperty(key, default)` to simplify things here:
```
System.getProperty("blox.tests.apiUrl", System.getenv("BLOX_TESTS_APIURL"));
```

build.gradle
```diff
@@ -243,8 +243,23 @@ configure(projectsToRelease) {
             }
         }
     }
-
     rootProject.release.dependsOn(release)
 }
 
 
+def testArtifactProject = ["end-to-end-tests"].collect { project(it) }
+configure(testArtifactProject) {
```
wjbuys: Since you're only dealing with one project, and that project's `testJar` task is not defined by a plugin like the `postprocessCloudFormationTemplate` task above, I think you can simplify this to:

```
configure(project("end-to-end-tests")) {
  task release(type: Copy) {
    from testJar
    into rootProject.file("build/blox-release/${project.name}")
  }
  rootProject.release.dependsOn(release)
}
```

wjbuys (approved):
LGTM, but some Gradle simplifications are possible

end-to-end-tests/src/test/java/com/amazonaws/blox/AbstractEndToEndTest.java
dramaticlly: good suggestions! will change

build.gradle
dramaticlly: will simplify, thanks!

dramaticlly:
updated with new commit, pending reviewer feedback

wbingli (approved):
LGTM

